Issue #95, opened 2024-12-05
Add support for page-level metadata


lcawl:
Relates to https://github.com/elastic/docs-builder/issues/49 

Our current proposal for representing the version/context of each page in the new information architecture and doc system needs a method for clearly identifying all the contexts that a page is applicable to.

In the current docs system, context needs to be derived from the text itself, the "book" breadcrumbs and version selector. Since books tended to apply to a single product or context, that was usually simpler to infer. For example:

![Image](https://github.com/user-attachments/assets/0a2835c3-ccad-4a1a-ac9f-038d767f63fc)

In the new docs system and information architecture you can't assume pages that are "near" each other are all applicable to the same context, so we should be as clear as possible about the context. For example:

![Image](https://github.com/user-attachments/assets/1c578e1c-66a4-4d69-8f19-41228be0933e)

At this moment, we are proposing that the page-level metadata uses the same deployment type, lifecycle, and version dimensions as described in https://github.com/elastic/docs-builder/issues/94. The assumption being that what is specified at the page-level is applicable to the entire page unless it's overridden by admonitions at the section, page, or line level.

NOTE: We haven't gotten input from the design team yet on the best possible way to make these admonitions informative but not too disruptive. But the vision is informed by the type of "badges" we had in the Serverless docs. 

## Short term

In the short term we would want a consistent way to apply this metadata at a page level and just show them all.
Ideally it would be possible to set this metadata in a configuration file for all the pages in a folder, for example, so that we only needed to set it at a page level if it differed from the default for that content set.

## Long term

In the longer term, we likely need input from design and user feedback on whether it's more useful to only show a subset of contexts (e.g. only show badges that indicate where it *is* supported or vice versa).

In the even longer term, it would be nice to have this metadata affect search and filtering functionality.

Mpdreamz:
Closing this as implemented even though we might go through a few more refinement cycles.

Issue #94, opened 2024-12-05
Add support for context-specific admonitions


lcawl:
Relates to https://github.com/elastic/docs-builder/issues/49, https://github.com/elastic/docs-builder-example/blob/main/docs/source/markup/admonitions.md

In order to support versioning in the new docs system, we need to add some built-in admonitions or "directives" that are context-specific and can be applied at the section, paragraph, or line-level.

In the existing doc system, they look like this:

![Image](https://github.com/user-attachments/assets/756ca78b-41b3-4ed3-ba07-e216c350da66)

However this is insufficient for features have varied states across multiple contexts. For example something can be GA in one context but tech preview in another.

In the new doc system, we should have a way to accomplish something like this:

![Image](https://github.com/user-attachments/assets/5dc85cee-36e4-4172-a983-70d48eb5fad2)

You can see an example of how we played with extending the admonitions in the old system to accomplish this in https://github.com/elastic/docs/pull/3121

Similar to the page-level metadata, we think that that admonitions need to communicate the following dimensions:

- Deployment type. For example:
    - Elastic Cloud Hosted
    - Elastic Cloud Serverless
    - Elastic Cloud Enterprise
    - Elastic Cloud Kubernetes
    - Self-managed (need to confirm the terminology for this Elastic Stack on-prem self-managed context)
- Availability / product stage
    - technical preview (exists in current docs system per https://github.com/elastic/docs?tab=readme-ov-file#beta-dev-and-preview-experimental)
    -  beta (ditto)
    - dev (ditto, though it's uncertain whether it's ever used or still needed)
    - deprecated (exists in current docs system per https://github.com/elastic/docs?tab=readme-ov-file#additions-and-deprecations)
    - coming (ditto)
    - discontinued (historically we've immediately removed content when the feature ceases to be supported, but this might not be the case with pages that contain information that spans versions)
    - unavailable (for content that doesn't exist in a specific context and is never coming or not coming anytime soon)
    - ga (replaces "added" in the current docs system since it was not entirely clear how/if that overlapped with beta/preview states)
- Version
    - Only required for some admonitions (e.g. deprecated in X.Y.Z, ga in YYYY-MM-DD) and contexts (e.g. not necessary for serverless admonitions until/if we start versioning it)

[...]

PR #103, opened 2024-12-12
Add initial support for page level annotations for product support

Mpdreamz:


colleenmcginnis:
@lcawl do you think `ess` should be `hosted`?

Mpdreamz:
It should @colleenmcginnis I completely read over the `fka` in the new chart!

bmorelli25:
I talked with @colleenmcginnis about this today. A couple thoughts:

1. The shorthand of option one is convenient but we worry that it might be confusing to new or irregular contributors. Option two is more verbose, but perhaps that is actually a good thing as theres a higher likelyhood that folks can see the frontmatter and understand the 1:1 correlation between frontmatter and the badges that appear at the top of a page.

2. The other topic we discussed relates to this excerpt from issue #95 

> At this moment, we are proposing that the page-level metadata uses the same deployment type, lifecycle, and version dimensions as described in https://github.com/elastic/docs-builder/issues/94. The assumption being that what is specified at the page-level is applicable to the entire page unless it's overridden by admonitions at the section, page, or line level.

You mentioned that this is a first step towards page level annotations so I imagine further functionality will come in later PRs. Just wanted to make sure you saw this.

Mpdreamz:
Thanks @bmorelli25 

I updated the syntax to:

```
title: Elastic Docs v3
applies:
  stack: ga 8.1
  serverless: tech-preview
  hosted: beta 8.1.1
  eck: beta 3.0.2
  ece: unavailable
```

Based on your conversation with @colleenmcginnis. More details are in the PR description.

Mpdreamz:
This now also includes an initial html output for these front matter annotations:

image

PR #116, opened 2024-12-17
Update Server command

bmorelli25:
I can only get the server to kick on if I use the command in docs-builder-example.


Mpdreamz (approved):
LGMT!

Issue #106, opened 2024-12-17
[Migration tooling] Admonitions


bmorelli25:
### Summary

* Supported admonitions in Asciidoc: NOTE, TIP, IMPORTANT, CAUTION, WARNING
* Supported admonitions in V3: attention, caution, note, tip

Some context in https://github.com/elastic/docs-builder/issues/3.

### To do

Either map existing admonitions to new admonitions, or refine the list of supported admonitions.

### Conversion output
~~~
```{admonition} '#{title}'
:class: important

#{content}
```
~~~

Mpdreamz:
We explicitly don't/won't support MYST syntax which directly influences HTML. 

So we never use `:class:` to set the admonition type or influence any other directive for that matter.

We also explicitly don't support the open `{admonition}` and require the type to be specified as part of the directive name

````markdown
```{note}
```

```{caution}
```

```{tip}
```

```{attention}
```
````

Are the supported admonitions, adding support for others is relatively straightforward if needed.

bmorelli25:
Currently, admonitions are collapsed by default. We need to change the default behavior to be non-collapsible and open by default.

Mpdreamz:
Eek, that's a regression I introduced yesterday. Now the absence of `:open:` is parsed correctly again @bmorelli25

bmorelli25:
Supported admonitions should be `Note`, `Warning`, `Important`, and `Tip`. We currently support `note`, `caution`, `tip`, and `attention`. `Note` and `Tip` are okay as-is. `Caution` (yellow) should be renamed to `Warning`. `Attention` (red) should be renamed to `Important`.

bmorelli25:
All bugs are fixed and admonitions look good in migration script 👍

PR #117, opened 2024-12-19
Update Syntax Guide

bmorelli25:
### Summary

With the creation of [docs-builder-example](https://github.com/elastic/docs-builder-example) and https://elastic.github.io/docs-builder-example/, I think it makes sense to remove some of the noise from this repository. This PR:

* Removes example Observability, Semantic search, and Search labs content. This will still be viewable in [docs-builder-example](https://github.com/elastic/docs-builder-example).
* Removes out-of-date versioned content and nested content.
* Moves the automated settings docs into the markup guide with all other markup content.
* Renames the Markup guide to Syntax guide.
* Updates syntax documentation to better reflect current state and to call out differences between Asciidoc and V3 syntax. 

There's still more work to be done, but it might be worth merging these changes in their current state so we can continue building on this new structure.


bmorelli25:
It looks like req.md is required for some tests. I'll look more at this tomorrow.

bmorelli25:
okay, I learned a lot

src/Elastic.Markdown/_static/custom.css
```diff
@@ -15,6 +15,14 @@ This hides the ugly one */
   display: none !important;
 }
 
+.sd-tab-content {
+  padding-top: 20px !important;
+}
+
+body {
+  background-color: #FF0000;
```
Mpdreamz, line 23: This is screaming at me :smile_cat:

bmorelli25: Wait does that work for you? I can't get the CSS to work on my end and forgot about those. I'll update.

tests/Elastic.Markdown.Tests/SiteMap/NavigationTests.cs
```diff
@@ -18,12 +18,11 @@ public void ParsesNestedFoldersAndPrefixesPaths()
 	{
 		Configuration.ImplicitFolders.Should().NotBeNullOrEmpty();
 		Configuration.ImplicitFolders.Should()
-			.Contain("markup")
-			.And.Contain("elastic/observability");
+			.Contain("testing");
```
bmorelli25, line 21: I wanted to be able to test that `testing/req.md` exists as well, but I had trouble figuring out how to do that. After some trial and error, I learned that `Configuration.ImplicitFolders.Should()` only looks at folders (not files). Duh. That makes sense. I then tried doing `Configuration.ImplicitFolders.Files.Should()` but that didn't work. Is there a way to test for the existence of a specific file in an implicit folder? Maybe I'm thinking about this wrong and that's not something we want to test for though.

bmorelli25:
Some learnings from this work:

- Supported admonitions should be `Note`, `Warning`, `Important`, and `Tip`. We currently support `note`, `caution`, `tip`, and `attention`. `Note` and `Tip` are okay as-is. `Caution` (yellow) should be renamed to `Warning`. `Attention` (red) should be renamed to `Important`. I've added this same note to https://github.com/elastic/docs-builder/issues/106.
- There is a bug when multiple tab widgets appear on the same page. I've added this bug report and a related feature enhancement request to https://github.com/elastic/docs-builder/issues/119.
- It'd be nice to have a directive that doesn't try to parse or hide syntax. I'm currently documenting our syntax by escaping everything inside of a code block like this:
 
```
:::none
\`\`\`\{tip}
This is a tip.
\`\`\`
:::
```

The myst docs, for example, have a [`myst-example` directive](https://github.com/executablebooks/MyST-Parser/blame/master/docs/syntax/roles-and-directives.md#L51) that makes their syntax documentation easier to write (and cleaner).

tests/Elastic.Markdown.Tests/SiteMap/NavigationTests.cs
Mpdreamz: Yeah `ImplicitFolders` is not resolved until `DocumentationSet` is instantiated  at which point we use those folders to glob for `<implicit_folder>/*.md` 

The test was to ensure this list uses full relative paths e.g `a/b/c` not just `c`. 

Will amend the PR to update this.

bmorelli25: Ahhh okay that makes sense 👍

bmorelli25:
And now we know that your code block testing works 😅 

```
Error: Code block with annotations is not followed by a list
```

Mpdreamz:
> It'd be nice to have a directive that doesn't try to parse or hide syntax. 

We do! 😄 

```
:::none
\`\`\`\{tip}
This is a tip.
\`\`\`
:::
```

Should be:

````
```markdown
:::{tip}
This is a tip
:::
```
````

or better with 4 backticks:


`````
````markdown
```yaml
key: value
```
````
`````

I'm amending the PR now to update this (and adding the callout code docs)

**Comment by Mpdreamz** [comment]
Need to update the published docs-builder to not validate callouts that are in a nested code block 

e.g

`````markdown
```` javascript
```
let x = 1; <1>
```
````
`````

Should not be validated to be followed by a list and 

`````markdown
```` javascript
```
let x = 1; // comment
```
````
`````

Should not append a list magically either. 

https://github.com/elastic/docs-builder/pull/120 addresses this

Mpdreamz (approved):
LGTM!

bmorelli25:
Awesome! Thank you!

PR #124, opened 2025-01-04
Update supported admonition types

bmorelli25:
### Summary

Per https://github.com/elastic/docs-builder/issues/106#issuecomment-2555481597, supported admonition types should be `Note`, `Warning`, `Important`, and `Tip`. This list was determined after analyzing supported admonitions in our current Asciidoc system and cross-referencing with our style guide.

This PR removes support for `Caution` and `Attention` and replaces them with support for `Warning` and `Important`. It also adds a missing `AdmonitionUnsupportedTest`.

I've updated #3 to reflect these supported admonition types as that issue is referenced in the error logs.

### Results

Tested locally with:
```
dotnet publish "src/docs-builder/docs-builder.csproj" -c Release -o .artifacts/publish \
    --self-contained true /p:PublishTrimmed=true /p:PublishSingleFile=false /p:PublishAot=true -a arm64

./.artifacts/publish/docs-builder serve
```
Screenshot 2025-01-03 at 10 25 37 PM
Screenshot 2025-01-03 at 10 25 32 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# docs-builder. 
+# docs-builder.
 
-You've reached the home of the latest incarnation of the documentation tooling. 
+You've reached the home of the latest incarnation of the documentation tooling.
 
 This repository is host to:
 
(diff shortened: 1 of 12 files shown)


bmorelli25:
I'm not sure how to get the test to pass as it uses the docs-builder image.

Mpdreamz (approved):
Lgtm!

We'll have to switch over to validate the docs after we build using the artifacts to break the chicken and egg feedback loop. 


The current GitHub action we use is for everyone else.

PR #143, opened 2025-01-07
Add global substitutions support to Markdown processing

Mpdreamz:
This update introduces the ability to define global substitutions in `docset.yml`, which can be used across all files. It also enforces a rule to prevent redeclaration of global substitutions in a file's front matter. Tests and documentation have been updated to reflect and verify these changes.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -15,13 +15,6 @@ env:
   NUGET_PACKAGES: ${{ github.workspace }}/.nuget/packages
 
 jobs:
-  docs:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - name: Build documentation
-        uses: elastic/docs-builder@main
-  
   build:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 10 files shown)


Mpdreamz:
Docs are failing because we use our GitHub action not a build from the PR will update this too.

.github/workflows/pr.yml
```diff
@@ -15,13 +15,6 @@ env:
   NUGET_PACKAGES: ${{ github.workspace }}/.nuget/packages
 
 jobs:
-  docs:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - name: Build documentation
-        uses: elastic/docs-builder@main
```
reakaleek, line 23: Is it possible to use the local action?

```yaml
steps:
  - uses: actions/checkout@v4
  - uses: .
```

This would the use the action of the checked-out ref.

reakaleek (approved):

.github/workflows/pr.yml
Mpdreamz: it is not, our GitHub action uses a prebuilt container image that gets published from `main`. 

This ensures our Github Action is *ready to run* and no-one needs to wait for it to build the artifacts first. 

This ensures our GitHub actions runs in seconds for everyone else 😸

PR #146, opened 2025-01-07
docs: pin GitHub Actions version commit SHA in README

rwaight:
The purpose of this PR is to improve the baseline security for using GitHub Actions with the docs-builder; mainly to improve user awareness as they use the `README` from this `elastic/docs-builder` repo to deploy documentation using GitHub Actions.

This PR pins the GitHub actions to the commit SHA in the `README` example, with a comment including the version.
* For reference, see https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

This also adds notes to the example workflow with a link to the action in the GitHub Marketplace.

This is related to https://github.com/elastic/docs-builder/pull/147 and https://github.com/elastic/docs-builder-example/pull/12

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,10 @@ jobs:
   docs:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout the repo
+        # Verified creator: https://github.com/marketplace/actions/checkout
+        # GitHub Action for checking out a repo
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - name: Build documentation
         uses: elastic/docs-builder@main
 ```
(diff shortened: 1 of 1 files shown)


Mpdreamz:
I updated our readme and index.md documentation to use our GitHub action for deploying to GitHub instead which greatly simplifies the steps:

```yaml
environment:
  name: github-pages
  url: ${{ steps.deployment.outputs.page_url }}
steps:
  - uses: actions/checkout@v4
    
  - name: Publish Github
    uses: elastic/docs-builder/actions/publish@main
    id: deployment
```

This introduced the current conflict.

rwaight:
@Mpdreamz would you be okay with a new PR where the actions within `elastic/docs-builder/actions/publish` are pinned?

Mpdreamz:
@rwaight discussed this today with the team:

According to our info sec guidelines: 

https://elasticco.atlassian.net/wiki/spaces/EN/pages/48871919/Github+Workflows+Actions+-+Security+Best+Practice

It's indeed ok to depend on tags for trusted vendors (e.g GitHub and ourselves) :)

Since `elastic/docs-builder/actions/publish` only uses trusted vendors I rather rely on tags to ensure we get bug fixes. 


`elastic/docs-builder/actions/publish-vercel` is the only GitHub action we use that uses an external vendor so we should depend on SHA there 💯 agree. We would welcome an equivalent PR of this there!

Thanks for bringing this topic forward! :)

PR #147, opened 2025-01-07
ci: pin actions version commits and add verified creator notes

rwaight:
The purpose of this PR is to improve the baseline security for using GitHub Actions with the docs-builder; mainly to improve user awareness as they use this `elastic/docs-builder` repo to deploy documentation using GitHub Actions.

This PR pins the GitHub actions to the commit SHA, with a comment including the version.
* For reference, see https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

This also adds notes to the workflow with a link to the action in the GitHub Marketplace.

This is related to https://github.com/elastic/docs-builder/pull/146 and https://github.com/elastic/docs-builder-example/pull/12

### DIFF
--- a/.github/workflows/license.yml
+++ b/.github/workflows/license.yml
@@ -11,7 +11,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout the repo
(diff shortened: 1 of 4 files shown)


Mpdreamz:
As discussed here https://github.com/elastic/docs-builder/pull/146#issuecomment-2577583525 we are good to depend on tags for `GitHub` and `Elastic` published actions.

PR #148, opened 2025-01-07
Implement breadcrumbs fixes #138

### LINKED ISSUE #138: Breadcrumbs don't work
### Summary

Breadcrumbs always show up as `Elastic / Elastic Docs v3`

Image

### Expected

The breadcrumbs should update and change as you navigate throughout the docs.

Mpdreamz:


bmorelli25:
Looks good! Definitely good enough for MVP. A few thoughts, likely for a future enhancement issue:

* We should figure out the approach we want to take with deeply nested/long heading pages. I think a scroll is fine for now, but we likely want to do something different in the long run.
 
Screenshot 2025-01-07 at 12 13 19 PM

* We probably want to use `navigation_title` for both the TOC and the Breadcrumbs. This will greatly reduce the length of some of our longer breadcrumb sequences and allow us to prioritize SEO repercussions in page `title`s. It also creates a visual mapping between breadcrumb/TOC heading.

* A lot of breadcrumbs omit the current page you're on. This reduces duplication and further shortens long breadcrumb sequences. For example, instead of this:
  Screenshot 2025-01-07 at 12 20 30 PM
  
  We can do this:
  Screenshot 2025-01-07 at 12 20 30 PM

Mpdreamz:
Those are great points and except for the first one quick wins, will amend in the morning!

reakaleek (approved):

reakaleek:
Do you think a test for this could be beneficial?

Mpdreamz:
> Do you think a test for this could be beneficial?

yes! good call calling out my laziness 😸

tests/Elastic.Markdown.Tests/SiteMap/BreadCrumbTests.cs
```diff
@@ -0,0 +1,25 @@
...
+	[Fact]
+	public void ParsesATableOfContents()
+	{
+		var doc = Generator.DocumentationSet.Files.FirstOrDefault(f => f.RelativePath == "testing/nested/index.md") as MarkdownFile;
+
+		doc.Should().NotBeNull();
+
+		doc!.Parent.Should().NotBeNull();
+
+		doc.YieldParents().Should().HaveCount(2);
```
reakaleek, line 22: nit: I think another critical path worth testing is if the page has `0` parents. Nevertheless, thank you for adding a test!

reakaleek (approved):

Issue #139, opened 2025-01-07
Nested TOC bug


bmorelli25:
### Summary

Clicking on a nested page in the TOC causes the TOC to collapse when the page loads:

https://github.com/user-attachments/assets/7d00634d-2636-4610-8197-68cf592eb958

### Expected behavior

The TOC should remain open and uncollapsed when viewing a nested page

reakaleek:
ref:
- https://github.com/elastic/docs-builder/pull/149

PR #149, opened 2025-01-07
Fix collapsing toc when clicking on a nested page

reakaleek:
Closes https://github.com/elastic/docs-builder/issues/139

## Changes

 * Expands all collapsed parent sections in the navigation tree that contain the current page link.
 * This ensures the current page is visible in the navigation by opening all ancestor collapsible sections.
 * The function works by:
    1. Finding the navigation link matching the current page URL
    2. Walking up the DOM tree from that link
    3. Expanding any collapsed parent sections by toggling their CSS classes

https://github.com/user-attachments/assets/75351ec1-0e29-4ee8-bbc1-866ca4a82830

## Notes

This is a hacky fix since the source code for this functionality is minified and just copied here. 
Doing this properly should be handled after the next major release.


src/Elastic.Markdown/_static/shibuya.js
```diff
@@ -14,3 +14,28 @@ iconify-icon/dist/iconify-icon.mjs:
...
+      if (parent.classList && parent.classList.contains('_collapse')) {
+        parent.classList.remove('_collapse');
+        parent.classList.add('_expand');
+      }
+      parent = parent.parentElement;
+    }
+  }
+}
+
+expandCurrentPage();
```
reakaleek, line 41: It seems like the GH UI does not render this due to its content. 

Add a comment here to see the diff.

Mpdreamz (approved):
LGMT! 💯 

Once we get new design we can focus on the css an javascript from scratch. We currently include way too much.

Issue #125, opened 2025-01-04
Anchor tag bugs


bmorelli25:
### Bug 1

**What happens**
Anchor tags are only generated for H1 and H2 level headings.  

**What I expect to happen**
Anchor tags should be generated for all heading levels. 

### Bug 2

**What happens**
 If a heading includes backticks, the ID is not unique or correct. For example, the heading ``## `note` `` creates the following anchor: `markdig.syntax.inlines.codeinline`. This also impacts the OTP (on this page) widget on the right side of the page:

<img width="335" alt="Image" src="https://github.com/user-attachments/assets/6590e25b-f865-420c-ab2a-f25d4b52fbe0" />

**What I expect to happen**
Backticks should be ignored and a unique anchor tag should be created.

### Bug 3

**What happens**
Linking to valid anchor tags on a different page fails the link checker. For example:

```
[Notes](tables.md#notes)
```

```
Error: `notes` does not exist in .
    ┌─[/github/workspace/docs/source/syntax/links.md]
    │
 [35](https://github.com/elastic/docs-builder/actions/runs/12638318856/job/35214323557?pr=129#step:4:36) │ I link to the [Notes](tables.md#notes) heading on the [Tables](tables.md) page.
    ·              ───────────────                                                   
    ·                                                                                 
    │
    └─
	1 Errors / 0 Warnings

:: info: Program[0] generate :: Finished in '00:00:00.0220[37](https://github.com/elastic/docs-builder/actions/runs/12638318856/job/35214323557?pr=129#step:4:38)5
```

**What I expect to happen**
The link checker should not fail on valid anchor links.

bmorelli25:
Another bug:

Linking to headings across pages doesn't work. For example, this link:

```
I link to the [Notes](tables.html#notes) heading on the [Tables](tables.html) page.
```

Takes users to:

```
http://localhost:5000/syntax/tables.html#notes#notes
```

bmorelli25:
Never mind on the above comment. I was using `.html` instead of `.md`. The original bug still stands.

bmorelli25:
Update to the update:

The link renders when you use `.md`, but the link checker fails:

```
[Notes](tables.md#notes)
```

```
Error: `notes` does not exist in .
    ┌─[/github/workspace/docs/source/syntax/links.md]
    │
[...]

PR #153, opened 2025-01-08
Add support for custom heading anchors in Markdown.

Mpdreamz:
Custom heading anchors allow more control over anchor names in generated HTML. Inline annotations can now define specific anchors, and normalization ensures consistent formatting. Updated relevant tests, documentation, and rendering logic to support this feature.

You can now override the autogenerated anchors for headings with 

```markdown
## My header [my-completely-different-anchor]
```

Updated the syntax guide to document this new feature.

This also addresses the bugs reported in #125

### DIFF
--- a/docs/source/syntax/links.md
+++ b/docs/source/syntax/links.md
@@ -37,7 +37,7 @@ I link to the [Notes](tables.md#notes) heading on the [Tables](tables.md) page.
 Headings will automatically create anchor links in the resulting html. 
 
 ```markdown
-## This Is An Header
+## This Is A Header
 ```
 
 Will have an anchor link injected with the name `this-is-an-header`.
@@ -46,7 +46,7 @@ Will have an anchor link injected with the name `this-is-an-header`.
 If you need more control over the anchor name you may specify it inline
 
 ```markdown
-## This Is An Header [but-this-is-my-anchor]
+## This Is A Header [but-this-is-my-anchor]
 ```
 
 Will result in an anchor link named `but-this-my-anchor` to be injected instead. Do note that these inline anchors will be normalized.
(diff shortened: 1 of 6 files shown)


reakaleek (commented):
What do you think about using curly braces instead of square brackets?

I think curly braces are less likely to be part of a heading. 

Kramdown has a similar approach.

They use this:

```md
Hello        {#id}
-----

# Hello      {#id}

# Hello #    {#id}
````

Source: https://kramdown.gettalong.org/syntax.html

IMO, Kramdown also cleverly utilizes the `#`, which makes it even less likely to be part of the heading.

**Comment by Mpdreamz** [comment]
I personally prefer `[ ]` since it's the syntax for links and will highlight quite nicely in editors. 

It will also be easier to disambiguate variable substitutions and these anchors. 


A more technical reason I'm not thrilled about brackets is that markdig has an extension to attach arbitrary data to elements using brackets at the end. 

We don't use that today but we might in the future. Want to leave that door open

**Comment by reakaleek** [comment]
Understood. 

What do you think about the `#` as in `[#id]`.

Semantically, this also resembles an URL fragment, making it clearer that this will be the ID of the element.

File: docs/source/syntax/links.md [hunk 1]
```diff
@@ -22,12 +22,39 @@ A link contains link text (the visible text) and a link destination (the URI tha
...
-```
\ No newline at end of file
+```
+
+## Heading anchors
+
+Headings will automatically create anchor links in the resulting html. 
+
+```markdown
+## This Is An Header
```
reakaleek, line 40:
```suggestion
## This Is A Header
```

docs/source/syntax/links.md
```diff
@@ -22,12 +22,39 @@ A link contains link text (the visible text) and a link destination (the URI tha
...
+```markdown
+## This Is An Header [but-this-is-my-anchor]
+```
+
+Will result in an anchor link named `but-this-my-anchor` to be injected instead. Do note that these inline anchors will be normalized.
+
+Meaning 
+
+```markdown
+## This Is An Header [What about this for an anchor!]
```
reakaleek, line 57:
```suggestion
## This Is A Header [What about this for an anchor!]
```

Mpdreamz:
I've updated the docs to use `[#anchor]`

reakaleek:
SGTM. Thank you

PR #155, opened 2025-01-08
Fix an edgecase where substitutions could not track the delimiter properly when an empty line preceded it and followed it

Mpdreamz:


Mpdreamz:
Self reviewing this too @reakaleek to unblock migration.

I won't make a habit of this I swear ! 😸

Issue #156, opened 2025-01-08
Unhandled exception when linking to localhost


bmorelli25:
The following steps (1+2) renders correctly:

```
  Follow these instructions to get started with docs-builder on your machine.
  
  ::::{tab-set}
  
  :::{tab-item} macOS
  
  ### macOS Installation
  
  1. **Download the Binary:**
     Download the latest macOS binary from [releases](https://github.com/elastic/docs-builder/releases/latest/):
     ```sh
     curl -LO https://github.com/elastic/docs-builder/releases/latest/download/docs-builder-mac-arm64.zip
     ```
  
  2. **Extract the Binary:**
     Unzip the downloaded file:
     ```sh
     unzip docs-builder-mac-arm64.zip
     ```
  
  :::
  
  ::::
```

Adding a third step which consists of a link to localhost breaks the docs build:

```
  Follow these instructions to get started with docs-builder on your machine.
  
  ::::{tab-set}
  
  :::{tab-item} macOS
  
  ### macOS Installation
  
  1. **Download the Binary:**
     Download the latest macOS binary from [releases](https://github.com/elastic/docs-builder/releases/latest/):
     ```sh
     curl -LO https://github.com/elastic/docs-builder/releases/latest/download/docs-builder-mac-arm64.zip
     ```
  
  2. **Extract the Binary:**
     Unzip the downloaded file:
     ```sh
     unzip docs-builder-mac-arm64.zip
     ```
  
  3. **Run the Binary:**
     Use the `serve` command to start serving the documentation at [http://localhost:5000](http://localhost:5000):
     ```sh
     ./docs-builder-mac-arm64 serve
     ```
  
  :::
  
  ::::
```

The error:

```
[...]

PR #157, opened 2025-01-09
Fix links to localhost

reakaleek:
Closes https://github.com/elastic/docs-builder/issues/156

### DIFF
--- a/src/Elastic.Markdown/Myst/InlineParsers/DiagnosticLinkInlineParser.cs
+++ b/src/Elastic.Markdown/Myst/InlineParsers/DiagnosticLinkInlineParser.cs
@@ -60,7 +60,7 @@ public class DiagnosticLinkInlineParser : LinkInlineParser
 
 		if (Uri.TryCreate(url, UriKind.Absolute, out var uri) && uri.Scheme.StartsWith("http"))
 		{
-			var baseDomain = string.Join('.', uri.Host.Split('.')[^2..]);
+			var baseDomain = uri.Host == "localhost" ? "localhost" : string.Join('.', uri.Host.Split('.')[^2..]);
 			if (!context.Configuration.ExternalLinkHosts.Contains(baseDomain))
 			{
 				processor.EmitWarning(
--- a/src/Elastic.Markdown/IO/ConfigurationFile.cs
+++ b/src/Elastic.Markdown/IO/ConfigurationFile.cs
@@ -23,7 +23,7 @@ public record ConfigurationFile : DocumentationFile
 	public HashSet<string> Files { get; } = new(StringComparer.OrdinalIgnoreCase);
 	public HashSet<string> ImplicitFolders { get; } = new(StringComparer.OrdinalIgnoreCase);
 	public Glob[] Globs { get; } = [];
-	public HashSet<string> ExternalLinkHosts { get; } = new(StringComparer.OrdinalIgnoreCase) { "elastic.co", "github.com", };
+	public HashSet<string> ExternalLinkHosts { get; } = new(StringComparer.OrdinalIgnoreCase) { "elastic.co", "github.com", "localhost", };
 
 	private readonly Dictionary<string, string> _substitutions = new(StringComparer.OrdinalIgnoreCase);
 	public IReadOnlyDictionary<string, string> Substitutions => _substitutions;


src/Elastic.Markdown/Myst/InlineParsers/DiagnosticLinkInlineParser.cs
```diff
@@ -60,7 +60,7 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
 
 		if (Uri.TryCreate(url, UriKind.Absolute, out var uri) && uri.Scheme.StartsWith("http"))
 		{
-			var baseDomain = string.Join('.', uri.Host.Split('.')[^2..]);
+			var baseDomain = uri.Host.StartsWith("localhost") ? "localhost" : string.Join('.', uri.Host.Split('.')[^2..]);
```
reakaleek, line 63:
```suggestion
			var baseDomain = uri.Host == "localhost" : string.Join('.', uri.Host.Split('.')[^2..]);
```

src/Elastic.Markdown/Myst/InlineParsers/DiagnosticLinkInlineParser.cs
```diff
@@ -60,7 +60,7 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
 
 		if (Uri.TryCreate(url, UriKind.Absolute, out var uri) && uri.Scheme.StartsWith("http"))
 		{
-			var baseDomain = string.Join('.', uri.Host.Split('.')[^2..]);
+			var baseDomain = uri.Host == "localhost" ? "localhost" : string.Join('.', uri.Host.Split('.')[^2..]);
```
reakaleek, line 63: I need your help here @Mpdreamz.

This fixes the error. However, I would have expected that it emits a warning because "localhost" is not defined in the `external_hosts` configuration. But it does not.

reakaleek: I found this https://github.com/elastic/docs-builder/blob/main/src/Elastic.Markdown/IO/ConfigurationFile.cs#L26 

and I would have expected that I need to add `localhost` here. But it's not needed.

Mpdreamz: Interesting I am getting: 

```
Warning: External URI 'http://localhost:8080/' is not allowed. Add 'localhost' to the 'external_hosts' list in /Users/mpdreamz/Projects/docs-builder/docs/source/docset.yml to allow links to this domain.
    ┌─[/Users/mpdreamz/Projects/docs-builder/docs/source/testing/req.md]
```

You might hit a chicken and egg where you ran docs-builder before and after the fix and the file is not marked as invalid and thus incremental compile is not re-evaluating it.


```
:: info: DocumentationGenerator[0] Re-evaluating /Users/mpdreamz/Projects/docs-builder/docs/source/testing/req.md
```

If you clear the output directory `rm -rf .artifacts/docs/html` or run the tool with the `--force` flag do you see the error message too?

reakaleek: Ah!

I only get the warning when I run `dotnet run`

If I run `dotnet run -- serve` it doesn't show the warning.

This could be worth a follow issue, WDYT?

reakaleek: Alright, I added `localhost` to the default list https://github.com/elastic/docs-builder/pull/157/commits/021ab63bfe64c5fa9ca5a717e2b5212ac70c469a

Mpdreamz: So with the `serve` command I want to surface the errors in the browser itself rather than having folks needing to dig through the CLI :) 

We'll need a issue for that though

reakaleek: I think we can draw a lot of inspiration from existing frameworks with live reload functionality (e.g. nextjs)

And I think they actually do both. Showing the error in the console as well as in the browser.

But we can definitely discuss it in the issue.

Mpdreamz (approved):

src/Elastic.Markdown/Myst/InlineParsers/DiagnosticLinkInlineParser.cs
reakaleek: Follow-up issue: https://github.com/elastic/docs-builder/issues/158

Issue #169, opened 2025-01-10
"Not linked in toc:" shouldn't fail on files included with `include` directive


bmorelli25:
### Summary

I have a file named `output.md` that is the output of a script. I do not want to include this file in the TOC. Instead, I include this file in a different markdown file using the include directive:

```
:::{include} scripts/codeowner-gen/output.md
:::
```

However, the build gives this `fail` because `output.md` is not included in the TOC:

```
:: fail: Documentation.Builder.Log[0] Not linked in toc: migration/scripts/codeowner-gen/output.md (/Users/brandon/Documents/GitHub/docs-builder/docs/source/docset.yml:)
:: info: DocumentationGenerator[0] Created documentation set for: /Users/brandon/Documents/GitHub/docs-builder/docs/source
:: info: DocumentationGenerator[0] Source directory: /Users/brandon/Documents/GitHub/docs-builder/docs/source Exists: True
:: info: DocumentationGenerator[0] Output directory: /Users/brandon/Documents/GitHub/docs-builder/.artifacts/docs/html Exists: True
```

Mpdreamz:
Two notes:

I think it would make senses to enforce includes come from `_include` folders. These would automatically be exempt from the TOC check. On top of that their usage will be immediately obvious. These `_include` folders may appear anywhere in the tree so they can be scoped logically. 


Files can be excluded through globbing today e.g this PR that turns the root of `docs-content` into a docset but excludes `README.md` from participating in the docs build.

 https://github.com/elastic/docs-content/blob/feature/move-to-docs-builder/docset.yml#L4

Mpdreamz:
another benefit of enforcing includes to live in `_include` folders is that we can much easier assume the role of a markdown file programmatically and e.g warn of include files that are not actually included anywhere.

bmorelli25:
Makes sense and ++ to implementing this

PR #170, opened 2025-01-10
Add to Migration, Syntax, and Contribution Guides

bmorelli25:
### Summary

* Adds to migration, syntax, and contribution guides
* Adds issue templates for bugs and enhancements
* Adds scripts helpful for migration tasks


docs/source/migration/scripts/attribute-converter/attr.py
```diff
@@ -0,0 +1,59 @@
+import re
```
bmorelli25, line 1: @elastic/docs-engineering I wonder if there's somewhere better to put these scripts instead of in the docs dir?

bmorelli25:
The build is currently failing with

```
Error: Not linked in toc: migration/scripts/codeowner-gen/output.md
```

Opened a bug for this: https://github.com/elastic/docs-builder/issues/169

docs/source/migration/scripts/attribute-converter/attr.py
Mpdreamz: @bmorelli25 can you describe a bit what the intended use case is for these ?

Who runs them and when? is this a continuous thing?

bmorelli25: Good question. They're documented in the source files in this PR. But a brief summary is:

1. **CODEOWNER list for documentation freeze**
  This script creates a list of repositories and the directories in those repositories that the `@docs-freeze-team` (just @KOTungseth) should be added as codeowner of. During a documentation freeze, the `@docs-freeze-team` must approve any pull requests that modify `.asciidoc` files before they can be merged. This prevents folks from merging docs changes during the freeze.

1. **Attribute converter**
  This script converts the attributes in https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc to the supported yaml format for docs-builder.

1. **HTML to Markdown Tab Converter**
  The tab converter script is used to convert HTML tabs into Markdown tabs within documentation files. The script scans through specified directories, finds all Markdown (`.md`) files, and replaces any HTML-based tab structures with their Markdown equivalents.<br>  
  _This is no longer necessary as @siamakp-elastic added this capability to the migration tool. I will remove this._

Mpdreamz: 1. Is this something that needs to run continuously in the future or is this a one of thing? 


2. Same question really, but also:
   * Do we need all of these in the new architecture?
   * Rather then having these live in all of the isolated `docset.yml` files should `docs-builder` natively support them?

bmorelli25: > Is this something that needs to run continuously in the future or is this a one of thing?

Okay so after a meeting today, the CODEOWNER script won't be used so we can get rid of it. We will need to rerun the attribute converter eventually (Jan 29) but I can do that so there's no need to put it here (won't be used after that date). I'll remove both from this PR.

> Do we need all of these in the new architecture?

All, no. Most, yes.

> Rather then having these live in all of the isolated docset.yml files should docs-builder natively support them?

Yes. The more I play with this, the more a single yml file where these can be defined would be ideal.

bmorelli25:
As discussed above, I removed the scripts from this PR. That also fixed the build error. I'm going to get this merged so that the GH pages is updated.

PR #176, opened 2025-01-13
Upload links.json to s3

reakaleek:
## Context

We want to upload the links.json file to the Link Index (s3 bucket) for later validation.

## Changes

This adds the `--upload-to-link-index` option, which uploads the links.json file to the s3 bucket.

It only happens under certain conditions, e.g.:

- when it's running on main
- if it's can derive the repository name from the `GITHUB_REPOSITORY` environment variable

### DIFF
--- a/src/docs-builder/Cli/Commands.cs
+++ b/src/docs-builder/Cli/Commands.cs
@@ -4,16 +4,16 @@
 using System.IO.Abstractions;
 using Actions.Core.Services;
 using ConsoleAppFramework;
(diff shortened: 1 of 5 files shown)


Mpdreamz (changes requested):
Codewise this looks great!

I do think this would be better off in an isolated binary. Purely to keep dependencies in `docs-builder` itself light.

reakaleek:
We decided to use the aws cli in the github actions workflow for now.

PR #187, opened 2025-01-14
Fix tab groups

### LINKED ISSUE #119: Multiple tab groups on one page don't work
Bug:
Multiple tab groups on one page don't work. The first tab group works as expected. If you try to change the tab in the second group, the first tab group changes instead

Feature request:
A common use case for tabs in our documentation is installation instructions (https://www.elastic.co/guide/en/observability/current/get-started-with-apm-server-binary.html). In these cases, it makes sense to have the concept of related tabs. This means if a user switches from macOS to linux, for example, all related tab groups on the page will also change to the linux tab.

reakaleek:
Closes #119 

## Syntax

See tabs.md in this PR

## Fixed Behaviour

https://github.com/user-attachments/assets/50cf11dd-fe66-45b6-a660-efabf72b669e


bmorelli25:
Tested and works great! So much easier to use than our asciidoc tab syntax too.

PR #208, opened 2025-01-15
Limit diagnostics output and enhance logging clarity

Mpdreamz:
Introduced pagination for diagnostics console output to handle only the first 100 errors/warnings and added clear messages for truncated results. Improved logging consistency in various components and added null safety when reading Git-related data.

This PR also caught a performance regression in the handling of codeblocks. It was too eagerly trying to attempt substitutions on each line.

### DIFF
--- a/src/Elastic.Markdown/Diagnostics/DiagnosticsChannel.cs
+++ b/src/Elastic.Markdown/Diagnostics/DiagnosticsChannel.cs
@@ -124,6 +124,7 @@ public class DiagnosticsCollector(IReadOnlyCollection<IDiagnosticsOutput> output
 
 	public virtual async Task StopAsync(CancellationToken cancellationToken)
 	{
+		Console.WriteLine("Stopping diagnostics");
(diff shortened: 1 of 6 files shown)


reakaleek (approved):

src/Elastic.Markdown/Diagnostics/DiagnosticsChannel.cs
```diff
@@ -124,6 +124,7 @@ protected virtual void HandleItem(Diagnostic diagnostic) { }
 
 	public virtual async Task StopAsync(CancellationToken cancellationToken)
 	{
+		Console.WriteLine("Stopping diagnostics");
```
reakaleek, line 127: Should this be a logger?

Or is it a left over from debugging?

Mpdreamz: Indeed a leftover 😢 thanks for catching it ammended the PR.

PR #210, opened 2025-01-15
Add support for 'continue-on-error' input in publish action

Mpdreamz:
This adds an optional 'continue-on-error' input to the publish action. It allows the workflow to proceed even if documentation building fails, based on the provided value. Defaults to 'false' to maintain the current behavior.

### DIFF
--- a/actions/publish/action.yml
+++ b/actions/publish/action.yml
@@ -14,6 +14,10 @@ inputs:
   strict:
     description: 'Treat warnings as errors'
     required: false
+  continue-on-error:
+    description: 'Treat warnings as errors'
+    required: false
+    default: "false"
 
 runs:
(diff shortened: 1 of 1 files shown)


reakaleek (approved):
Can you elaborate a bit more on this?

if the build fails, what will be deployed to pages?

reakaleek:
> > if the build fails, what will be deployed to pages?
> 
> We will always convert all the pages to the output, so we'll always deploy a complete docset. If this option is set it might not be one that validates though.

Thanks for the explanation!

Mpdreamz:
This is a temporary feature while folks are migrating over. Will kill this feature afterwards since we don't want bad `links.json` to be published 💥

PR #265, opened 2025-01-17
Change default serve port to 3000 and add ability to override

reakaleek:


reakaleek:
I understand that changing the default port can introduce some disruption.

But it's a very common port for local web development. Hence, I still wanted to suggest it.

I can also revert that.

reakaleek:
Also, I'm afraid this affects the docker image. I couldn't find how it serves on port 8080

src/docs-builder/Http/DocumentationWebHost.cs
```diff
@@ -53,7 +54,7 @@ public DocumentationWebHost(string? path, ILoggerFactory logger, IFileSystem fil
 		builder.Services.AddHostedService<ReloadGeneratorService>();
 
 		//builder.Services.AddSingleton(logger);
-
+		builder.WebHost.UseUrls($"http://localhost:{port}");
```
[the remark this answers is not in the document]
reakaleek: Thank you for the hint!

https://github.com/elastic/docs-builder/pull/265/commits/fe72d13ddc000321c6567504e8d615e924e019aa

src/docs-builder/Http/DocumentationWebHost.cs
```diff
@@ -54,7 +54,13 @@ public DocumentationWebHost(string? path, int port, ILoggerFactory logger, IFile
 		builder.Services.AddHostedService<ReloadGeneratorService>();
 
 		//builder.Services.AddSingleton(logger);
-		builder.WebHost.UseUrls($"http://localhost:{port}");
+
+		// See https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#dotnet_running_in_container-and-dotnet_running_in_containers
+		// This way use the default port when running in a container
+		if (string.IsNullOrEmpty(Environment.GetEnvironmentVariable("DOTNET_RUNNING_IN_CONTAINER")) &&
```
Mpdreamz, line 60: Just checking one will suffice 😸 both will be set (don't ask my why!).

reakaleek: Okay 😅 

https://github.com/elastic/docs-builder/pull/265/commits/c9b04bba8dde4c3fe953e0093dbc730ecfd68848

reakaleek: I'm starting to wonder.

IIUC, the main reason for the differing ports for the native binary and the docker image are dotnet defaults.

Would it be a better UX if both use the same port?

Mpdreamz: https://learn.microsoft.com/en-us/dotnet/core/compatibility/containers/8.0/aspnet-port has some background as to why its 8080. 

In .net the default changes based on environments: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-9.0

Which is not something we leverage.  Docker is a `prod` environments hence the different port as the rest of the run environments which are `dev` by default..

It makes sense to normalize this to one non privileged port and forego the docker env check to keep it simple.

reakaleek: Do I understand correctly? We skip the environment check and always do `UseUrls`?

Mpdreamz: yes thats the less long winded conclusion 😸  

We'll have to update https://github.com/elastic/docs-builder-example or maybe we should archive that repository now that we have the migration repositories.

reakaleek: https://github.com/elastic/docs-builder/pull/265/commits/281e30ec2d8b0e2dc48c2dcc3e384401a197ecdc

Mpdreamz (approved):

PR #274, opened 2025-01-20
Add support for extensionless URLs (pretty URLs)

reakaleek:
## Details
This will add support for extensionless URLs.

### `serve` command

E.g http://localhost:5000/migration/syntax.html becomes http://localhost:5000/migration/syntax

However, using `.html` extensions still works for backward compatibility.

### `generate` command

In the generate command, we change the file and folder structure to always have an index.html file for every page.

e.e `/migration/syntax.html` becomes `/migration/syntax/index.html`.

This way, we can strip the `/index.html` segment from the URL and still serve the correct file.

an example of this is deployed at https://dwnz7p9ulv07a.cloudfront.net/migration/versioning


src/Elastic.Markdown/Slices/HtmlWriter.cs
```diff
@@ -77,7 +77,26 @@ public async Task WriteAsync(IFileInfo outputFile, MarkdownFile markdown, Cancel
 			outputFile.Directory.Create();
 
 		var rendered = await RenderLayout(markdown, ctx);
-		var path = Path.ChangeExtension(outputFile.FullName, ".html");
+		// var path = Path.ChangeExtension(outputFile.FullName, ".html");
```
reakaleek, line 80:
```suggestion
```

Mpdreamz:
> In the generate command, we change the file and folder structure to always have an index.html file for every page.

> e.e /migration/syntax.html becomes /migration/syntax/index.html.

I am not sure this aids in keeping the docs simple for new contributors. 
We kind of lose that 1:1 mapping of URL to file: turning `/migration/syntax.md` to `/migration/syntax/index.html` 

Don't really have a full fledged opinion on this (yet) though need to mull it over a bit ⌛

reakaleek:
> > In the generate command, we change the file and folder structure to always have an index.html file for every page.
> 
> > e.e /migration/syntax.html becomes /migration/syntax/index.html.
> 
> I am not sure this aids in keeping the docs simple for new contributors. We kind of lose that 1:1 mapping of URL to file: turning `/migration/syntax.md` to `/migration/syntax/index.html`
> 
> Don't really have a full fledged opinion on this (yet) though need to mull it over a bit ⌛

I was actually highly inspired by https://middlemanapp.com/advanced/pretty-urls/

bmorelli25:
Related to https://github.com/elastic/docs-projects/issues/329

Mpdreamz:
This LGTM for the most part! 

I think we should make `MarkdownFile` as the index property through an `IsIndex` prop on the class.

We now have quite a few places where we handle `index.md` specially through a magic string however 

```yaml
- folder: mover
  children:
    - file: first-page.md
    - file: second-page.md
```

Here `first-page.md` is the index file for the `mover` folder.

Mpdreamz (approved):
LGTM, we can address my feedback in a follow up PR.

reakaleek:
This should wait for https://github.com/elastic/docs-builder/pull/442 before being merged.

PR #276, opened 2025-01-21
Advertise docs-builder updates when running from command line

Mpdreamz:
- Advertise docs-builder updates when running from the command line
- Advertise at the end of command invocation

image


reakaleek (approved):
LGTM.

I think the cherry on top would be a link to upgrade/install instructions

Mpdreamz:
> I think the cherry on top would be a link to upgrade/install instructions

Added!

PR #278, opened 2025-01-21
Refactor TOC processing to handle excluded and unreachable files

Mpdreamz:
Updated the TOC processing logic to emit errors for excluded or unreachable files and ensure valid navigation indexes. These changes improve diagnostics during documentation builds and prevent invalid entries in the output structure.

Markdown files who's parent matches an exclusion glob never receive a `NavigationIndex` since they are effectively unreachable. 

This caused issues in https://github.com/elastic/elasticsearch.md as can be seen by this [workflow run](https://github.com/elastic/elasticsearch.md/actions/runs/12857035309/job/35844425339)

Since several migrated files now start with an underscore but `docset.yml` includes the following:

```
exclude:
  - '_*.md'
```

Yet the TOC still references files that start with an underscore. We now emit a better error message for this:

image

Another issue is that children of these excluded files never receive a `NavigationIndex` and thus defaulted to `0`. This blew up later when we create an index by `NavigationIndex`.

```yaml
  - file: _data_store_architecture.md
    children:
      - file: nodes-shards.md
      - file: node-roles-overview.md
      - file: docs-replication.md
      - file: shard-allocation-relocation-recovery.md
        children:
          - file: shard-allocation-awareness.md
```

These are now also emitted as configuration errors properly:

image

And we no longer crash the build, we recover and build all files again if this does occur.

### DIFF
--- a/src/docs-assembler/conf.yml
+++ b/src/docs-assembler/conf.yml
@@ -1,4 +1,14 @@
+
 repos:
+  docs-content:
+    repo: "git@github.com:elastic/apm-aws-lambda.git"
+  docs-builder:
+    repo: "git@github.com:elastic/apm-aws-lambda.git"
+  docs-builder:
+    repo: "git@github.com:elastic/apm-aws-lambda.git"
+
+
+repos-old:
   apm-aws-lambda:
     repo: "git@github.com:elastic/apm-aws-lambda.git"
   apm-k8s-attacher:
--- a/src/docs-assembler/conf.yml
+++ b/src/docs-assembler/conf.yml
@@ -1,14 +1,4 @@
-
 repos:
-  docs-content:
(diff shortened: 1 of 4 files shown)


src/docs-assembler/conf.yml
```diff
@@ -1,4 +1,14 @@
...
 repos:
+  docs-content:
+    repo: "git@github.com:elastic/apm-aws-lambda.git"
+  docs-builder:
+    repo: "git@github.com:elastic/apm-aws-lambda.git"
+  docs-builder:
+    repo: "git@github.com:elastic/apm-aws-lambda.git"
+
+
+repos-old:
```
reakaleek: Was this added intentionally?

Mpdreamz: Nope removed it as you commented 😄

PR #279, opened 2025-01-21
Add preview workflow

reakaleek:
## Details

Add deployment of preview environments on a PR

## Follow-ups
- Deploy static preview for main branch
- Create similar action for consumers

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -35,18 +35,11 @@ jobs:
 
       - name: Publish AOT
         run: ./build.sh publishbinaries
-
-      # we run our artifact directly please use the prebuild
-      # elastic/docs-builder@main GitHub Action for all other repositories!
-      - name: Build documentation
-        env:
-          PR_NUMBER: ${{ github.event.pull_request.number }}
-        run: .artifacts/publish/docs-builder/release/docs-builder --strict --path-prefix "/${GITHUB_REPOSITORY}/pull/${PR_NUMBER}"
-      
+        
       - uses: actions/upload-artifact@v4
         with:
-          name: docs
-          path: .artifacts/docs/html
+          name: docs-builder-binary
+          path: .artifacts/publish/docs-builder/release/docs-builder
           if-no-files-found: error
           retention-days: 1
 
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -6,6 +6,9 @@ on:
 permissions:
   contents: read
   packages: read
+  id-token: write
+  pull-requests: write
+  deployments: write
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -36,4 +39,17 @@ jobs:
       # we run our artifact directly please use the prebuild
       # elastic/docs-builder@main GitHub Action for all other repositories!
       - name: Build documentation
-        run: .artifacts/publish/docs-builder/release/docs-builder --strict
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: .artifacts/publish/docs-builder/release/docs-builder --strict --path-prefix "/${GITHUB_REPOSITORY}/pull/${PR_NUMBER}"
+      
+      - uses: actions/upload-artifact@v4
+        with:
+          name: docs
+          path: .artifacts/docs/html
+          if-no-files-found: error
+          retention-days: 1
+
+  preview:
+    needs: build
+    uses: ./.github/workflows/preview.yml
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,0 +1,85 @@
+name: preview
+
+on:
+  workflow_call: ~
+
+permissions:
+  id-token: write
+  pull-requests: write
+  deployments: write
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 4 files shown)


.github/workflows/preview.yml
```diff
@@ -0,0 +1,56 @@
...
+        id: bootstrap
+        uses: ./.github/actions/bootstrap
+
+      - name: Publish AOT
+        run: ./build.sh publishbinaries
+
+      # we run our artifact directly please use the prebuild
+      # elastic/docs-builder@main GitHub Action for all other repositories!
+      - name: Build documentation
+        run: .artifacts/publish/docs-builder/release/docs-builder --strict
```
Mpdreamz, line 29: Should we cache this from the `pr.yml` build and then move the general docs build as part of this workflow ?

In other words we make this depend on the pr workflow.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Generate AWS Role ARN
```
Mpdreamz, line 30: I think we also want to extend our `publish` github action to include this.

Mpdreamz (commented):
Some general comments (realizing this is a draft).

.github/workflows/preview.yml
```diff
@@ -0,0 +1,108 @@
...
+        id: deployment
+        with:
+          result-encoding: string
+          script: |
+            const response = await github.rest.repos.createDeployment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                ref: "${{ github.event.pull_request.head.ref }}",
+                environment: `preview-${context.issue.number}`,
```
bmorelli25, line 29: Should this be preview-{repo}-{issue-number} so it's unique across different repos?

.github/workflows/preview.yml
reakaleek: https://github.com/elastic/docs-builder/pull/279/commits/1acb9d9a7795078e7cc77f3b7b04f982ac708a81

.github/workflows/pr.yml
```diff
@@ -36,4 +39,17 @@ jobs:
...
       - name: Build documentation
-        run: .artifacts/publish/docs-builder/release/docs-builder --strict
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: .artifacts/publish/docs-builder/release/docs-builder --strict --path-prefix "/${GITHUB_REPOSITORY}/pull/${PR_NUMBER}"
+      
+      - uses: actions/upload-artifact@v4
+        with:
+          name: docs
+          path: .artifacts/docs/html
```
Mpdreamz, line 49: Cache our binary instead of the html. 

Our html is small but for e.g docs-content it will be many megabytes (e.g just the output of the converted obs docs is 100mb zipped due to all the images). 

Secondly we can then either grab the binary or run our github action inside the preview action. 

With the preview github action being the only one we expose to everyone to add to their PR builds.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,85 @@
...
+                log_url: `https://github.com/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}?pr=${context.issue.number}`,
+            })
+            return deployment.data.id
+
+      - uses: actions/checkout@v4
+        
+      - uses: actions/download-artifact@v4
+        with:
+          name: docs
+          path: .artifacts/docs/html
```
Mpdreamz, line 47: here we should conditionally download the cached binary or run our github action to generate the docs.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,85 @@
...
+
+      - uses: ./.github/actions/aws-auth
```
Mpdreamz, line 49: How does this work when called from other elastic repositories? 

I'd prefer this to be an action since we don't want to polute other UI's with all these steps (correct me if im wrong) but then auth might be an issue?

Mpdreamz (changes requested):
Few nits, can't wait to merge this in though so awesome!

.github/workflows/preview.yml
```diff
@@ -0,0 +1,85 @@
...
+permissions:
+  id-token: write
+  pull-requests: write
+  deployments: write
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create Deployment
```
Mpdreamz, line 15: Not necessary for this PR but we might want to protect this to only run if doc changes are part of the PR?

.github/workflows/preview.yml
reakaleek: Sorry, I should have been clearer in my description.

This is implementation is not supposed to be used for the consumer, only for this repo.

I was planning to create a consumer action in a follow-up.

Ofcourse it will be highly inspired by this, and we will see if there can be any commons that we can refactor out.

.github/workflows/preview.yml
reakaleek: My intention for this preview was not to create a preview of the docs only, but actually also a preview of changes within the docs-builder codebase.

.github/workflows/preview.yml
Mpdreamz: SGTM!

Mpdreamz (commented):
LGTM!

Mpdreamz (approved):

PR #334, opened 2025-01-23
Fix preview-cleanup pt3

reakaleek:
We can delete all deployments to clean-up the UI, but we cannot delete the environment.

We need a GH PAT for that with permissions to delete environments.

This is okay for now.


Mpdreamz (approved):
++

We can figure this out later too, its not the end of the world that the deployment leads to a non existant url after pulling.

Most folks wouldn't expect it to be there anymore.

PR #353, opened 2025-01-27
Add anchors to link json artifact

Mpdreamz:
Refactored `LinkReference` to use a dictionary for better anchor tracking and adjusted anchor validations. Added extensive test coverage, revamped test setup, and improved generator logic for markdown files.```


src/Elastic.Markdown/IO/State/LinkReference.cs
```diff
@@ -15,8 +15,9 @@ public record LinkReference
 	[JsonPropertyName("url_path_prefix")]
 	public required string? UrlPathPrefix { get; init; }
 
+	/// Mapping of relative filepath and all the page's anchors for deeplinks
 	[JsonPropertyName("links")]
-	public required string[] Links { get; init; } = [];
+	public required Dictionary<string, string[]> Links { get; init; } = [];
 
 	[JsonPropertyName("cross_links")]
 	public required string[] CrossLinks { get; init; } = [];
```
reakaleek, line 23: Do you think CrossLinks should have the same format? (Dictionary)

src/Elastic.Markdown/IO/State/LinkReference.cs
```diff
@@ -15,8 +15,9 @@ public record LinkReference
 	[JsonPropertyName("url_path_prefix")]
 	public required string? UrlPathPrefix { get; init; }
 
+	/// Mapping of relative filepath and all the page's anchors for deeplinks
 	[JsonPropertyName("links")]
-	public required string[] Links { get; init; } = [];
+	public required Dictionary<string, string[]> Links { get; init; } = [];
```
reakaleek, line 20: I'm wondering if an actual object instead of a string array would be more future proof.

e.g.

```json

"links": {
  "link.md": {
    "anchors": []
  }
}
```

This way we can add other attributes in the future, if necessary.

Just thinking out loud here. Also, I cannot think of anything right now.

reakaleek (approved):
🔥

src/Elastic.Markdown/IO/State/LinkReference.cs
Mpdreamz: Lets see if that becomes evident when we start working on resolving crosslinks.

src/Elastic.Markdown/IO/State/LinkReference.cs
Mpdreamz: Good call out! we can nest maybe deprecated (renamed links) here in the future too.

reakaleek (approved):

PR #357, opened 2025-01-28
Add error exit code if there are changes for the asciidoc comment workflow

reakaleek:


reakaleek:
@bmorelli25 WDYT about this?

This is not blocking, since it won't be a required status check. However, the check will be clearly red.

bmorelli25 (approved):
Love it 👍

PR #361, opened 2025-01-28
Allow _docset.yml and _toc.yml as configuration filenames

Mpdreamz:
This allows docsets to use underscore variants of the configuration files.

* `docset.yml`, `_docset.yml` in order of precedence
* `toc.yml`, `_toc.yml` in order of precedence

This will aid with content sets that hold a lot of content at the root.


reakaleek (approved):

Mpdreamz:
cc @bmorelli25 @KOTungseth let me know if this is something we want to improve DX for our writers.

bmorelli25 (approved):
💯 !!

reakaleek (approved):

PR #374, opened 2025-01-29
Use GitHub API instead of git to check diff

reakaleek:
## Changes

Without a checkout action, the [tj-actions/changes-files](https://github.com/tj-actions/changed-files) action will use the GitHub API to to determine the diff.
This way we can avoid checking out the forked PR and minimize the risk of using `pull_request_target` together
with `actions/checkout`.

Refs:
  - https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/
  - https://github.com/tj-actions/changed-files?tab=readme-ov-file#using-githubs-api-octocat

### DIFF
--- a/.github/workflows/comment-on-asciidoc-changes.yml
+++ b/.github/workflows/comment-on-asciidoc-changes.yml
@@ -12,17 +12,14 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout the repository
-        uses: actions/checkout@v4
-        with:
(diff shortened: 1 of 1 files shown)


reakaleek:
cc @donoghuc 

Thank you for your feedback. I was able to verify it on a test repo. It works 🚀

donoghuc (approved):
Great! Avoiding code checkouts is a huge win for this usecase 🚀

Mpdreamz (approved):
Great!

PR #376, opened 2025-01-30
Add `docs-builder mv` command

reakaleek:
## Changes

Add `mv` command to docs-builder CLI.

#### `docs-builder mv ./docs/migration/ia.md ./docs/new-folder/ia.md`

image

#### `docs-builder mv --help`

```shell
Usage: mv [arguments...] [options...] [-h|--help] [--version]

Move a file or folder from one location to another and update all links in the documentation

Arguments:
  [0] <string?>    The source file or folder path to move from
  [1] <string?>    The target file or folder path to move to

Options:
  --dry-run <bool?>      Dry run the move operation (Default: null)
  -p|--path <string?>    Defaults to the`{pwd}` folder (Default: null)
```

#### `docs-builder mv ./docs/migration/ia.md ./docs/new-folder/ia.md && git status`

```shell
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        deleted:    docs/migration/ia.md
        modified:   docs/migration/index.md

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        docs/new-folder/
```

## Limitations

- Currently, it can only move single files not folder
- The `docset.yml` and `toc.yml` files are not updated. However, you will get an error while building.


src/docs-builder/Cli/Move.cs
```diff
@@ -0,0 +1,228 @@
...
+		if (fileSystem.File.Exists(target))
+		{
+			_logger.LogError($"Target file {target} already exists");
+			return false;
+		}
+
+		return true;
+	}
+
+	private async Task ProcessMarkdownFile(
```
Mpdreamz, line 142: I was hoping that we could leverage markdig here. 

E.g we parse the document AST and include a linkrewriter parser (or make it part of our existing `DiagnosticLinksParser`. To mutate the links.

Then we can write the AST back out again as markdown. 

markdig supports roundtipping like this.

reakaleek: Makes totally sense.

Given the time and the need for it already today/tomorrow, do you think we can leave it this way for now and refactor it in a follow-up using Markdig?

Mpdreamz: 100%! progress over perfection.

Mpdreamz (approved):
LGTM!

PR #381, opened 2025-01-30
Add support for hidden files in documentation TOC

Mpdreamz:
Hidden files can now be excluded from navigation and indexing but still linked directly if needed. Adjustments include handling hidden attributes in TOC parsing and ensuring related navigation logic respects these exclusions.

Hidden pages can be included just as you would any `file` in the TOC. 

```
   hidden: my-page.md
```

* Still only one place to configure files. 
* Participates in dangling documentation detection
* Is picked up as a `MarkdownFile` in our codebase like any other.
* Includes a hidden flag so that we:
  * Never render it in any navigation item.
  * Always include a noindex header.
  * Advertise it as hidden in links.json artifact.

https://github.com/user-attachments/assets/f6d0f4ed-0d60-4ac8-8718-0867437e9359

### DIFF
--- a/docs/configure/content-set/navigation.md
+++ b/docs/configure/content-set/navigation.md
@@ -73,15 +73,18 @@ toc:
   - file: index.md
 ```
 
-The table of contents can be created independent of the directory structure of the files it defines. You can use directories to define nesting in the TOC, but you don't have to. For example, both of the following create the same nav structure:
+The TOC in principle follows the directory structure on disk.
+
+#### `folder:`
 
 ```yaml
   ...
-  - file: subsection/index.md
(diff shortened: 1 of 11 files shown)


reakaleek (approved):

shainaraskas (approved):
!! NICE!

if I want to hide multiple files in the same TOC, do I need individual `hidden` entries, or does it accept an array? either works just interested for clarity

Mpdreamz:
You can place multiple hidden entries just like you would multiple file entries. 

Youm ay have hidden pages anywhere in the TOC they themselves however may not have any children.

PR #383, opened 2025-01-30
Add move to target folder support to `mv` command

Mpdreamz:


reakaleek (approved):
Thank you for improving!

PR #385, opened 2025-01-30
Implement whole folder moves to the `mv` command

Mpdreamz:
Piggybacking on the hard work of @reakaleek this now implements folder to folder support to the `mv` command.


reakaleek (approved):
🚀

PR #389, opened 2025-01-31
Set up parceljs

reakaleek:
## Context

Prepare redesign.

## Changes

Set up [parceljs](https://parceljs.org/).

- Move css files into `src/Elastic.Markdown/Assets/legacy` and import it in `styles.css`

## Notes
- I was unable to move JS because it heavily relies on global variables, which is hard to achieve with JS modules.
- The new `main.js` file is empty and does nothing for now.

## Follow-up
- Set up tailwind


reakaleek:
Dang, it doesn't work as expected on the deployment

Mpdreamz (approved):
Nice work getting it all integrated!

PR #392, opened 2025-02-01
Add redesign feature flag

reakaleek:
## Changes

- Set up tailwind
- Add a "REDESIGN" feature flag
- Add basic layout when the "REDESIGN" flag is enabled
- Set up hljs

## Notes

You can enable the feature flag if you set the environment variable `REDESIGN=true`

e.g. 

```
REDESIGN=true ./build.sh watch
```

## Result

image


src/Elastic.Markdown/Slices/Layout/_Head.cshtml
```diff
@@ -7,17 +7,24 @@
...
-	<link media="print" rel="stylesheet" type="text/css" href="@Model.Static("print.css")"/>
+	<link rel="stylesheet" type="text/css" href="@Model.Static("pygments.css")"/>
+	<link rel="stylesheet" type="text/css" href="@Model.Static("shibuya.css")"/>
+	<link rel="stylesheet" type="text/css" href="@Model.Static("mystnb.css")"/>
+	<link rel="stylesheet" type="text/css" href="@Model.Static("copybutton.css")"/>
+	<link rel="stylesheet" type="text/css" href="@Model.Static("togglebutton.css")"/>
+	<link rel="stylesheet" type="text/css" href="@Model.Static("sphinx-design.min.css")"/>
+	<link media="print" rel="stylesheet" type="text/css" href="@Model.Static("/_static/print.css")"/>
+	<link rel="stylesheet" type="text/css" href="@Model.Static("custom.css")"/>
+	<link rel="stylesheet" type="text/css" href="@Model.Static("atom-one-light.css")"/>
```
reakaleek, line 18: Revert the old css here, so that the parcel setup only has the new styles.

Mpdreamz (changes requested):
Shall we just redesign as we go along? 

I'm not that keen on having a fork in the road. Since the old design is not released nor was never meant to be released. 

I feel confident we get the new design to a good enough state during the freeze. Wdyt?

Mpdreamz (approved):
Scratch that, We need this flag at least for the short interim period.where we rebuild all the basic functionally.

PR #406, opened 2025-02-03
New "safe" preview deployment workflows

reakaleek:
## Context

The current preview deployment workflow required `pull_request_target` to be able to deploy PRs from forks.

Using `pull_request_target` in conjunction with checking out the fork is considered a security risk.

## Changes

This introduces a workflow setup that requires three workflows. 
1. A workflow that builds the docs using the `pull_request` event and uploads it as artifacts
2. A workflow that downloads the artifacts and deploys the docs to the preview environment using the `workflow_run` event
3. A workflow that destroys the destroys the preview environment using the `pull_request_target` event (but without a **checkout** step)

This way untrusted code will never be checked out.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -42,7 +42,3 @@ jobs:
           path: .artifacts/publish/docs-builder/release/docs-builder
           if-no-files-found: error
           retention-days: 1
-
-  preview:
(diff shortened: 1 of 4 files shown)


bmorelli25 (approved):
🚀

PR #408, opened 2025-02-04
Fix deploy preview workflow

reakaleek:
There is a conflict with the downloaded artifact in a previous step

Happened in https://github.com/elastic/docs-content/actions/runs/13133333037/job/36643123193

### DIFF
--- a/.github/workflows/preview-deploy.yml
+++ b/.github/workflows/preview-deploy.yml
@@ -65,7 +65,8 @@ jobs:
         run: |
           gh run download ${{ github.event.workflow_run.id }} \
             --repo "${GITHUB_REPOSITORY}" \
-            --name docs
+            --name docs \
+            --dir html
 
       - uses: elastic/docs-builder/.github/actions/aws-auth@main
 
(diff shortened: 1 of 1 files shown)


reakaleek:
🤔 But it worked just fine here

.github/workflows/preview-deploy.yml
```diff
@@ -65,15 +65,16 @@ jobs:
...
+            --dir html
 
       - uses: elastic/docs-builder/.github/actions/aws-auth@main
 
       - name: Upload to S3
         env:
           PR_NUMBER: ${{ steps.pull_request.outputs.number }}
         run: |
-          aws s3 sync . "s3://elastic-docs-v3-website-preview/${GITHUB_REPOSITORY}/pull/${PR_NUMBER}" --delete --exclude "pull_request.json"
+          aws s3 sync ./html "s3://elastic-docs-v3-website-preview/${GITHUB_REPOSITORY}/pull/${PR_NUMBER}" --delete
```
Mpdreamz, line 77: This works today because we publish `links.json` under `html`. 

`.artifacts/docs/links.json` would have been better than `.artifacts/docs/links.json`

PR #422, opened 2025-02-05
Add path-pattern input to preview-build.yml workflow

reakaleek:
## Changes

- Add `path-pattern` input for the preview-build workflow
- Run build and deployment only if the path-pattern matches

## Details

This means a PR to the repository will only build and deploy the documentation if there are changes to the docs (assuming the pattern matches). But repo owners can still set the commit status as required status check, because it will still be green.

### DIFF
--- a/.github/workflows/preview-build.yml
+++ b/.github/workflows/preview-build.yml
@@ -32,7 +32,6 @@ jobs:
         id: check-files
         uses: tj-actions/changed-files@d6e91a2266cdb9d62096cebf1e8546899c6aa18f # v45.0.6
         with:
-          since_last_remote_commit: true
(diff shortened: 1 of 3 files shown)


.github/workflows/preview-build.yml
```diff
@@ -13,15 +13,29 @@ on:
         type: string
         required: false
         default: 'true'
+      path-pattern:
+        description: 'Path pattern to filter files. Only if changed files match the pattern, the workflow will continue.'
+        type: string
+        default: '**'
```
reakaleek, line 19:
```suggestion
        default: 'docs/**'
```

Mpdreamz (approved):

Issue #347, opened 2025-01-27
Evaluate if `.EnabeTrackTrivia` is necessary


reakaleek:
In https://github.com/elastic/docs-builder/pull/345 we had to remove the extra indent caused by `.EnableTrackTrivia`

However, we should also check if TrackTrivia is really needed.

Related comment: https://github.com/elastic/docs-builder/pull/345#event-16074367517

Issue #405, opened 2025-02-03
Definition lists don't get consistent spacing


shainaraskas:
### Describe the bug

If a definition consists of a single paragraph and is preceded by another definition, it doesn't get a line break after it 

<img width="708" alt="Image" src="https://github.com/user-attachments/assets/4748e828-3c25-45b0-ac22-96d4bbecb3c4" />

### Expected behavior

Line breaks between definitions

### Steps to reproduce

put this anywhere
```
Consumption-based billing
:   You pay for the actual product used, regardless of the application or use case. This is different from subscription-based billing models where customers pay a flat fee restricted by usage quotas, or one-time upfront payment billing models such as those used for on-prem software licenses.

    You can purchase credits for a single or multi-year contract. Consumption is on demand, and every month we deduct from your balance based on your usage and contract terms. This allows you to seamlessly expand your usage to the full extent of your requirements and available budget, without any quotas or restrictions.

Resource-based pricing
:   You are charged for the actual computing resources you use: capacity (consisting of RAM and CPU plus disk allowance), data transfer, and snapshot storage. This normalizes billing dimensions across different use cases (for example, the number of agents for APM, the number of hosts for Security, the volume of data ingest for Observability, and so on) making it easy to compare and aggregate costs.

B
:   hello

Elastic Consumption Unit (ECU)
:   An ECU is a unit of aggregate consumption across multiple resources over time.

    Each type of computing resource (capacity, data transfer, and snapshot) that you consume has its own unit of measure. For example, capacity is measured in GB-hour, data transfer in GB, snapshot storage in GB-month and snapshot API requests in thousands of requests.

    In order to aggregate consumption across different resource types, all resources are priced in ECU. ECU has a fixed exchange rate to fiat currency of 1 ECU = $1.00.

    Check [Using Elastic Consumption Units for billing](../../../deploy-manage/cloud-organization/billing/billing-models.md) for more details.
```

### Tooling

- [x] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

bmorelli25:
It looks like this is happening because definition lists without a line break are rendered as text inside `<dd>` blocks. In your example:

```
<dd>   hello</dd>
```

But text _with_ line breaks are rendered as paragraph elements within the `<dd>` blocks. In your example:

```
<dd>
[...]

PR #435, opened 2025-02-06
Fix lists continuations.

Mpdreamz:
List continuations that have blank lines are considered loose. As per commonmark spec these should render as paragraphs.

`EnableTrackTrivia()` was breaking this behaviour.

Validated we don't need triva for error message highlighting.

Closes https://github.com/elastic/docs-builder/issues/347

### DIFF
--- a/docs/syntax/applies.md
+++ b/docs/syntax/applies.md
@@ -52,9 +52,7 @@ Are equivalent, note `all` just means we won't be rendering the version portion 
 ## This section has its own applies annotations [#sections]
 
 :::{applies}
-:stack: unavailable
-:serverless: tech-preview
-:cloud: ga
+:serverless: unavailable
 :::
 
 :::{note}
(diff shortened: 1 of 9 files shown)


reakaleek:
Closes https://github.com/elastic/docs-builder/issues/347

reakaleek (approved):

bmorelli25:
Dumb question: is this also related to the `<p>` issue in definition lists? See: https://github.com/elastic/docs-builder/issues/405#issuecomment-2638315942

Mpdreamz:
> Dumb question: is this also related to the <p> issue in definition lists? See: https://github.com/elastic/docs-builder/issues/405#issuecomment-2638315942

That's so not dumb I am stuck wondering if your self deprecation is genuine 😸 Good catch! yes this will fix #405 too. Added a test to confirm.

PR #447, opened 2025-02-07
Add pages navigation

reakaleek:
## Changes

Add the pages navigation (left sidebar)

**Features**:
- Feels like a SPA
  - Scroll Position stays the same when clicking through the navigation items
  - State of collapsed/expanded groups stays while changing pages
- Almost pure css implementation
  - only rehydrating the state between page navigations uses JS 
- Scroll to current navigation item (if it's not in the viewport)
- Progressive Enhancement
  - The navigation works also without javascript
  - But the experience is enhanced, if javascript is enabled.

## Result

https://github.com/user-attachments/assets/9706c4c1-9441-4f66-9339-61aebaf2813e


src/Elastic.Markdown/Slices/_Layout.cshtml
```diff
@@ -37,31 +39,40 @@
...
+			<nav id="pages-nav" class="sticky top-22 w-80 z-10 max-h-[calc(100vh-var(--spacing)*22)] overflow-y-auto">
+				@(new HtmlString(Model.NavigationHtml))
+			</nav>
+		</aside>
+		<main class="markdown-content w-full max-w-screen pt-6 pb-6 px-6">
 			<ol class="flex-1 mb-6" itemscope="" itemtype="https://schema.org/BreadcrumbList">
-				<li class="inline text-gray-500" itemprop="itemListElement" itemscope="" itemtype="https://schema.org/ListItem">
-					<a itemprop="item" href="/">
-						<span itemprop="name">Elastic</span>
+				<li class="inline text-ink text-sm hover:text-ink leading-[1.2em] tracking-[-0.02em]" itemprop="itemListElement" itemscope="" itemtype="https://schema.org/ListItem">
```
Mpdreamz, line 50: I am not a fan of these inline templated classes:

`leading-[1.2em]`

Using custom component classes:
https://tailwindcss.com/docs/styling-with-utility-classes#using-custom-css 
will be a better long term play in terms of maintainability. 

Most of these templated hardcoded values are better off as CSS variables. E.g we want to normalize padding/margin's as much as possible with little variation. As per: https://cieden.com/book/sub-atomic/spacing/spacing-best-practices 


Similar the `top-22 w-80 z-10 max-h-[calc(100vh-var(--spacing)*22)]` above.

reakaleek: > `leading-[1.2em]`



I'm struggling a bit to find a pragmatic way to do this.

Do you think we can do it in a follow-up? The right sidebar (table of contents) will be very similar, with subtle differences, and I think it's a good time to resolve that issue, when implementing it.

Mpdreamz: Absolutely! Progress over perfection.

PR #455, opened 2025-02-10
Improve safety of callout handling in EnhancedCodeBlockParser.

Mpdreamz:
Adjusted logic to ensure stricter detection of classic callouts by checking for a closing angle bracket at the end of the span. Also updated regex generation in CallOutParser to use a timeout for better performance and stability.

This brings the full build of the [reference branch in asciidocalypse](https://github.com/elastic/asciidocalypse/tree/reference/docs) back down to ~45s on my local laptop again (from 4 minutes prior).

Most of the remaining time is in the unoptomized `RenderNavivation()` subbing that out for something that is cached brings the total build further down to ~10s. Will save that work for another PR once the redesign kicks in!

### DIFF
--- a/src/Elastic.Markdown/DocumentationGenerator.cs
+++ b/src/Elastic.Markdown/DocumentationGenerator.cs
@@ -72,7 +72,6 @@ public class DocumentationGenerator
 
 		await ExtractEmbeddedStaticResources(ctx);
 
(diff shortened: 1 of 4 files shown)


reakaleek (approved):
Wow. 🚀

PR #487, opened 2025-02-13
Add previous/next button design

reakaleek:
https://github.com/user-attachments/assets/3c0ddc53-ffc1-4dd1-802b-8c86e21ace85


Mpdreamz (approved):
LGTM! can we now fully roll on with the redesign? :)


EDIT: meaning can we only ship the redesign now and remove the old?

PR #491, opened 2025-02-13
Initial support for cross link resolving

Mpdreamz:
This is the first stab at it: 

- We need a index.json file that lists all links.json files with their latest ref 
- This allows us to cache links.json files locally. 

All links resolve to: `https://docs-v3-preview.elastic.dev/elastic/<repos>/tree/main` currently. 

The index.json file should also list the `current` branch name so we do not assume `main`.

### DIFF
--- a/src/Elastic.Markdown/CrossLinks/CrossLinkResolver.cs
+++ b/src/Elastic.Markdown/CrossLinks/CrossLinkResolver.cs
@@ -14,7 +14,7 @@ namespace Elastic.Markdown.CrossLinks;
 public interface ICrossLinkResolver
 {
 	Task FetchLinks();
-	bool TryResolve(Action<string> errorEmitter, Uri crossLinkUri, [NotNullWhen(true)]out Uri? resolvedUri);
+	bool TryResolve(Action<string> errorEmitter, Uri crossLinkUri, [NotNullWhen(true)] out Uri? resolvedUri);
 }
 
 public class CrossLinkResolver(ConfigurationFile configuration, ILoggerFactory logger) : ICrossLinkResolver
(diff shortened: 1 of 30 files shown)


src/Elastic.Markdown/CrossLinks/CrossLinkResolver.cs
```diff
@@ -0,0 +1,93 @@
...
+	public static LinkReference Deserialize(string json) =>
+		JsonSerializer.Deserialize(json, SourceGenerationContext.Default.LinkReference)!;
+
+	public async Task FetchLinks()
+	{
+		using var client = new HttpClient();
+		var dictionary = new Dictionary<string, LinkReference>();
+		foreach (var link in _links)
+		{
+			var url = $"https://elastic-docs-link-index.s3.us-east-2.amazonaws.com/elastic/{link}/main/links.json";
```
reakaleek, line 35: I think this can be `master` in some cases

src/Elastic.Markdown/CrossLinks/CrossLinkResolver.cs
```diff
@@ -0,0 +1,93 @@
...
+			_logger.LogInformation($"Fetching {url}");
+			var json = await client.GetStringAsync(url);
```
reakaleek, line 37: What happens if the url doesn't exit yet?

While we onboard repositories, not all of the might be available from the start.

src/Elastic.Markdown/CrossLinks/CrossLinkResolver.cs
Mpdreamz: Aye, that's why we need a links index file that points to all current index files.

src/Elastic.Markdown/CrossLinks/CrossLinkResolver.cs
Mpdreamz: For now I'm only going to make `docs-content` mandatory to resolve. 

All others we'll blindly transform to urls for now. (will follow up with another PR to introduce this behavior).

reakaleek (approved):

PR #496, opened 2025-02-14
Only validate `docs-content` cross links for now

Mpdreamz:
Introduced a `_declaredRepositories` field in `CrossLinkResolver` to track valid repositories explicitly. Updated logic to validate cross-links against declared repositories and handle errors or missing links more gracefully. Extended tests to ensure correctness and consistency with declared repository functionality.

This allows us to temporarily allow cross links to repositories that do not publish a links.json file.

With this PR we only validate:

- docs-content links completely

For all other repositories we only validate that they are declared in docset.yml


reakaleek (approved):
Wondering if can do it a bit "dynamic" for now.

e.g.:

```
check if `elastic-docs-v3-preview.elastic.dev/path/to/index.html` exists

if exists:
  do the validation
else
  do a warning
```

This logic could be removed later, when all repositories are onboarded

Mpdreamz:
I don't think its a good idea to do an http call foreach of these. 

If all goes well reference content gets put into their respective repositories after the 20th and we setup the preview/publish links for all of them the week after. 

We only need to run in this partial mode 2 weeks. 

After that I'll make cross links to docs-content errors if not correct and others warnings while we (the docs-team) address all of them. 

cc @bmorelli25 @KOTungseth ^ plan for crosslinks.

PR #501, opened 2025-02-14
Add Documentation Assembler GitHub Action

Mpdreamz:
Introduce a new GitHub Action named "Documentation Assembler" to run specified assembler commands using a Docker image.

### DIFF
--- a/actions/assembler/action.yml
+++ b/actions/assembler/action.yml
@@ -1,0 +1,15 @@
+name: 'Documentation Assembler'
+description: 'Run the documenation assembler commands'
+
+branding:
+  icon: 'filter'
+  color: 'red'
+
+inputs:
+  command:
+    description: 'The assembler command to run'
+    required: true
+
+runs:
+  using: 'docker'
+  image: "docker://ghcr.io/elastic/docs-assembler:edge"
--- a/src/docs-assembler/Cli/RepositoryCommands.cs
+++ b/src/docs-assembler/Cli/RepositoryCommands.cs
@@ -1,0 +1,94 @@
+// Licensed to Elasticsearch B.V under one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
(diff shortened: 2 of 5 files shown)


actions/assembler/action.yml
```diff
@@ -0,0 +1,15 @@
+name: 'Documentation Assembler'
+description: 'Run the documenation assembler commands'
+
+branding:
+  icon: 'filter'
+  color: 'red'
```
reakaleek, line 6: nit: This is not really required. And I think you can only ever see this if it was published in the marketplace

src/docs-assembler/Cli/RepositoryCommands.cs
```diff
@@ -0,0 +1,94 @@
...
+					var name = kv.Key;
+					var repository = kv.Value;
+					var checkoutFolder = Path.Combine(Paths.Root.FullName, $".artifacts/assembly/{name}");
+
+					var sw = Stopwatch.StartNew();
+					dict.AddOrUpdate(name, sw, (_, _) => sw);
+					Console.WriteLine($"Checkout: {name}\t{repository}\t{checkoutFolder}");
+					var branch = repository.Branch ?? "main";
+					var args = new StartArguments(
+						"git", "clone", repository.Origin, checkoutFolder, "--depth", "1"
```
reakaleek, line 59: What about sparse-checkout? Did you already try that?

reakaleek (approved):

src/docs-assembler/Cli/RepositoryCommands.cs
Mpdreamz: I followed the 4 clone patterns defined here: https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/

But can't remember the results :)

Will go over it again when we set up our runner.

reakaleek: Yup. In our case we can only download the docs folder.

PR #510, opened 2025-02-14
Remove UseSoftlineBreakAsHardlineBreak from MarkdownParser

Mpdreamz:
This change ensures that soft line breaks are no longer treated as hard line breaks during markdown parsing. It enhances compatibility with standard markdown behavior.

This implements the request made here: https://github.com/elastic/docs-builder/discussions/343

This should align better with how writers write as per: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

NOTE this is high impactful on `asciidocalypse` and `docs-content`. 

Introduces a new `soft_line_endings` configuration option to `docset.yml` control line break behavior in Markdown parsing. When enabled, soft line endings are converted to hard HTML breaks (`<br />`). 

NOTE: The new default of this config option is `false` so won't turn soft enters in the markdown to hard breaks in the HTML.

cc @karenzone @shainaraskas


bmorelli25:
Well that was fast 😅. I'll check this out and test against docs-content later today!

reakaleek (approved):

karenzone (commented):
Fantastic news! Thanks for making this happen--and so quickly!

bmorelli25:
So I must be doing something wrong. I don't notice any differences in the output when testing this against docs-content. I also tried to write a script that diffs the HTML output of building with this change vs building without this change. After correcting for changes in edit_this_page links, the HTML diff is empty. Does that sound possible?

shainaraskas:
@bmorelli25 it sounds possible to me - the reason why this change is relatively safe is because the migration tool imported paragraphs as lines, and put hard line breaks between paragraphs.

consider trying to point it at a branch that you have altered to have the "problematic" content, e.g.

```
This should appear as two paragraphs.
Because soft line breaks are turned off.
```

## current behavior

```
To integrate with Active Directory, you configure an `active_directory` realm and map Active Directory users and groups to roles in the role mapping file.
TEST TEST!
```

becomes

image


and 

```
To integrate with Active Directory, you configure an `active_directory` realm and map Active Directory users and groups to roles in the role mapping file.

TEST TEST!
```

becomes 

image


## what we want is

```
To integrate with Active Directory, you configure an `active_directory` realm and map Active Directory users and groups to roles in the role mapping file.
TEST TEST!
```

to be 

image


and 

```
To integrate with Active Directory, you configure an `active_directory` realm and map Active Directory users and groups to roles in the role mapping file. 

TEST TEST!
```

to be

image


looks like a CSS change to the `p` default class might also be necessary to support the latter, so you'll probably see two paragraphs kissing, like this:

image

going to quickly regex again for this case - when I did it before, the only lines sitting directly beside each other were inside of code blocks (which this change does not impact).

shainaraskas:
yep, quick scan shows we only have consecutive lines of text in:

- front matter 
- code blocks
- lists
- intros to lists
- directives + their contents
- definition lists

used `[a-z,0-9,\.]\n{1}[^\d\W]`

shainaraskas:
@Mpdreamz are we able to merge this?

PR #516, opened 2025-02-17
Enable new design by default

reakaleek:
## Changes

This enables the new design by default.

## Follow-ups

- Clean up unused files


Mpdreamz (approved):
LGTM, we need to update our github workflows too.

PR #528, opened 2025-02-18
Fix pages navigation

reakaleek:
## Context

There is a chance that navigation items have the same slug. Since the slug is used to identify the input checkbox, the wrong navigation group gets opened/closed.

## Changes

- Use a UUID for the input id
- Add `pointer-events-none` to avoid opening the group. (The group is visually opened twice, before the page is loaded and after the page is loaded)


reakaleek:
@Mpdreamz maybe we can try something like https://github.com/Cysharp/Ulid, which seems to be faster in generating ids.

WDYT?

Mpdreamz (approved):

Mpdreamz:
I don't think we generate UUID's at a rate which will require ULID, I'm also eager to start caching the whole navigation generation.

PR #535, opened 2025-02-18
Fix layout shift on page load

reakaleek:
## Context

On https://docs-v3-preview.elastic.dev/elastic/docs-content/tree/main you can see the layout shifting when navigating through pages.

## Changes

This sets the min width of the content container so the layout doesn't shift.


Mpdreamz (approved):

reakaleek:
This did not work as expected.

PR #537, opened 2025-02-18
Cache Navigation

Mpdreamz:
https://github.com/user-attachments/assets/fcdd8e19-2dae-46a7-9695-8a0403f4a3b8

Caches navigation for significantly faster build times.

Uses javascript to set the current page selected and expand its parent. Something we already do as we keep track of which nodes the user expanded.

Includes a few javascript speedups too.

- don't iterate over all inputs to set the previously selected, instead loop over selected state.
- when saving we only iterate over the checked inputs

We burn an overall navigation uuid into nav, if the users state does not match this we reset the users state. This prevents needlessly growing the users sessionStorage with UUID's that won't be valid anymore since the HTML is regenerated.

### DIFF
--- a/src/Elastic.Markdown/Assets/pages-nav.ts
+++ b/src/Elastic.Markdown/Assets/pages-nav.ts
@@ -1,33 +1,62 @@
 import {$, $$} from "select-dom";
 
-type NavExpandState = { [key: string]: boolean };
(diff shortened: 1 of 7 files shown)


reakaleek (approved):
🚀 🚀

PR #542, opened 2025-02-19
Rename Elastic Cloud to Elastic Cloud Hosted for applies_to displayed values

florent-leborgne:
Elastic Cloud includes both Elastic Cloud Hosted and Serverless, so we rather need "ess" to map to "Elastic Cloud Hosted" rather than "Elastic Cloud" to avoid ambiguity on our pages.

When something applies to "all Elastic Cloud", we can show both `Elastic Cloud Hosted` and `Serverless`, also to avoid any ambiguity.


Mpdreamz (approved):
LGTM!

PR #543, opened 2025-02-19
Change one Security serverless "applies to" duplicate to missing Elasticsearch serverless

florent-leborgne:
Seems like ES is missing and Security is duplicated. Hopefully this actually fixes it 😅 Otherwise we get this: 
image


Mpdreamz (approved):
Good catch!

Issue #443, opened 2025-02-06
Repeated entries in navigation


lcawl:
### Describe the bug

Some pages appears twice in the navigation even though they're mentioned only once in the `toc.yml` file. For example:

![Image](https://github.com/user-attachments/assets/7e7483cb-8b18-4d49-af63-2fc14a212410)

I see this happenening in places of the [toc](https://github.com/elastic/asciidocalypse/blob/reference/docs/docset.yml) like 

```
    - toc: elasticsearch/docs/reference/elasticsearch-plugins
```

... which references https://github.com/elastic/asciidocalypse/blob/reference/docs/elasticsearch/docs/reference/elasticsearch-plugins/toc.yml

The first file referenced in that `toc.yml` is shown twice:

```
toc:
  - file: elasticsearch-plugins.md
 ```

I searched on the navigation title just to be sure there wasn't a duplicate file lurking.

### Expected behavior

I expected these pages to be used for the section in the navigation, but not to re-appear within that section.

### Steps to reproduce

Build the `reference` branch: https://github.com/elastic/asciidocalypse/tree/reference

### Tooling

- [x] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

lcawl:
There was a similar problem in the ECS reference section, so I've worked around it for now by bumping that section up a level in the navigation tree in https://github.com/elastic/asciidocalypse/pull/20

lcawl:
In https://github.com/elastic/asciidocalypse/pull/93 I renamed the first file in the `toc.yml` to `index.md` and the problem went away. Is this a hard requirement?

Mpdreamz:
No this is a bug! we should address it so we don't force everyone to use `index.md` everywhere which is a terrible experience in editors.

lcawl:
> No this is a bug! we should address it so we don't force everyone to use `index.md` everywhere which is a terrible experience in editors.

Yay, I've done that to move forward for now but agree it's not ideal.

PR #558, opened 2025-02-19
Update navigation.md with file name requirement

lcawl:
Relates to https://github.com/elastic/docs-builder/issues/443

Assuming that the current behaviour is working as designed, I think we need to call out the requirement that the first file always be named `index.md` (otherwise it seems to result in duplicated navigation entries).  If that's a bug, however, this PR can be closed.

### DIFF
--- a/docs/configure/content-set/navigation.md
+++ b/docs/configure/content-set/navigation.md
@@ -73,6 +73,10 @@ toc:
   - file: index.md
 ```
 
+:::{important}
+The first file must always be named `index.md`.
+:::
+
 The TOC in principle follows the directory structure on disk.
 
 #### `folder:`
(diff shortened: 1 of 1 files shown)


docs/configure/content-set/navigation.md
```diff
@@ -73,6 +73,10 @@ toc:
   - file: index.md
 ```
 
+:::{important}
+The first file must always be named `index.md`.
+:::
```
Mpdreamz, line 78: This is actually not true, the first listed file is the `index`. 

We should avoid having a ton of `index.md` files to aid with contributing IMO.

Mpdreamz (changes requested):

docs/configure/content-set/navigation.md
lcawl: Yay, I'm glad it's not a requirement. I'll close this PR.

PR #574, opened 2025-02-21
pages-nav: Fix spacing shift when clicking on a group

reakaleek:
## Context

When clicking on a group and the group expands, there is a visible layout shift

https://github.com/user-attachments/assets/e1ab1c4a-5acf-4e2d-9ca4-7c9374823f8d

## Result

https://github.com/user-attachments/assets/4b608988-1920-4736-b617-6fd59208815e


Mpdreamz (approved):
Jitter begone!

Issue #568, opened 2025-02-20
tab groups can't be indented into lists


shainaraskas:
### Describe the bug

if you indent a tab set into a numbered list and two+ tab sets exist on the page, the tab set operates the first set of tabs all the time.

https://github.com/user-attachments/assets/df600d35-d2d2-4f32-ac1c-df63740155af

it's not immediately apparent that something is wrong, and this is hard to diagnose - would be good to fix it to avoid accidental breakage

### Expected behavior

I can indent a tab set and it operates independently

### Steps to reproduce

use this snippet (sorry it's ugly and long)

````
## Step 1: Add a new realm configuration [ad-realm-configuration]

1. Add a realm configuration of type `active_directory` to `elasticsearch.yml` under the `xpack.security.authc.realms.active_directory` namespace. At a minimum, you must specify the Active Directory `domain_name` and `order`.

    See [Active Directory realm settings](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-settings.html#ref-ad-settings) for all of the options you can set for an `active_directory` realm.

    :::{note} 
    Binding to Active Directory fails if the domain name is not mapped in DNS. 

    In a self-managed cluster, if DNS is not being provided by a Windows DNS server, you can add a mapping for the domain in the local `/etc/hosts` file.
    :::

  ::::{tab-set}

  :::{tab-item} Single domain
  The following realm configuration configures {{es}} to connect to `ldaps://example.com:636` to authenticate users through Active Directory:

  ```yaml
  xpack:
    security:
      authc:
        realms:
          active_directory:
            my_ad:  
              order: 0 <1>
              domain_name: ad.example.com <2>
              url: ldaps://ad.example.com:636 <3>
  ```

  1. The order in which the `active_directory` realm is consulted during an authentication attempt.
  2. The primary domain in Active Directory. Binding to Active Directory fails if the domain name is not mapped in DNS.
  3. The LDAP URL pointing to the Active Directory Domain Controller that should handle authentication. If you don’t specify the URL, it defaults to `ldap:<domain_name>:389`.
  :::

  :::{tab-item} Forest
  Set the `domain_name` setting to the forest root domain name.

  You must also set the `url` setting, since you must authenticate against the Global Catalog, which uses a different port and might not be running on every Domain Controller.

  For example, the following realm configuration configures {{es}} to connect to specific Domain Controllers on the Global Catalog port with the domain name set to the forest root:

  ```yaml
  xpack:
    security:
      authc:
[...]

PR #578, opened 2025-02-21
Fix tabs in list

reakaleek:
Closes https://github.com/elastic/docs-builder/issues/568

## Context

When tabs are in a list or is a child of any block, then the index, which is used as ID is is wrong because it only searches for siblings.

## Changes

This change uses the line number as ID for simplicity. It doesn't represent the actual number of the tab set but it's unique and effecient.

## Result

https://github.com/user-attachments/assets/87600b8c-164a-482c-a751-7ac75342d279

### DIFF
--- a/src/Elastic.Markdown/Myst/Directives/TabSetBlock.cs
+++ b/src/Elastic.Markdown/Myst/Directives/TabSetBlock.cs
@@ -2,11 +2,8 @@
 // Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information
 
-using System.Linq;
-using System.Reflection.Metadata;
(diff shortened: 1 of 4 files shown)


Mpdreamz (approved):
Simple == brilliant

src/Elastic.Markdown/Myst/Directives/TabSetBlock.cs
```diff
@@ -10,6 +10,8 @@ namespace Elastic.Markdown.Myst.Directives;
 public class TabSetBlock(DirectiveBlockParser parser, ParserContext context)
 	: DirectiveBlock(parser, context)
 {
+	public string Id { get; private set; } = Guid.NewGuid().ToString();
+
```
reakaleek, line 14:
```suggestion
```

Issue #580, opened 2025-02-21
Ctrl+F searches collapsed nav items with no user feedback


shainaraskas:
### Describe the bug

When you `ctrl`+`f` for a common word, (e.g. `data`), navigating through the results using `enter` looks like it's not working, and there are tons of results. 

What I suspect is happening, is it's detecting all of the instances of the word in the sidebar, but because the sidebar is collapsed, you can't see it navigating through the options. 

The reason I think this is because when I expand a node that I know contains the word after initiating `ctrl`+`f` and hit `enter`, the number doesn't increase but the word becomes highlighted. also, if I keep hitting `enter`, I'll finally see an instance be selected in the sidebar, but it will disappear (into collapsed child items) shortly after

Some collapsed elements _should_ be searchable: e.g. unselected tabs and collapsed dropdowns

### Expected behavior

- only text currently visible in nav is searched
- all content of the current article is searched (including unselected tabs and collapsed dropdowns)

### Steps to reproduce

`ctrl`+`f` for the word `data`

### Tooling

- [x] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

PR #583, opened 2025-02-21
Exclude navigation elements from browser search if collapsed

reakaleek:
Closes https://github.com/elastic/docs-builder/issues/580

## Context

The sub lists of the navigation items had `height: 0` and were expanded based on the height. This was done because the initial idea was to expand with an animation.

## Change

Instead of hiding with `height: 0` use `display: hidden`, this way it's excluded from the browser search.


bmorelli25 (approved):
tested and works for me!

PR #586, opened 2025-02-22
feature/htmx poc main pages cherry pick

Mpdreamz:
This adds seperate `.page.html` output adjacent to the regular HTML output. 

On e.g asciidocalypse a single html page is `5mb` due to the massive navigation thats included on each page. 

This PR creates a `.page.html` that HTMX will get and prefetch instead which is typically only in the `4-80kb` range. Which should aid with latency while navigating through the docs.

The `serve` command has also been updated to render pages instead when receiving `.page.html` requests.


reakaleek:
Not sure why, but the link in the markdown doesn't use hx anymore.

reakaleek:
> Not sure why, but the link in the markdown doesn't use hx anymore.

Nevermind. This is already the case also in `main`.

reakaleek (approved):

PR #587, opened 2025-02-24
Add inbound links validation

Mpdreamz:
Adds two commands to `docs-assembler`

```
$ docs-assembler link validate-inbound-all
```

and 

```
$ docs-assembler link validate-inbound-local
```

### validate-inbound-all

Validates all repositories published to the link index. Each `links.json` has `cross_links` section declaring all cross links used in that docset. This command will validate each cross link actually still resolves. 

### validate-inbound-local

Similarly this validates all published cross_links still resolve using a local `links.json` file to override the definition for a given repository. 

This is exposed as a github action:

```
    - name: Validate Inbound Links
      uses: elastic/docs-builder/actions/validate-inbound-local@main
```

Which may be called after a documentation build. Will hook this up to be part of the default publish flow in a follow up PR.

This will **only** validate crosslinks to the current repository. So a failure in one repository **shouldn't** fail all repositories.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -226,6 +226,8 @@ dotnet_diagnostic.IDE0057.severity = none
 
 dotnet_diagnostic.IDE0051.severity = suggestion
 dotnet_diagnostic.IDE0059.severity = suggestion
+dotnet_diagnostic.IDE0055.severity = suggestion
+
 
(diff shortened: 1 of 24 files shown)


reakaleek (approved):
🚀 as always

PR #588, opened 2025-02-24
Fix htmx link in markdown content

reakaleek:
## Changes

Add the `hx-get` attribute to links in the markdown content


Mpdreamz:
The authoring tests can be fixed by adding `hx-get` here: https://github.com/elastic/docs-builder/blob/main/tests/authoring/Framework/HtmlAssertions.fs#L97

reakaleek:
> The authoring tests can be fixed by adding `hx-get` here: https://github.com/elastic/docs-builder/blob/main/tests/authoring/Framework/HtmlAssertions.fs#L97

Thank you for the hint!

Mpdreamz (approved):

PR #630, opened 2025-02-27
On ci skip if /docs folder has no docset.yml silently

Mpdreamz:


reakaleek (approved):

Mpdreamz:
Tested locally @reakaleek 

```
 env GITHUB_ACTIONS=true ./.artifacts/publish/docs-builder/release/docs-builder --path src
info ::e.d.t.f.topwatchFilter:: generate :: Starting...

::set-output name=skip::true
info ::e.d.t.f.topwatchFilter:: generate :: Finished in '00:00:00.0858646'
```

PR #634, opened 2025-02-27
preview-builds: Add `preview-url-path` input

reakaleek:
This way we can set `/reference` for reference repos

### DIFF
--- a/.github/workflows/preview-build.yml
+++ b/.github/workflows/preview-build.yml
@@ -30,6 +30,11 @@ on:
         type: string
         default: 'false'
         required: false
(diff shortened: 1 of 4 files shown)


Mpdreamz (changes requested):
While a quick win, I think its best if `docs-builder` emits this and we add it as an output of the build action. 

Given the number of repositories I think we should do everything we can so that they can run without any arguments and we never have to update them 🤞

reakaleek:
Closing in favour of https://github.com/elastic/docs-builder/pull/642

PR #657, opened 2025-02-28
Share inbound-links commands with `docs-builder`

Mpdreamz:
```bash
docs-builder inbound-links --help
Usage: [command] [options...] [-h|--help] [--version]

Converts a source markdown folder or file to an output folder

Options:
  -p|--path <string?>         Defaults to the`{pwd}/docs` folder (Default: null)
  -o|--output <string?>       Defaults to `.artifacts/html` (Default: null)
  --path-prefix <string?>     Specifies the path prefix for urls (Default: null)
  --force <bool?>             Force a full rebuild of the destination folder (Default: null)
  --strict <bool?>            Treat warnings as errors and fail the build on warnings (Default: null)
  --allow-indexing <bool?>    Allow indexing and following of html files (Default: null)

Commands:
  generate                                 Converts a source markdown folder or file to an output folder
         <para>global options:</para>
         --log-level level
  inbound-links validate                   Validate a single repository against the published cross_links in all published links.json files.
  inbound-links validate-all               Validate all published cross_links in all published links.json files.
  inbound-links validate-link-reference    Validate a locally published links.json file against all published links.json files in the registry
  mv                                       Move a file from one location to another and update all links in the documentation
  serve                                    Continuously serve a documentation folder at http://localhost:3000.
         File systems changes will be reflected without having to restart the server.
```

Allows writers to check the local links.json file after building e.g:

```
$ docs-builder
$ docs-builder inbound-links validate-link-reference
```

The latter command will check the local `.artifacts/docs/html/links.json`'s outbound links against the link registry. 

Allowing writers to shorten their feedback loop.

### DIFF
--- a/src/Elastic.Documentation.Tooling/Diagnostics/Console/ConsoleDiagnosticsCollector.cs
+++ b/src/Elastic.Documentation.Tooling/Diagnostics/Console/ConsoleDiagnosticsCollector.cs
@@ -25,8 +25,12 @@ public class ConsoleDiagnosticsCollector(ILoggerFactory loggerFactory, ICoreServ
 			_errors.Add(diagnostic);
 	}
 
+	private bool _stopped;
 	public override async Task StopAsync(Cancel cancellationToken)
 	{
+		if (_stopped)
+			return;
+		_stopped = true;
 		var repository = new ErrataFileSourceRepository();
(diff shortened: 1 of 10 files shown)


reakaleek (approved):

reakaleek:
Seems like the preview url is broken here.

PR #676, opened 2025-03-05
Add layout system and NotFound layout

reakaleek:
## Changes

- Add a layout attribute in frontmatter
  - This way we can choose a different layout if necessary.
- Add a hidden 404.md file with `layout: NotFound`
- Add `layout: LandingPage` to index.md
- Remove the "landing-page" feature flag and use layout instead

## Result

image

> [!NOTE]  
> The `NotFound` layout only works on production and local development.
> I can only set an absolute path for the error page in cloudfront, which doesn't work well with the preview environment.

For previews we just show a generic 404 page.

image

example: https://docs-v3-preview.elastic.dev/this-does-not-exist

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,6 +1,6 @@
 ---
 navigation_title: Elastic Docs v3
-layout: LandingPage
+layout: landing-page
 ---
 
 # Welcome to Elastic Docs v3
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,6 +1,5 @@
 ---
 navigation_title: Elastic Docs v3
-layout: landing-page
 ---
 
 # Welcome to Elastic Docs v3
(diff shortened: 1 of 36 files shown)


Mpdreamz (approved):
Not: kebab case for layout names in config

reakaleek:
> Not: kebab case for layout names in config

https://github.com/elastic/docs-builder/pull/676/commits/a94fe8e38f4c5c2949e79c7dca719a1ab1fc3929

docs/index.md
```diff
@@ -1,5 +1,6 @@
 ---
 navigation_title: Elastic Docs v3
+layout: landing-page
```
reakaleek, line 3:
```suggestion
```

PR #677, opened 2025-03-05
[docs] Update custom anchor links section in heading syntax docs

colleenmcginnis:
Update custom anchor links section in heading syntax docs to use an example that is consistent with how the migration tool handled custom anchor IDs. 

Note: I think using `[#some-id]` is also valid, but since the migration tool didn't include the `#` I don't think it should be the primary example. If anyone feels strongly about communicating in the docs that both `[#some-id]` and `[some-id]` are valid, I can also add a note.

### DIFF
--- a/docs/syntax/headings.md
+++ b/docs/syntax/headings.md
@@ -85,7 +85,7 @@ You can also specify a custom anchor link using the following syntax.
 
 ```markdown
 
-#### Heading [#custom-anchor]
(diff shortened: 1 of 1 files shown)


leemthompo (approved):
1. Thank you
2. I for one don't feel strongly about mentioning possibility of using the `#` prefix

reakaleek (approved):

Issue #664, opened 2025-03-03
not enough padding around images


shainaraskas:
### Describe the bug

images abut text, making both the surrounding text and image text hard to read, and making the page feel crowded.

<img width="645" alt="Image" src="https://github.com/user-attachments/assets/44d64276-253f-4a4a-b7a8-fd2abf686db5" />

<img width="672" alt="Image" src="https://github.com/user-attachments/assets/ed8f4457-4c33-4f09-b545-e6c813d56697" />

### Expected behavior

images should be prefaced by an empty line

### Tooling

- [ ] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

karenzone:
++

reakaleek:
Might be fixed by https://github.com/elastic/docs-builder/pull/682

reakaleek:
https://github.com/elastic/docs-builder/pull/682 was merged. Could you please confirm if this fixes your use cases? Thank you 🙏

shainaraskas:
@reakaleek yep it's better!

<img width="587" alt="Image" src="https://github.com/user-attachments/assets/23e846f3-cafd-41e4-a322-fd64f2797e40" />

Issue #439, opened 2025-02-06
Images go automatically inline even when on a distinct line or after a line break


florent-leborgne:
### Describe the bug

There are some cases where images show inline when they're not supposed to. This happens to me especially in lists. While there are workarounds (like changing the migrated syntax to regular markdown syntax and realigning the block), this is too time consuming given the number of images we have, and it would be best if we coud just fix the parsing, or bulk transform all image blocks into regular markdown images + align them correctly with their parent list item.

Example of sources (it's the same if we remove the empty line): 

<img width="542" alt="Image" src="https://github.com/user-attachments/assets/184d089f-d876-4c67-af4f-4ecce521843d" />

Renders as: 

<img width="811" alt="Image" src="https://github.com/user-attachments/assets/0cc892a0-1be2-4d26-8859-41bd1e9700ca" />

### Expected behavior

The image renders on a different line.

### Tooling

- [x] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

reakaleek:
https://github.com/elastic/docs-builder/pull/682 was merged. Could you please confirm if this fixes your use cases? Thank you 🙏

florent-leborgne:
The issue seems resolved, thanks!

PR #682, opened 2025-03-05
Fix image styling

reakaleek:
## Context

tailwindcss resets the img to use `display: block` and `vertical-align` middle. Hence, inline images were broken.

In the case of the image directive, which should be `display: block`, the surrounding `a` tag was `display: inline`

## Changes

- This resets the tailwind css reset 
- And set the a tag that surrounds the an image to `display: block`
- Add the default top margin to the `a` tag.

## Related Issues
- https://github.com/elastic/docs-builder/issues/664
- https://github.com/elastic/docs-builder/issues/439


bmorelli25 (approved):
LGTM

Issue #595, opened 2025-02-24
[not exactly a bug] don't use monospace for definition lists


shainaraskas:
### Describe the bug

not all definition lists are for props. we also use definition lists for concepts:

<img width="768" alt="Image" src="https://github.com/user-attachments/assets/335ffeb2-a808-4aae-ac61-93ff8df51b18" />

right now, when we want to specify that something is code in a definition list, we apply our own formatting using backticks:

<img width="772" alt="Image" src="https://github.com/user-attachments/assets/14375a71-3f06-4c30-a4dc-6fcd6b70e575" />

I recommend adjusting the default formatting to use use bold body font, so the component is more flexible.

### Expected behavior

not monospace

### Tooling

- [x] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

PR #689, opened 2025-03-06
Use default font for definition lists

reakaleek:
Closes https://github.com/elastic/docs-builder/issues/595

## Result

image


Mpdreamz (approved):
Lgtm, should they be bold too?

PR #695, opened 2025-03-07
Allow comments between code blocks and callouts

### LINKED ISSUE #694: [Feature Request]: Allow code comments between code block and annotation list
### Prerequisites

@georgewallace is working with the Elasticsearch team to get Elasticsearch code block testing to work with Markdown files. The test looks for code comments with specific keywords after a code block to determine how to test the contents of the code block. However, sometimes we want to test code blocks that use inline annotations, and code blocks with annotations expect a _list_ to come immediately after the code block.

cc @leemthompo @szabosteve 

### Proposed Solution

There are probably a few different ways to solve this problem including:

* Allowing code comments between the code block and the annotation list.
* Collaborating with the Elasticsearch team to pass arguments to their tests in another way that does not use code comments.
* Collaborating with the Elasticsearch team to expect code comments before the code snippet or after the annotation list. 
* Something else? 

### Examples and Research

_No response_

### Alternative Solutions

_No response_

### Additional Context

Here's an example containing code comments between code blocks and the related annotation list: https://github.com/elastic/elasticsearch/compare/main...colleenmcginnis:elasticsearch:add-code-comments#diff-f13f1c308355eb2742e8d445e05b9eb42d0a20f9264f33178ec4bee42a3166c6. 

Nice to have

bmorelli25:
## Problem

Code blocks with annotations require a numbered list to follow them to explain its callouts. However, as pointed out in #694, an issue arises when Elasticsearch test directives (using `% TEST`) are placed between a code block and its annotation list. Our current implementation doesn't differentiate these comments from `HtmlBlock`s. This causes Elasticsearch test comments to error as there are now multiple blocks between a code block and its callouts.

**This PR...**
1. Identifies `CommentBlock`s by checking the block type name
2. Allows any number of comment blocks between a code block and its list, while still enforcing the rule that only one non-comment, non-list block can exist in this position
3. Handles removal of comment blocks
4. Adds tests for the following scenarios:
    * Code block followed by comments
    * Code block followed by multiple types of comments
    * Code block with both comment blocks and a paragraph before the callout list
    * Error cases with too many non-comment blocks
    * Comment blocks with no callout list
    * Comment blocks after the callout list

* Closes #694


src/Elastic.Markdown/Myst/CodeBlocks/EnhancedCodeBlockHtmlRenderer.cs
```diff
@@ -107,6 +107,22 @@ private static int CountIndentation(StringSlice slice)
...
+		if (block.GetType().Name == "CommentBlock")
+		{
+			return true;
+		}
+
+		if (block is HtmlBlock htmlBlock)
+		{
+			var content = htmlBlock.Lines.ToString();
+			return content.TrimStart().StartsWith('%');
+		}
```
reakaleek, line 121: In which cases is this needed?

Out of curiousness I removed this part and ran the tests. And everything is still green.

src/Elastic.Markdown/Myst/CodeBlocks/EnhancedCodeBlockHtmlRenderer.cs
```diff
@@ -107,6 +107,22 @@ private static int CountIndentation(StringSlice slice)
 		return indentCount;
 	}
 
+	private static bool IsCommentBlock(Block block)
+	{
+		if (block.GetType().Name == "CommentBlock")
```
reakaleek, line 112: nit: 
```suggestion
		if (block is CommentBlock)
```
(just for consistency)

reakaleek (approved):
🚀  LGTM

src/Elastic.Markdown/Myst/CodeBlocks/EnhancedCodeBlockHtmlRenderer.cs
bmorelli25: You're right! My first attempt at finding comments was to parse HTML blocks starting with `%` but that didn't work as comments create `CommentBlock`s not `HtmlBlock`s. This should be removed.

bmorelli25:
```suggestion
```

src/Elastic.Markdown/Myst/CodeBlocks/EnhancedCodeBlockHtmlRenderer.cs
bmorelli25: Hmm. It looks like this results in:

```
error CS0246: The type or namespace name 'CommentBlock' could not be found (are you missing a using directive or an assembly reference?)
```

bmorelli25: Reverted for now as I don't know how to fix this code so that `if (block is CommentBlock)` works. Open to any suggestions!

Mpdreamz: Need to add `using Elastic.Markdown.Myst.Comments;`  that will allow the `is` check

Issue #626, opened 2025-02-27
[Feature Request]: Allow some variables in links


colleenmcginnis:
### Prerequisites

- [x] I have searched existing issues to ensure this feature hasn't already been requested
- [x] I have tested using the latest version of docs-builder

### What problem are you trying to solve?

I did read through the conversation in https://github.com/elastic/docs-builder/issues/182, but I _think_ was specific to not using variables for cross-repo links. 

The Beats team [shared another use case](https://github.com/elastic/beats/pull/42897#issuecomment-2686148975) for using variables in URLs that is unrelated to cross-repo links. They [use a variable in links to a third-party site](https://github.com/efd6/beats/blob/9f84d3bc6cbccee3ead09d51389ef2c6859634a9/x-pack/filebeat/docs/inputs/input-cel.asciidoc?plain=1#L4-L5) so that they only have to update one line when they bump the version they are using (instead of having to update ~90 lines). 

### Proposed Solution

Allow some variables in links (or any other approach to addressing the use case described above).

### Alternative Solutions

The alternative would be to just find-and-replace the version in all links in the file. 

### How important is this feature to you?

Nice to have

bmorelli25:
Note: When adding docs, add var substitution in link docs, not string substitution. While string substitution will still be possible, that's not how we want to sell this feature.

reakaleek:
https://github.com/elastic/docs-builder/pull/715 will allow substitutions only in external links.

reakaleek:
https://github.com/elastic/docs-builder/pull/715 was merged.

You can now add substitutions in links. However, it only works if the resolved url starts with http. Hence, it should be a http:// or https:// link

PR #715, opened 2025-03-11
Allow substitutions in external links

reakaleek:
Closes https://github.com/elastic/docs-builder/issues/626

### DIFF
--- a/src/Elastic.Markdown/Myst/InlineParsers/DiagnosticLinkInlineParser.cs
+++ b/src/Elastic.Markdown/Myst/InlineParsers/DiagnosticLinkInlineParser.cs
@@ -106,9 +106,8 @@ public class DiagnosticLinkInlineParser : LinkInlineParser
 			var replacedUrl = url.ReplaceSubstitutions(processor.GetContext());
 			if (replacedUrl.Contains("{{"))
 			{
-				processor.EmitWarning(link,
-					"The url contains a template expression. Please do not use template expressions in links. " +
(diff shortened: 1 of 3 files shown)


src/Elastic.Markdown/Myst/InlineParsers/DiagnosticLinkInlineParser.cs
```diff
@@ -100,6 +101,27 @@ private static void ValidateAndProcessLink(LinkInline link, InlineProcessor proc
 	{
 		var url = link.Url;
 
+		if (url?.Contains("{{") == true)
+		{
+			var replacedUrl = url.ReplaceSubstitutions(processor.GetContext());
+			if (replacedUrl.Contains("{{"))
+			{
+				processor.EmitWarning(link,
+					"The url contains a template expression. Please do not use template expressions in links. " +
```
Mpdreamz, line 110: I think the error message here is misleading, we allow template expressions but this indicates a substitution key was not found and thus left the expression in the url.

Mpdreamz (changes requested):

src/Elastic.Markdown/Myst/InlineParsers/DiagnosticLinkInlineParser.cs
reakaleek: https://github.com/elastic/docs-builder/pull/715/commits/de5933b8cbe88fcbe0d1b6e136536297d00a185c

PR #738, opened 2025-03-14
Adjust path resolution to enable Windows support.

cotti:
On situations where we receive a prebuilt resource path string, an OS check is performed beforehand.

### DIFF
--- a/src/docs-builder/Http/DocumentationWebHost.cs
+++ b/src/docs-builder/Http/DocumentationWebHost.cs
@@ -4,6 +4,7 @@
 
 using System.IO.Abstractions;
 using System.Net;
+using System.Runtime.InteropServices;
 using Documentation.Builder.Diagnostics.LiveMode;
 using Elastic.Documentation.Tooling;
 using Elastic.Markdown;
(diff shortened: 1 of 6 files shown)


src/docs-builder/Http/DocumentationWebHost.cs
```diff
@@ -149,10 +149,13 @@ private static async Task<IResult> ServeDocumentationFile(ReloadableGeneratorSta
 	{
 		var generator = holder.Generator;
 
+		if (Environment.OSVersion.Platform == PlatformID.Win32NT)
```
Mpdreamz, line 152:
```suggestion
RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
```

Mpdreamz (changes requested):
One small nitpick!

src/docs-builder/Http/DocumentationWebHost.cs
cotti: Done!

Mpdreamz (approved):
Get it in :D

PR #744, opened 2025-03-14
Separate static url prefix from path prefix

Mpdreamz:
It still defaults to UrlPathPrefix but with the assembler build we need a way to ensure static files can
be rooted seperately from the path prefix of the page.

To help with development we now have a 'dev' environment which defaults to http://localhost:4000/docs

We can now view the output of `docs-assembler repo build all` with `docs-builder serve-static`
which will host the combined docs under http://localhost:4000

### DIFF
--- a/src/Elastic.Markdown/BuildContext.cs
+++ b/src/Elastic.Markdown/BuildContext.cs
@@ -39,6 +39,15 @@ public record BuildContext
 		init => _urlPathPrefix = value;
 	}
 
+	private readonly string? _staticUrlPathPrefix;
+	public string? StaticUrlPathPrefix
+	{
+		get => !string.IsNullOrWhiteSpace(_staticUrlPathPrefix)
+			? $"/{_staticUrlPathPrefix.Trim('/')}"
+			: UrlPathPrefix;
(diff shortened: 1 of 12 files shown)


src/Elastic.Markdown/BuildContext.cs
```diff
@@ -39,6 +39,15 @@ public string? UrlPathPrefix
 		init => _urlPathPrefix = value;
 	}
 
+	private readonly string? _staticUrlPathPrefix;
+	public string? StaticUrlPathPrefix
+	{
+		get => !string.IsNullOrWhiteSpace(_staticUrlPathPrefix)
+			? $"/{_staticUrlPathPrefix.Trim('/')}"
```
reakaleek, line 46: Do we need to use `Path.DirectorySeparatorChar` here?

```suggestion
			? $"/{_staticUrlPathPrefix.Trim(Path.DirectorySeparatorChar)}"
```

reakaleek (approved):

src/Elastic.Markdown/BuildContext.cs
Mpdreamz: No this are always url paths with `/`

PR #753, opened 2025-03-15
Remove tj-actions usage, eventhough we had it pinned as gitref

Mpdreamz:
This will do a full docs build even for PR's that do not change docs. However because we run reasonably fast the impact is limited. 

We do want to restore some checks so we don't always upload to S3 for the preview environment which might take a bit longer.

NOTE: the actions repo itself now 404's (maybe as a precaution by github)

### DIFF
--- a/src/docs-assembler/Sourcing/RepositorySourcesFetcher.cs
+++ b/src/docs-assembler/Sourcing/RepositorySourcesFetcher.cs
@@ -89,7 +89,7 @@ public class RepositoryCheckoutProvider(ILoggerFactory logger, AssembleContext c
 		{
 			_logger.LogInformation("Pull: {Name}\t{Repository}\t{RelativePath}", name, repository, relativePath);
 			// --allow-unrelated-histories due to shallow clones not finding a common ancestor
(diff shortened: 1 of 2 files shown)


src/docs-assembler/Sourcing/RepositorySourcesFetcher.cs
```diff
@@ -89,28 +89,28 @@ private Checkout CloneOrUpdateRepository(Repository repository, string name, Con
 		{
 			_logger.LogInformation("Pull: {Name}\t{Repository}\t{RelativePath}", name, repository, relativePath);
 			// --allow-unrelated-histories due to shallow clones not finding a common ancestor
-			ExecIn(name, checkoutFolder, "git", "pull", "--depth", "1", "--allow-unrelated-histories", "--no-ff");
+			ExecIn(checkoutFolder, "git", "pull", "--depth", "1", "--allow-unrelated-histories", "--no-ff");
```
Mpdreamz, line 92: unrelated but is a fix for: https://github.com/elastic/docs-internal-workflows/actions/runs/13865451267

reakaleek (approved):

PR #758, opened 2025-03-17
Add content hash as query string to static file paths

### LINKED ISSUE #703: [Feature Request]: Add content hash to asset file names
### Prerequisites

We currently have static asset names like `main.js`, `style.css` etc. This means that when we update these resources, the browser will not know if the cache needs to be invalidated.

### Proposed Solution

Add the content hash as a querystring to the `Model.Static` method.

Make sure to memoize the method.

This way, we can set a higher cache TTL and make sure it's invalidated when the content changes.

### Examples and Research

_No response_

### Alternative Solutions

_No response_

### Additional Context

Check if cloudfront querystring caching is enabled.

Important. The output of the website may be unexpected on subsequent deployments.

Example:

Image

reakaleek:
Closes #703 

image

### DIFF
--- a/src/Elastic.Markdown/Slices/Index.cshtml
+++ b/src/Elastic.Markdown/Slices/Index.cshtml
@@ -17,7 +17,7 @@
 		GithubEditUrl = Model.GithubEditUrl,
 		AllowIndexing = Model.AllowIndexing,
 		Features = Model.Features,
-		GetContentHashFunc = Model.GetContentHashFunc
+		StaticFileContentHashProvider = Model.StaticFileContentHashProvider
 	};
 }
 <section id="elastic-docs-v3">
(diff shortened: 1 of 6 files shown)


src/Elastic.Markdown/Slices/Index.cshtml
```diff
@@ -16,7 +16,8 @@
 		StaticUrlPathPrefix = Model.StaticUrlPathPrefix,
 		GithubEditUrl = Model.GithubEditUrl,
 		AllowIndexing = Model.AllowIndexing,
-		Features = Model.Features
+		Features = Model.Features,
+		GetContentHashFunc = Model.GetContentHashFunc
```
Mpdreamz, line 20: This can just be the `StaticFileContentHashProvider` instance

Mpdreamz (changes requested):
One nit, great work!

src/Elastic.Markdown/Slices/Index.cshtml
reakaleek: https://github.com/elastic/docs-builder/pull/758/commits/d3e876fb0c3b26d8271a6bb61a41162dadb62bb7

PR #763, opened 2025-03-17
Skip preview jobs on fork repositories

reakaleek:
## Changes

This change will make the jobs skip on forks.
But they will still run on the upstream repo, even on PRs from forks.

## Context

This workflow does not work on workflows running on the fork itself because of the aws auth step.

## Related
- https://github.com/elastic/kibana/pull/214886

### DIFF
--- a/.github/workflows/preview-build.yml
+++ b/.github/workflows/preview-build.yml
@@ -45,7 +45,7 @@ permissions:
 
 jobs:
   build:
-    if: github.event.repository.fork == false
+    if: github.event.repository.fork == false # Skip running the job on the fork itself (It still runs on PRs on the upstream from forks)
     concurrency:
       group: ${{ github.workflow }}-${{ github.event.pull_request.head.ref || github.ref }}
       cancel-in-progress: ${{ startsWith(github.event_name, 'pull_request') }}
--- a/.github/workflows/preview-build.yml
+++ b/.github/workflows/preview-build.yml
@@ -45,6 +45,7 @@ permissions:
 
 jobs:
   build:
+    if: github.event.repository.fork == false
     concurrency:
       group: ${{ github.workflow }}-${{ github.event.pull_request.head.ref || github.ref }}
       cancel-in-progress: ${{ startsWith(github.event_name, 'pull_request') }}
--- a/.github/workflows/preview-cleanup.yml
+++ b/.github/workflows/preview-cleanup.yml
@@ -12,6 +12,7 @@ permissions:
 
 jobs:
   destroy:
+    if: github.event.repository.fork == false
     runs-on: ubuntu-latest
     steps:
       - name: Delete GitHub environment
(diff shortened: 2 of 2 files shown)


Mpdreamz (approved):

.github/workflows/preview-build.yml
```diff
@@ -45,6 +45,7 @@ permissions:
 
 jobs:
   build:
+    if: github.event.repository.fork == false
```
Mpdreamz, line 48: Don't we still want to ensure the docs build but only skip the preview environment?

Mpdreamz (changes requested):

.github/workflows/preview-build.yml
reakaleek: IMO, putting the condition here makes it simple.

If we only skip the preview environment the logic becomes more complex. 
It already is pretty complex with all the logical conjunctions.

And I don't think we have a use case for running the build on a forked repository.

I think we should only do it if there is reason for it.

Mpdreamz: We want to ensure the documentation has no build errors and all the links validate for forked PR's as well.

reakaleek: Do you know if there are devs/techwriters creating PRs own their own fork as part of their daily workflow?

IMHO, what we care about is the PR on the upstream repo.

Mpdreamz: Ahh I thought this skipped the job from running on PR builds from forks to the main repository. 

Are you saying this disables the job from running on PR's TO a forked repository only?

reakaleek: Yes. 

I did not test it.. I can still do it though.

But this was also the reason for the original draft in https://github.com/elastic/kibana/pull/214886.

@Ikuni17 wanted to disable the run on his own fork.

Mpdreamz: Roger that! maybe add a comment to highlight this?

Mpdreamz (approved):

.github/workflows/preview-build.yml
reakaleek, line 48:
```suggestion
    if: github.event.repository.fork == false # Skip running the job on the fork itself (It still runs on PRs on the upstream from forks)
```

.github/workflows/preview-cleanup.yml
```diff
@@ -12,6 +12,7 @@ permissions:
 
 jobs:
   destroy:
+    if: github.event.repository.fork == false
```
reakaleek, line 15:
```suggestion
    if: github.event.repository.fork == false # Skip running the job on the fork itself (It still runs on PRs on the upstream from forks)
```

.github/workflows/preview-build.yml
reakaleek: https://github.com/elastic/docs-builder/pull/763/commits/4e4938dc356efbe417053d501b64c99ab376333f

Issue #704, opened 2025-03-10
[Feature Request]: Preload fonts


reakaleek:
### Prerequisites

- [x] I have searched existing issues to ensure this feature hasn't already been requested
- [x] I have tested using the latest version of docs-builder

### What problem are you trying to solve?

When you open the documentation for the first time (no caches), then you can see the font changing (FOUT: flash of unstyled text)

### Proposed Solution

preload the font so it doesn't happen

### How important is this feature to you?

Nice to have

PR #765, opened 2025-03-18
Preload fonts in use as per CSS definitions

cotti:
In order to preload fonts, a hint must be given to the browser in the form of a <link> tag:

`<link rel="preload" href="/path/to/font.woff2" as="font" type="font/woff2" crossorigin>`

This PR includes a cached repository for the font definitions in .css files so we don't need to worry about adding these if fonts change.

Handles https://github.com/elastic/docs-builder/issues/704

### DIFF
--- a/src/Elastic.Markdown/Helpers/Preloader.cs
+++ b/src/Elastic.Markdown/Helpers/Preloader.cs
@@ -1,0 +1,37 @@
+// Licensed to Elasticsearch B.V under one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+using System.Diagnostics;
+using System.Reflection;
+using System.Text.RegularExpressions;
+
+namespace Elastic.Markdown.Helpers;
+
+public static partial class FontPreloader
+{
+	private static List<string>? FontUriCache = null!;
+
+	public static async Task<IEnumerable<string>> GetFontUrisAsync() => FontUriCache ??= await LoadFontUrisAsync();
+	public static async Task<List<string>> LoadFontUrisAsync()
+	{
+		FontUriCache = [];
+		var assembly = Assembly.GetExecutingAssembly();
+		var stylesResourceName = assembly.GetManifestResourceNames().First(n => n.EndsWith("styles.css"));
+
+		using var cssFileStream = new StreamReader(assembly.GetManifestResourceStream(stylesResourceName)!);
+
+		var cssFile = await cssFileStream.ReadToEndAsync();
+		var matches = FontUriRegex().Matches(cssFile);
+
+		foreach (Match match in matches)
+		{
+			if (match.Success)
+				FontUriCache.Add($"/_static/{match.Groups[1].Value}");
+		}
+		return FontUriCache;
+	}
+
+	[GeneratedRegex(@"url\([""']?([^""'\)]+)[""']?\)", RegexOptions.Multiline | RegexOptions.Compiled)]
+	private static partial Regex FontUriRegex();
+}
--- a/src/Elastic.Markdown/Helpers/Preloader.cs
+++ b/src/Elastic.Markdown/Helpers/Preloader.cs
@@ -14,7 +14,7 @@ public static partial class FontPreloader
 	private static IReadOnlyCollection<string>? FontUriCache = null!;
 
 	public static async Task<IReadOnlyCollection<string>> GetFontUrisAsync() => FontUriCache ??= await LoadFontUrisAsync();
-	public static async Task<IReadOnlyCollection<string>> LoadFontUrisAsync()
+	private static async Task<IReadOnlyCollection<string>> LoadFontUrisAsync()
 	{
 		var cachedFontUris = new List<string>();
 		var assembly = Assembly.GetExecutingAssembly();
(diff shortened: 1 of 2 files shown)


src/Elastic.Markdown/Helpers/Preloader.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using System.Diagnostics;
+using System.Reflection;
+using System.Text.RegularExpressions;
+
+namespace Elastic.Markdown.Helpers;
+
+public static partial class FontPreloader
+{
+	private static List<string>? FontUriCache = null!;
```
Mpdreamz, line 13: readonly field or getter only property. 

Also IReadOnlyCollection instead of List.

src/Elastic.Markdown/Helpers/Preloader.cs
```diff
@@ -0,0 +1,37 @@
...
+
+	public static async Task<IEnumerable<string>> GetFontUrisAsync() => FontUriCache ??= await LoadFontUrisAsync();
+	public static async Task<List<string>> LoadFontUrisAsync()
```
Mpdreamz, line 16: Can be made private

src/Elastic.Markdown/Helpers/Preloader.cs
```diff
@@ -0,0 +1,37 @@
...
+
+		foreach (Match match in matches)
+		{
+			if (match.Success)
+				FontUriCache.Add($"/_static/{match.Groups[1].Value}");
+		}
+		return FontUriCache;
+	}
+
+	[GeneratedRegex(@"url\([""']?([^""'\)]+)[""']?\)", RegexOptions.Multiline | RegexOptions.Compiled)]
```
Mpdreamz, line 35: Maybe stricter including `src:` key?

Mpdreamz (changes requested):
Small nitpicks, other than that lgtm.

src/Elastic.Markdown/Helpers/Preloader.cs
```diff
@@ -0,0 +1,37 @@
...
+
+		using var cssFileStream = new StreamReader(assembly.GetManifestResourceStream(stylesResourceName)!);
+
+		var cssFile = await cssFileStream.ReadToEndAsync();
+		var matches = FontUriRegex().Matches(cssFile);
+
+		foreach (Match match in matches)
+		{
+			if (match.Success)
+				FontUriCache.Add($"/_static/{match.Groups[1].Value}");
```
reakaleek, line 30: I think we need to add the url path prefix here.

image

But nice! I can see it's the first being loaded.

src/Elastic.Markdown/Slices/Layout/_Head.cshtml
```diff
@@ -1,7 +1,12 @@
 @inherits RazorSlice<LayoutViewModel>
+@using Elastic.Markdown.Helpers
 <head>
 	<title>@Model.Title</title>
-	<link rel="stylesheet" type="text/css" href="@Model.Static("styles.css")"/>
+	<link rel="stylesheet preload" as="style" type="text/css" href="@Model.Static("styles.css")" crossorigin/>
```
reakaleek, line 5: Should it be just `rel="preload"` ?

src/Elastic.Markdown/Slices/Layout/_Head.cshtml
```diff
@@ -1,7 +1,12 @@
...
+	@foreach (var fontFile in await FontPreloader.GetFontUrisAsync())
```
reakaleek, line 6: Can we put this before the stylesheet?
Does it even make sense to put before the stylesheet?

src/Elastic.Markdown/Helpers/Preloader.cs
reakaleek: What do you think about adding also the file extension here.

the url function could in theory load any kind of file.

src/Elastic.Markdown/Helpers/Preloader.cs
reakaleek: Does this need to be thread safe?

reakaleek: Ah. Nevermind. If I understood correctly, the whole logic is only executed once.

src/Elastic.Markdown/Slices/Layout/_Head.cshtml
cotti: CSS will break if the link isn't defined as `stylesheet` there. If anything, we can remove `preload`, as it being non-blocking for something that should stall page render if not available makes less sense a bit...

cotti: (And it kinda is still blocking nonetheless)

src/Elastic.Markdown/Slices/Layout/_Head.cshtml
cotti: I've seen a bit of both, but it's safe to put them first. Doing that!

src/Elastic.Markdown/Slices/Layout/_Head.cshtml
reakaleek: All good. If this works how it's supposed to work. 
I was just unaware of the syntax and wanted to ask.

Thank you for the explanation!

PR #772, opened 2025-03-18
Fix layout and horizontal scrollable tables

reakaleek:
## Context

The content overflows if there is a table that is wider than the main content column.

## Details

Fixes the broken layout caused by wide tables which at the same time fixes the horizontal scrolling of tables at certain widths.

## Previous behaviour

https://github.com/user-attachments/assets/27d67248-88d8-4039-94ce-af08c0a4784e

## Fixed behaviour

https://github.com/user-attachments/assets/73404a56-815c-47eb-9c19-d4c84633e4c5

### DIFF
--- a/src/Elastic.Markdown/Assets/markdown/table.css
+++ b/src/Elastic.Markdown/Assets/markdown/table.css
@@ -1,24 +1,27 @@
 @layer components {
 	.table-wrapper {
 
-		@apply max-w-full overflow-x-scroll my-4;
+		@apply w-full overflow-x-auto my-4;
 
 		table {
-			@apply w-full border-collapse border-1 border-grey-20;
+			@apply w-full border-collapse border-1 border-grey-30;
 		}
 		
 		th, td {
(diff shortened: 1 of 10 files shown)


Mpdreamz (approved):
I like that the full window now resizes down correctly. 

I dont think the table should be allowed to extend beyond its containers max width. E.g we now need to horizontally scroll to see the table as it extends beyond the max width of it's containers (2nd video).

reakaleek:
> I like that the full window now resizes down correctly. 
> 
> I dont think the table should be allowed to extend beyond its containers max width. E.g we now need to horizontally scroll to see the table as it extends beyond the max width of it's containers (2nd video). 

The horizontal scroll was actually implement before also. It was just broken at certain widths.

The main use case for it is the mobile view, which is a common responsive design pattern.

Thank you for the approval. I will create a GH discussion to discuss exactly this and and a possible strategy to handle large tables in the future.

reakaleek:
https://github.com/elastic/docs-builder/discussions/773

PR #774, opened 2025-03-19
Add table of contents scope management and refactor diagnostics

Mpdreamz:
Introduced `ITableOfContentsScope` to manage scope boundaries, improving link validation and table of contents processing. Updated diagnostics to emit hints for images outside the ToC scope and refactored functions for better reusability. Adjusted related tests and internal structures accordingly.

This attaches the scope (defining toc/docset.yml file)'s directory to `MarkdownFile` and others. 

Use case for this is that we want to error when folks include content on a page that does not live in the scope directory.

Since the scope directory may be relocated by the assembler it will lead to breaks.

image

Implemented as a hint for now to not break anyones builds while we work to update all repositories.

cc @shainaraskas @colleenmcginnis @lcawl

### DIFF
--- a/src/Elastic.Markdown/IO/Configuration/ITocItem.cs
+++ b/src/Elastic.Markdown/IO/Configuration/ITocItem.cs
@@ -15,4 +15,5 @@ public record FileReference(ITableOfContentsScope TableOfContentsScope, string P
 public record FolderReference(ITableOfContentsScope TableOfContentsScope, string Path, bool Found, bool InNav, IReadOnlyCollection<ITocItem> Children)
 	: ITocItem;
 
-public record TocReference(string Path, bool Found, bool InNav, IReadOnlyCollection<ITocItem> Children) : FolderReference(Path, Found, InNav, Children);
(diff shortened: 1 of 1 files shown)


src/Elastic.Markdown/IO/MarkdownFile.cs
```diff
@@ -224,9 +230,9 @@ protected void ReadDocumentInstructions(MarkdownDocument document)
 
 		var toc = GetAnchors(_set, MarkdownParser, YamlFrontMatter, document, subs, out var anchors);
 
-		_tableOfContent.Clear();
+		_pageTableOfContent.Clear();
```
reakaleek, line 233: While reading some navigation UX patterns I encountered the the term "On Page Table of Contents", which I think is very clear.

reakaleek: I think it removes a lot of ambiguity, since in the HTML it also says "On this page"

reakaleek (approved):
LGTM

PR #775, opened 2025-03-19
Fix/font preload static path

cotti:
The original implementation for preloading fonts did not take into account that builds could reside outside the root of the URL. This PR suggests acquiring from the Model the current static prefix, already in use by styles.css.


src/Elastic.Markdown/Slices/Layout/_Head.cshtml
```diff
@@ -2,7 +2,7 @@
 @using Elastic.Markdown.Helpers
 <head>
 	<title>@Model.Title</title>
-	@foreach (var fontFile in await FontPreloader.GetFontUrisAsync())
+	@foreach (var fontFile in await FontPreloader.GetFontUrisAsync(@Model.Static(string.Empty)))
```
Mpdreamz, line 5: Use model.urlpathprefix?

Mpdreamz (changes requested):

Mpdreamz (approved):

PR #778, opened 2025-03-19
Add ability to set description meta tag

reakaleek:
## Details

You can now add the `description` field to the FrontMatter block on every document.
If the field is not present, then it will be autogenerated based on the content.


src/Elastic.Markdown/Slices/DescriptionGenerator.cs
```diff
@@ -0,0 +1,132 @@
...
+			if (endIndex == -1)
+				endIndex = MaxLength;
+			result = string.Concat(result.AsSpan(0, endIndex + 1).Trim(), "...");
+		}
+
+		return result;
+	}
+
+#pragma warning disable IDE0060
+	private static void ProcessIncludeBlock(IncludeBlock include, StringBuilder description) => Console.WriteLine("Not implemented: ProcessIncludeBlock");
```
Mpdreamz, line 63:
```suggestion
	private static void ProcessIncludeBlock(IncludeBlock include, StringBuilder description) {}
```

Or use our logging infrastructure.

Mpdreamz: You could use discard parameters to not have the `#pragma`

reakaleek: Yeah. Interestingly, I had an error saying something like `Duplicate arg "_" something something`.

But I'm going to clean this up and just add a comment instead.

reakaleek: https://github.com/elastic/docs-builder/pull/778/commits/1f619098daff40ac053c937bfc6d5e6fc0d78bae

Mpdreamz (approved):

PR #781, opened 2025-03-19
Refactor navigation to be more anemic.

Mpdreamz:
This allows us to inject navigation using the assembler build (not yet implemented) without having to resolve all markdown files.

This still supports the PrimaryNav feature flag for local builds.

https://github.com/user-attachments/assets/fc97651a-d1e5-4c95-9770-c2e37012fa10


reakaleek (approved):
Thanks for cleaning up!

reakaleek:
I just noticed this introduced a visual bug.
Here is a potential fix: https://github.com/elastic/docs-builder/pull/787

reakaleek:
This is maybe also related: https://github.com/elastic/docs-builder/issues/788

PR #787, opened 2025-03-20
Fix pages navigation appearance

reakaleek:
## Context

The appearance of the pages navigation is broken.

image

## Changes

This determines correctly if the navigation is the overall root navigation, which is needed to determine the styling based on the depth of a `FileNavigation` or `GroupNavigation`.

### DIFF
--- a/src/Elastic.Markdown/Slices/HtmlWriter.cs
+++ b/src/Elastic.Markdown/Slices/HtmlWriter.cs
@@ -14,7 +14,7 @@ namespace Elastic.Markdown.Slices;
 
 public interface INavigationHtmlWriter
 {
-	Task<string> RenderNavigation(INavigation currentRootNavigation, bool isRoot, Cancel ctx = default);
+	Task<string> RenderNavigation(INavigation currentRootNavigation, Cancel ctx = default);
 
 	async Task<string> Render(NavigationViewModel model, Cancel ctx)
 	{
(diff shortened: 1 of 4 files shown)


src/Elastic.Markdown/Slices/HtmlWriter.cs
```diff
@@ -81,7 +79,8 @@ private async Task<string> RenderLayout(MarkdownFile markdown, MarkdownDocument
 		var html = markdown.CreateHtml(document);
 		await DocumentationSet.Tree.Resolve(ctx);
 
-		var navigationHtml = await NavigationHtmlWriter.RenderNavigation(markdown.RootNavigation, ctx);
+		var isRoot = DocumentationSet.Tree.Id == markdown.RootNavigation.Id;
```
Mpdreamz, line 82: Can this check live on this deleted line? 

https://github.com/elastic/docs-builder/pull/787/files#diff-d551730a87f5d4489d8ecb7870729ef7ac6bbec2b67e1fd2cdce4ee8a0243b17L52

The goal of `INavigationHtmlWriter` was to isolate the usage of `DocumentationSet.Tree` so that the assembler implementation of it does not have to rely on it (or an outside check for root).

Mpdreamz (changes requested):
Thanks for cleaning up after me! 

One small feedback

src/Elastic.Markdown/Slices/HtmlWriter.cs
reakaleek: I figure I can also use `isPrimaryNavEnabled` https://github.com/elastic/docs-builder/pull/787/commits/502e04ac75840e52fa35355df47b4817e4622ec0

Will this also work for the assembler?

reakaleek: In theory this will make the pages nav broken on the root page `/`. But since the landing page doesn't have a navigation it's ok.

Mpdreamz: So clean! Yeah assembler will enable primary nav for all 👍

Mpdreamz (approved):

Issue #310, opened 2025-01-22
"Edit this page" not redirecting correctly on local build


thekofimensah:
### Describe the bug

When on a page and you'd like to edit directly, the "Edit this page" button redirects to github but with a 404.

This only happens when I try to deploy locally, not using the pre-built version.

### Expected behavior

It should redirect to the correct git page and allow editing

### Steps to reproduce

Go to http://localhost:3000/search-your-data.html

Redirect points to

https://github.com/elastic/unavailable/edit/elasticsearch.md/docs/search-your-data.md

[Edit this page](https://github.com/elastic/unavailable/edit/elasticsearch.md/docs/search-your-data.md)

However in the https://bookish-dollop-9p66716.pages.github.io/search-your-data.html

The redirect points to [Edit this page](https://github.com/elastic/elasticsearch.md/edit/main/docs/search-your-data.md)

### Tooling

- [x] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

Mpdreamz:
Hey @thekofimensah what do 

```
git remote -vv
git branch -vv
ls .git/refs/heads/
```

output for you?

thekofimensah:
I was out sick, this is the output of the 3 commands


```
~/Desktop/elastic_co_work/elasticsearch.md main ❯ git remote -vv     12:51:31 PM

origin  https://github.com/elastic/elasticsearch.md.git (fetch)
origin  https://github.com/elastic/elasticsearch.md.git (push)
~/Desktop/elastic_co_work/elasticsearch.md main ❯ git branch -vv     12:51:45 PM

* main 81f09b7 [origin/main] update 1/19
~/Desktop/elastic_co_work/elasticsearch.md main ❯ ls .git/refs/heads/

total 8
drwxr-xr-x  3 kofi  staff    96B Jan 20 20:46 .
drwxr-xr-x  5 kofi  staff   160B Jan 20 20:46 ..
-rw-r--r--  1 kofi  staff    41B Jan 20 20:46 main
```

bmorelli25:
@Mpdreamz I'm able to replicate this. When building the docs-content repo, for example:

Clicking "edit" on this page `http://localhost:3000/solutions/index.html`
Links me to https://github.com/elastic/docs-builder/edit/docs-content/solutions/index.md

PR #790, opened 2025-03-20
Fix "Edit This File" path

### LINKED ISSUE #769: Edit this page links don't work on staging
### Describe the bug

Go to any page: https://staging-website.elastic.co/docs/
Click [Edit this page](https://github.com/elastic/docs-internal-workflows/edit/main/.artifacts/checkouts/docs-content/solutions/observability/apps.md)
Get taken to: https://github.com/elastic/{{internal-repo}}/edit/main/.artifacts/checkouts/docs-content/solutions/observability/apps.md

### Expected behavior

n/a

### Steps to reproduce

_No response_

### Tooling

- [ ] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

cotti:
This issue had two facets.

The first one, which was reported previously, is that certain projects had broken links to edit their files on Github, particularly on local deployment.

However, upon digging further into the current state of this link, it was found that most links were also broken due to incorrect branch detection. The second facet is this - we should look for .git/config inside the project's root folder rather than its docs/ folder.

Fixes #769 and #310

### DIFF
--- a/src/Elastic.Markdown/BuildContext.cs
+++ b/src/Elastic.Markdown/BuildContext.cs
@@ -72,8 +72,7 @@ public record BuildContext
 		if (ConfigurationPath.FullName != DocumentationSourceDirectory.FullName)
 			DocumentationSourceDirectory = ConfigurationPath.Directory!;
 
-
-		Git = GitCheckoutInformation.Create(DocumentationSourceDirectory, ReadFileSystem);
+		Git = GitCheckoutInformation.Create(rootFolder, ReadFileSystem);
 		Configuration = new ConfigurationFile(this);
 	}
 
--- a/src/Elastic.Markdown/BuildContext.cs
+++ b/src/Elastic.Markdown/BuildContext.cs
@@ -16,10 +16,11 @@ public record BuildContext
 	public IFileSystem ReadFileSystem { get; }
 	public IFileSystem WriteFileSystem { get; }
 
+	public IDirectoryInfo? DocumentationCheckoutDirectory { get; }
 	public IDirectoryInfo DocumentationSourceDirectory { get; }
 	public IDirectoryInfo DocumentationOutputDirectory { get; }
 
-	public ConfigurationFile Configuration { get; set; }
+	public ConfigurationFile Configuration { get; }
 
 	public IFileInfo ConfigurationPath { get; }
 
@@ -50,10 +51,14 @@ public record BuildContext
 	public BuildContext(DiagnosticsCollector collector, IFileSystem fileSystem)
 		: this(collector, fileSystem, fileSystem, null, null) { }
 
-	public BuildContext(DiagnosticsCollector collector, IFileSystem readFileSystem, IFileSystem writeFileSystem)
-		: this(collector, readFileSystem, writeFileSystem, null, null) { }
-
-	public BuildContext(DiagnosticsCollector collector, IFileSystem readFileSystem, IFileSystem writeFileSystem, string? source, string? output)
+	public BuildContext(
+		DiagnosticsCollector collector,
+		IFileSystem readFileSystem,
+		IFileSystem writeFileSystem,
+		string? source = null,
+		string? output = null,
+		GitCheckoutInformation? gitCheckoutInformation = null
+	)
 	{
 		Collector = collector;
 		ReadFileSystem = readFileSystem;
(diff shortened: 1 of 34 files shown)


src/Elastic.Markdown/IO/Discovery/GitCheckoutInformation.cs
```diff
@@ -68,7 +68,7 @@ public static GitCheckoutInformation Create(IDirectoryInfo source, IFileSystem f
 			branch = branch.Replace("ref: ", string.Empty);
 		}
 		else
-			branch = "detached/head";
+			branch = Environment.GetEnvironmentVariable("GITHUB_REF_NAME") ?? "detached/head";
```
reakaleek, line 71: Is this also used for PR previews?

src/Elastic.Markdown/BuildContext.cs
```diff
@@ -72,8 +72,7 @@ public BuildContext(DiagnosticsCollector collector, IFileSystem readFileSystem,
 		if (ConfigurationPath.FullName != DocumentationSourceDirectory.FullName)
 			DocumentationSourceDirectory = ConfigurationPath.Directory!;
 
-
-		Git = GitCheckoutInformation.Create(DocumentationSourceDirectory, ReadFileSystem);
+		Git = GitCheckoutInformation.Create(rootFolder, ReadFileSystem);
```
Mpdreamz, line 75: This was `SourceDirectory` because assembler cannot rely on `rootFolder` 

There is some traversing logic on line 50 in `GitCheckoutInformation.cs` to attempt to get the relative .git folder so not sure if this is needed. 

Long term `CheckoutDirectory` needs to be explicitly set in `BuildContext` vs relying on `Paths.Root`. (not needed as part of this PR).

Can you validate `DocumentationSourceDirectory` indeed does not work with the traversal logic? 

e.g in docs-content the docs are at the root of `DocumentationSourceDirectory` so is the `.git` folder. For most other's it actually lives on folder up from `docs`.

src/Elastic.Markdown/Slices/HtmlWriter.cs
```diff
@@ -88,7 +88,7 @@ private async Task<string> RenderLayout(MarkdownFile markdown, MarkdownDocument
 
 		var remote = DocumentationSet.Build.Git.RepositoryName;
 		var branch = DocumentationSet.Build.Git.Branch;
-		var path = Path.Combine(DocumentationSet.RelativeSourcePath, markdown.RelativePath);
+		var path = Path.Combine(DocumentationSet.RelativeSourcePath.Replace($"../{remote}", string.Empty), markdown.RelativePath).TrimStart(Path.DirectorySeparatorChar);
```
Mpdreamz, line 91: I was surprised `RelativeSourcePath` contains `../{remote}` but now I understand it: 
image


This happens when running locally with `--path` `SourceDirectory` reflects that but `Paths.Root` does not.


I will add `BuildContext.CheckoutDirectory` today so longer term we don't have to do replaces on paths which we really shouldn't do anywhere.

Mpdreamz (approved):

src/Elastic.Markdown/Slices/HtmlWriter.cs
Mpdreamz: Created https://github.com/elastic/docs-builder/pull/791 to aid with this PR

src/Elastic.Markdown/BuildContext.cs
Mpdreamz: See also: https://github.com/elastic/docs-builder/pull/791 which takes away this ambiguity.

src/Elastic.Markdown/Slices/HtmlWriter.cs
cotti: Yep, that solved it!

src/Elastic.Markdown/IO/Discovery/GitCheckoutInformation.cs
cotti: Yeah, it resolved to main on PR previews. I added an environment variable for ${{ github.event.pull_request.head.ref }}, and kept GITHUB_REF_NAME as a fallback value. We'd need the workflow file updated outside this though...?

reakaleek: Yeah, this change needs to in `main` already because `pull_request_target` operates in the context of the main branch.

reakaleek (approved):

PR #799, opened 2025-03-20
Includes GITHUB_PR_REF_NAME as a job-level environment variable.

cotti:
This variable was not being propagated in the workflow, causing the branch reference to fallback to main.


reakaleek (approved):
LGTM.

Can you please remove it from the step level?

Thank you 🙏

reakaleek (approved):

Issue #482, opened 2025-02-12
Copyright notice: symbol missing, date incorrect


marciw:
### Describe the bug

In the copyright notice in the v3 docs footer, the copyright symbol &copy; is missing and the date is 2024: 
<img width="316" alt="Image" src="https://github.com/user-attachments/assets/0145abb3-72b6-49fc-99a4-c22532b70650" />

### Expected behavior

I would expect this to be:

**&copy; 2025 Elasticsearch B.V. All Rights Reserved.**

(not bold; just making it obvious here)

Note: I'm pretty sure we don't need the comma after the year (the website has a period there, which seems incorrect instead of just unnecessary). I have an open email thread with Legal to confirm the punctuation and will update this issue when I've heard from them. 

But we could add the symbol and fix the date in the meantime

### Steps to reproduce

Open a docs page and look at the footer

### Tooling

- [ ] docs-builder
- [ ] migration tooling
- [x] I'm not sure

marciw:
Legal confirmed that this is correct for the footer:

-/-/-/-/-/-/-/-/-/-
© 2025 Elasticsearch B.V. All Rights Reserved.
-/-/-/-/-/-/-/-/-/-

👉 No comma or period after the date

They also provided this fuller version if we have a spot for that, maybe the main docs landing page or some sort of "about":

-/-/-/-/-/-/-/-/-/-
© 2025 Elasticsearch B.V. All Rights Reserved. Copying, publishing and/or distributing without written permission is strictly prohibited.
-/-/-/-/-/-/-/-/-/-

The fuller version would be on just one page or a few high-value pages; it's an anti-pattern for the footer.

(separator lines in this comment  -/-/- are just for readability here)

PR #824, opened 2025-03-24
Add footer and fix landing page layout

reakaleek:
Closes https://github.com/elastic/docs-builder/issues/482


cotti:
I was wondering about the darker section with the footer - if it'd be included for this issue.

The behavior here is to not let the left nav expand its height even when the actual content would allow it to, so it always has a scrollbar if there are enough items. Is it supposed to be like this?

cotti (approved):

PR #835, opened 2025-03-25
GlobalNavigationPathProvider: change definition for the newRelativePath

### LINKED ISSUE #830: [assembler] Pages not found when built locally
### Describe the bug

First reported in https://github.com/elastic/docs-builder/pull/780#issuecomment-2748555840

The global navigation builds without errors and all items are present, but clicking on some links results in "page can't be found". 

https://github.com/user-attachments/assets/c8b03e5c-bb8d-4403-b01f-defeb8988006

Initially thought this was caused by the lack of `toc_max_depth: 2` in the elasticsearch repo because most of the links with this problem were from the elasticsearch repo. However, I'm still seeing this problem even after merging https://github.com/elastic/elasticsearch/pull/125509.

After poking around a bit, I noticed that in `.artifacts/assembly/docs` some of the sub directory names look like part of the name is cut off at the beginning, but I couldn't deduce what the pattern was. I'm not sure if this is just something with my local setup or if this happens for others.

Image

But, if you do visit the incorrect path (for example, `http://localhost:4000/docs/reference/elasticsearch/-apis/`) the page is there, but it doesn't have the global nav.

Image

### Expected behavior

I expect that I can visit these pages by clicking the links in the global nav. 

### Steps to reproduce

1. Build `docs-builder` and `docs-assembler` locally.
2. Run `docs-assembler repo clone-all`.
3. Run `docs-assembler repo build-all`.
4. Run `docs-builder serve-static`.
5. Go to http://localhost:4000/docs/reference/elasticsearch/
6. Try clicking the [Configuration](http://localhost:4000/docs/reference/elasticsearch/configuration-reference/) link in the global nav.
    * Do you get page not found?
    * Is there content at http://localhost:4000/docs/reference/elasticsearch/onfiguration-reference/? 
7. Go back and try clicking [JVM settings](http://localhost:4000/docs/reference/elasticsearch/jvm-settings) in the global nav.
    * Does the page render as expected?

### Tooling

- [ ] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

cotti:
Fixes #830 

The issue is a subtle behavior detail on `TrimStart()` (and `TrimEnd()`). Rather than acting akin to `Substring()`, it continues to try and remove more of the input being manipulated as long as a character belongs in the trim set, until it finds a character not belonging to it.

Anyhow, the default behavior has been changed to verify if it starts with the originalPath, and if it does, perform a `Slice()` instead.

### DIFF
--- a/src/docs-assembler/Navigation/GlobalNavigationPathProvider.cs
+++ b/src/docs-assembler/Navigation/GlobalNavigationPathProvider.cs
@@ -98,7 +98,11 @@ public record GlobalNavigationPathProvider : IDocumentationFileOutputProvider
 		}
 
 		var originalPath = Path.Combine(match.Host, match.AbsolutePath.Trim('/')).TrimStart('/');
-		var newRelativePath = relativePath.AsSpan().TrimStart(originalPath).TrimStart('/').ToString();
+		var relativePathSpan = relativePath.AsSpan();
+		var newRelativePath = relativePathSpan.StartsWith(originalPath, StringComparison.Ordinal)
+			? relativePathSpan.Slice(originalPath.Length).TrimStart('/').ToString()
+			: relativePathSpan.TrimStart(originalPath).TrimStart('/').ToString();
+
 		var path = fs.Path.Combine(outputDirectory.FullName, toc.SourcePathPrefix, newRelativePath);
 
 		return fs.FileInfo.New(path);


reakaleek (approved):
LGTM. Nice find!

src/docs-assembler/Navigation/GlobalNavigationPathProvider.cs
```diff
@@ -98,7 +98,11 @@ public GlobalNavigationPathProvider(GlobalNavigationFile navigationFile, Assembl
 		}
 
 		var originalPath = Path.Combine(match.Host, match.AbsolutePath.Trim('/')).TrimStart('/');
-		var newRelativePath = relativePath.AsSpan().TrimStart(originalPath).TrimStart('/').ToString();
+		var relativePathSpan = relativePath.AsSpan();
```
Mpdreamz, line 101: LGTM! We have a few misuses of `TrimStart()` on `Span<T>` 😢 

Maybe as a follow up PR put this in a new extension method `TrimStartExact()` and update other places we misused `TrimStart()`.

Mpdreamz (approved):

PR #844, opened 2025-03-25
Path prefix duplication detection in navigation.yml

Mpdreamz:
Add build diagnostic error when `navigation.yml` contains duplicate `path_prefix` definitions.

Detects same prefix in different repositories:

image

And within the same repository:

image


cotti (approved):
LGTM

PR #846, opened 2025-03-25
Add ability to set different Google Tag Manager snippets for each environment

reakaleek:


src/docs-assembler/assembler.yml
```diff
@@ -3,11 +3,31 @@ environments:
     uri: https://www.elastic.co
     path_prefix: docs
     allow_indexing: false
-    enable_google_tag_manager: true
+    google_tag_manager:
+      enabled: true
+      script: |
+        <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
```
Mpdreamz, line 9: Can we break this up into the actual variable bits? 

It will be more tedious to configure but since the `navigation.yml` file might end up living in `docs-content` I am weary of raw including HTML from "user" input.

Mpdreamz (changes requested):

src/docs-assembler/assembler.yml
Mpdreamz: `assembler.yml` too 😸

reakaleek: SGTM.

https://github.com/elastic/docs-builder/pull/846/commits/24b955f422664d1af1fa1ed7c45b3b709765c88c

Mpdreamz (approved):
LGTM!

PR #850, opened 2025-03-26
Extend navigation validate command to check publish links

Mpdreamz:
Ensures we don't create configurations in assembler.yml conflicting with already published pages.

image

cc @lcawl @colleenmcginnis 

See also #849 

We will check for conflicts both on individual documentation sets PR's as well as when we make changes to `assembler.yml`


Mpdreamz:
@colleen the check is now updated to not use the published links relative paths verbatim but their resolved (against `navigation.yml`) counter part with the correct path_prefix defined in `navigation.yml`

This now happens if I try to create `reference/elasticsearch` in `docs-content`.

image

reakaleek (approved):

PR #857, opened 2025-03-26
Add elastic/detection-rules to assembler navigation.yml

colleenmcginnis:
Adds prebuilt rule docs to `assembler.yml` and `navigation.yml`.

With this configuration, I'm getting 58 errors like these:

```
error::e.d.t.d.Log           :: No toc for output path: 'detection-rules://../rules/cross-platform/defense_evasion_agent_spoofing_mismatched_id.toml' falling back to: '/Users/colleenmcginnis/GitHub/docs-builder/.artifacts/assembly/docs/_failed/detection-rules/../rules/cross-platform/defense_evasion_agent_spoofing_mismatched_id.toml' (/Users/colleenmcginnis/GitHub/docs-builder/src/docs-assembler/navigation.yml:0)
```

```
error::e.d.t.d.Log           :: Uncaught exception while processing file
System.InvalidOperationException: Expected a DocumentationGroup
   at Documentation.Assembler.Navigation.GlobalNavigationHtmlWriter.<RenderNavigation>d__9.MoveNext() + 0x300
--- End of stack trace from previous location ---
   at Elastic.Markdown.Slices.HtmlWriter.<RenderLayout>d__13.MoveNext() + 0x160
--- End of stack trace from previous location ---
   at Elastic.Markdown.Slices.HtmlWriter.<WriteAsync>d__14.MoveNext() + 0x2c8
--- End of stack trace from previous location ---
   at Elastic.Markdown.Extensions.DetectionRules.RuleDocumentationFileExporter.<ProcessFile>d__4.MoveNext() + 0x2ac
--- End of stack trace from previous location ---
   at Elastic.Markdown.DocumentationGenerator.<ProcessFile>d__25.MoveNext() + 0x60
--- End of stack trace from previous location ---
   at Elastic.Markdown.DocumentationGenerator.<>c__DisplayClass22_0.<<ProcessDocumentationFiles>b__0>d.MoveNext() + 0x80 (../rules/windows/persistence_dontexpirepasswd_account.toml:0)
```

You should be able to pull down this branch and clone and build to see the full list of errors.


Mpdreamz:
@colleenmcginnis this now almost works:

image

The placement in the global navigation is not great but urls resolve and the output paths are in the right place.

The urls end up looking like:

http://localhost:4000/docs/reference/security/prebuilt-rules/rules/network/command_and_control_accepted_default_telnet_port_connection

Mpdreamz:
I wonder if security rules should just have an overview page with links instead of injecting 2000 items into the navigation. 

cc @shainaraskas @jmikell821

shainaraskas:
> I wonder if security rules should just have an overview page with links instead of injecting 2000 items into the navigation.

@Mpdreamz I assume you mean making them hidden? if we hide them, then google won't index them either. not sure if we want that

Mpdreamz:
No more that we make `detection` rules a landing page with links to all the rules vs having them all load in the `reference` left menu. 

Expanding it is a quick way to get lost on the left menu too.

colleenmcginnis:
I think these are questions @jmikell821 needs to answer.

Mpdreamz:
It might also align better with follow up work @approksiu wants as well: creating a filterable overview page to navigate the rule sets.

jmikell821:
@Mpdreamz is it the long list of rules in the left nav that's causing a problem? I think a landing page with a link to rules is a good compromise for now, but I'll confirm with @approksiu. We did discuss previously creating a filterable view similar to this here: https://elastic.github.io/detection-rules-explorer/ but given our approaching deadline, I don't know if this is feasible right now?

approksiu:
@Mpdreamz Landing page is a good idea, I agree the long list of rules is a problem. Please let me know how the landing page would look. Thanks!

Mpdreamz:
@approksiu for the first version it would be a minimal listing page I could potentially group them under some headers if you prefer? if you do what would be the preferred grouping key?

approksiu:
Great work @Mpdreamz !

colleenmcginnis:
Looks good to me! One question:

>Rules are hidden from the navigation

Are the individual rule pages still set up to be indexed by search engines?

Mpdreamz:
> Are the individual rule pages still set up to be indexed by search engines? 

Yes they are still crawl-able however I just realized they won't be in our `sitemap.xml`

Mpdreamz (approved):

PR #864, opened 2025-03-27
Optimize hljs

reakaleek:
## Details

Only install and register a subset of languages to shrink the final JS bundle size.

The list of languages was determined by the following command on assembler build artifacts.

```
find . -type f -name "*.html" -exec grep -ho "class=\"language-[^ \"']*" {} \; | sort -u
```

## Result

Bundle size of ~260kb instead of +1mb.

Screenshot 2025-03-27 at 17 19 13

### DIFF
--- a/docs/syntax/code.md
+++ b/docs/syntax/code.md
@@ -316,3 +316,8 @@ This `code` is inline.
 :::
 
 ::::
+
+## Supported Languages
+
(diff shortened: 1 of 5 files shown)


Mpdreamz (approved):
Lgtm!

We also need to update `SupportedLanguages.cs`

reakaleek:
> Sync SupportedLanguages.cs

https://github.com/elastic/docs-builder/pull/864/commits/881d80082c9d7b98b3f63c64bb8c60874cd414f0

PR #870, opened 2025-03-28
Pin dependencies and configure dependabot

reakaleek:
- **Update package.json to use fixed versions for devDependencies and dependencies**
- **Add dependabot configuration for weekly npm updates in Elastic.Markdown**


.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "src/Elastic.Markdown"
+    schedule:
+      interval: "weekly"
+    labels:
+      - chore
```
Mpdreamz, line 8: Should we use the company renovate instead?

https://github.com/elastic/docs-builder/pull/1

Mpdreamz (approved):
LGTM left a q? about renovate though

.github/dependabot.yml
Mpdreamz: Discussed offline, we are going with dependabot!

PR #875, opened 2025-03-28
Add support for Myst inline role syntax with applies_to as first usecase

Mpdreamz:
This PR adds support for Myst inline role syntax.

The first `role` we support is `{applies_to}` and `{preview}``


Mpdreamz:
Myst documents 

```markdown
{role}`arguments`
```

Sphinx documents

```markdown
:role:`arguments`
```

The latter will stand out more if we ever support `argument-less` roles e.g

```markdown
This is paragraph with an inline :role:
```

Although we might want to use `:` delimited keywords for emoticons/icons too.

jmikell821:
Hello! Adding a few examples so we can consider these scenarios when formalizing guidelines. In Security, we often have granular changes (such as a new setting or privilege) that are merged into serverless first, then typically get rolled into the stack release. 

Here's an example where the UI is slightly different between serverless and the stack. 




Also, in this [Endpoint response actions doc](https://docs-v3-preview.elastic.dev/elastic/docs-content/tree/main/solutions/security/endpoint-response-actions#response-action-commands), you can see notated differences between serverless and stack privileges.

PR #886, opened 2025-03-31
Provides syntax for subscript & superscript examples

natasha-moore-elastic:
Preview: https://docs-v3-preview.elastic.dev/elastic/docs-builder/pull/886/syntax/sundries#subscript-superscript

### DIFF
--- a/docs/syntax/sundries.md
+++ b/docs/syntax/sundries.md
@@ -15,13 +15,34 @@ Note that there should be no space between the enclosing markers and the text.
 | `\*escaped symbols\*`           | \*escaped symbols\* |
 
 ## Subscript & Superscript
+ 
+::::{tab-set}
 
-| Syntax           | Result        |
-| ---------------- | ------------- |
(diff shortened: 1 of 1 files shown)


reakaleek:
Thank you for the contribution @natasha-moore-elastic 

We recently started to use the pattern of using tabs to show the markdown and the output. As an example: https://docs-v3-preview.elastic.dev/elastic/docs-builder/tree/main/syntax/dropdowns

Do you think this would work in this case as well?

natasha-moore-elastic:
> Thank you for the contribution @natasha-moore-elastic
> 
> We recently started to use the pattern of using tabs to show the markdown and the output. As an example: https://docs-v3-preview.elastic.dev/elastic/docs-builder/tree/main/syntax/dropdowns
> 
> Do you think this would work in this case as well?

@reakaleek sure, I can use that pattern instead 👍

reakaleek (approved):

Issue #873, opened 2025-03-28
Release note pages 404 in staging


bmorelli25:
### Summary

In staging, some release note pages are 404ing. @colleenmcginnis pointed out that it appears to be all of the pages that live in docs-content.

### Steps to repro

1. Head to https://staging-website.elastic.co/docs/release-notes
2. Click on any of the following: [Elastic Observability](https://staging-website.elastic.co/docs/release-notes/observability), [Elastic Security](https://staging-website.elastic.co/docs/release-notes/security), [Fleet and Elastic Agent](https://staging-website.elastic.co/docs/release-notes/fleet)
3. See 404

bmorelli25:
@colleenmcginnis spotted the problem: there's no index.md file in these directories. For example: https://github.com/elastic/docs-content/tree/main/release-notes/elastic-observability

We should update our build to error in this situation.

bmorelli25:
Interestingly enough, Elastic Cloud Serverless also 404s, but it gets closest to working. Clicking on the release note page brings you to: https://staging-website.elastic.co/docs/release-notes/cloud-serverless/hangelog. Notice the missing "c" from the beginning of "hangelog"

cotti:
> Interestingly enough, Elastic Cloud Serverless also 404s, but it gets closest to working. Clicking on the release note page brings you to: https://staging-website.elastic.co/docs/release-notes/cloud-serverless/hangelog. Notice the missing "c" from the beginning of "hangelog"
> 
> [@cotti](https://github.com/cotti) could this be the same `TrimStart()` error seen/fixed in [#835](https://github.com/elastic/docs-builder/pull/835)?

It is! Just found some other weird URLs, such as "docs/release-notes/cloud-serverless/precations" (should be deprecations). I'll prep a fix for this now.

PR #888, opened 2025-03-31
Add extension method for relative path trimming when using ReadOnlySpan<char>

cotti:
Follow-up to #835 

As mentioned by @bmorelli25 on https://github.com/elastic/docs-builder/issues/873#issuecomment-2762175027, there was another place using `TrimStart(string)` that cut away portions of the resulting path. This PR introduces an extension method for `ReadOnlySpan<char>` and applies it in the appropriate places.


reakaleek (approved):

bmorelli25:
Thank you!

PR #894, opened 2025-04-01
Add content_source (current/next) to assembler

Mpdreamz:
This allows repositories to declare what content source to use for `next` and/or `current`.

Both default to `main`. 

When checking out an environment 

```
docs-assembler repo clone-all --environment prod
```

The environment configuration determines to use `next` or `current`. 

Updated the checkout folder to include `next` or `current`. 

The default output folder remains the same


Mpdreamz:
The failure here is because it uses an outdated assembler but its picking up the updated `src/docs-assembler/assembler.yml` file vs the embedded one.

https://github.com/elastic/docs-builder/actions/runs/14200698014/job/39786692416?pr=894

reakaleek (approved):

Issue #131, opened 2025-01-06
Add support for block comments


bmorelli25:
### Summary

You can comment out a single line by putting `%` at the beginning of the line:

```md
% This is a comment that is not rendered in the built HTML
This is not a comment and is rendered
```

You cannot easily comment out multiple lines. You can add a `%` to the beginning of every line, but I don't think that's intuitive. We should support HTML comments or some other easy multiline comment option:

```md
<!---
your comment goes here
and here
-->
```

bmorelli25:
Done in #940.

PR #940, opened 2025-04-01
Add support for multiple line comments

cotti:
Handles #131 

Multiple line comments can be added between `<!--` and `-->`.

```
Before
<!--
Multiple
line
comments
-->
After
```

Will render

> Before
> After


docs/syntax/comments.md
```diff
@@ -10,4 +10,17 @@ Use `%` to add single-line comments.
 % This is a comment
 ```
 
-Make sure to add a space after the `%`.
\ No newline at end of file
+Make sure to add a space after the `%`.
+
+## Multiple line comments
+
+Use `<!---` and `-->` to add multiple line comment blocks.
```
reakaleek, line 17: I think it could be confusing or error prone if the starting has 3 dashes end the end has 2.

tests/Elastic.Markdown.Tests/Inline/CommentTest.cs
```diff
@@ -27,3 +27,37 @@ public void GeneratesAttributesInHtml() =>
...
+	end of comments
+	-->
+
+	also not a comment
+	"""
+)
+{
+
+	[Fact]
+	public void GeneratesAttributesInHtml() =>
```
reakaleek, line 46: Could you please add a test case where the comment contains a link that would normally fail e.g. a non existing link.

reakaleek (commented):
Nice 🚀

docs/syntax/comments.md
bmorelli25: That's likely my fault. I accidentally had one extra `-` in the issue. Let's update to the standard [html comment](https://www.w3schools.com/tags/tag_comment.asp) syntax of `<!--` and `-->`.

Being able to hit cmd+/ in VSC to comment out multiple lines is going to be great!

tests/Elastic.Markdown.Tests/Inline/CommentTest.cs
cotti: I've added a couple of link attributes in the multi-line comment in a new test, is that it? Is the objective to confirm if they are ignored too?

reakaleek: Exactly, this was an issue in the past.

And the logic is quite simple and I'm not sure if it would have worked also here

https://github.com/elastic/docs-builder/blob/9818d12717fa118d30a8f5ca884201089a7469c4/src/Elastic.Markdown/Myst/InlineParsers/DiagnosticLinkInlineParser.cs#L100

docs/syntax/comments.md
```diff
@@ -10,4 +10,17 @@ Use `%` to add single-line comments.
 % This is a comment
 ```
```
bmorelli25, line 12: Can you remove L3 from this file

reakaleek (approved):

docs/syntax/comments.md
```diff
@@ -10,4 +10,17 @@ Use `%` to add single-line comments.
...
 
-Make sure to add a space after the `%`.
\ No newline at end of file
+Make sure to add a space after the `%`.
+
+## Multiple line comments
```
reakaleek, line 15: Nit: I think the common term is "Multiline comment"

PR #944, opened 2025-04-02
Fix same page anchor links

reakaleek:


Mpdreamz (approved):

Mpdreamz:
Return of the cursorful videos! :)

Issue #724, opened 2025-03-12
Include /reference/ path on local builds


LikeTheSalad:
I was checking the guide to contribute docs locally, and when I ran docs-builder serve I got no errors, but I couldn't see the docs by opening http://localhost:3000/ (I was getting 404 errors). Luckily after taking a look at some messages in some Slack groups I came across this other link: http://localhost:3000/reference/ which did the trick :+1: and I was thinking that we should update the guide to include the /reference/ path in [this part](https://elastic.github.io/docs-builder/contribute/locally/#step-three) to avoid confusions in the future.

An alternative solution mentioned [here](https://elastic.slack.com/archives/C0JF80CJZ/p1741789836632549?thread_ts=1741788319.792849&cid=C0JF80CJZ) might be to show the full link in the console.

reakaleek:
When https://github.com/elastic/docs-builder/pull/956 is released the CLI will redirect to the first page it finds if there is no root `index.md` file when running `docs-builder serve`.
In case of reference repos this means it will redirect to `https://localhost:3000/reference`

PR #956, opened 2025-04-03
Redirect to first markdownfile on serve command in case index.md doesn't exist

reakaleek:
Closes https://github.com/elastic/docs-builder/issues/724

In case of reference repositories, when you open `https://localhost:3000` this will redirect to `https://localhost:3000/reference` if there is no root `index.md` file.


Mpdreamz (approved):
👍 💯

PR #963, opened 2025-04-03
Add update-link-index-all command

reakaleek:


src/docs-assembler/Cli/RepositoryCommands.cs
```diff
@@ -117,4 +127,59 @@ public async Task<int> BuildAll(
...
+			{ NarrativeRepository.RepositoryName, assembleContext.Configuration.Narrative }
+		};
+		await Parallel.ForEachAsync(repositories,
+			new ParallelOptions
+			{
+				CancellationToken = ctx,
+				MaxDegreeOfParallelism = Environment.ProcessorCount
+			}, async (kv, c) =>
+			{
+				var name = kv.Key.Trim();
```
Mpdreamz, line 151: `try/catch` the loop method and emit an error over the diagnostics channel. 

That way we can still build all the repositories that do validate and update those.

Mpdreamz (changes requested):
A few nits, other than that LGTM!

Mpdreamz (approved):

Issue #1061, opened 2025-04-08
[Design] Left navigation changes


KOTungseth:
We received design guidance from our Design team on the elastic.co/docs left navigation. 

I've include the notes here: https://www.figma.com/board/oDOHRhyn96Pu68VxhL1iGu/Navigation-changes?node-id=0-1&p=f&t=9AI9A3YiisBuhhxB-0

Changes include:

- Make the `Get started`, `Solutions and use cases`, etc dropdown an overlay instead of the up/down push behavior (I'm sorry, Jan)
- Move section arrows to the right hand side and align as a column
- Remove the lines and spaces between the main sections
- Render the navigation text the same size, but use bold for first level titles and indentation to separate hierarchically

PR #1065, opened 2025-04-09
Left pages nav design changes

reakaleek:
Part of https://github.com/elastic/docs-builder/issues/1061

## Changes

Left pages nav:

- Remove borders
- Remove spacing
- Adjust indentation
- Same font size

## Result

https://github.com/user-attachments/assets/12dc4d64-240b-453d-b601-510e0eb52aea

## Follow-up
- Dropdown should open in an overlay.


Mpdreamz (approved):
Much calmer!

The arrow buttons all being right aligned also helps with mobile navigation to minimize thumb movement.

PR #1089, opened 2025-04-10
Reorder Extend section

georgewallace:
This puts the threee contribution pages first and then the 2 around creation. This makes the ToC flow smoother for the extend section.

@KOTungseth This reorders the ToC for the Extend section based on the feedback you provided in our Landing page and Nav UX.


Mpdreamz (commented):
LGTM but will wait for @KOTungseth's review before merging.

Mpdreamz (approved):

Issue #1074, opened 2025-04-10
[Feature Request]: Numbered steps


theletterf:
### Prerequisites

- [x] I have searched existing issues to ensure this feature hasn't already been requested
- [x] I have tested using the latest version of docs-builder

### What problem are you trying to solve?

A special directive that renders content in a progression of numbered steps. This is useful for visually indicating procedures, for example.

<img width="977" alt="Image" src="https://github.com/user-attachments/assets/38eacc45-a8d7-40e2-950f-b343d05d7a64" />

### Proposed Solution

Something like this. The directive would take all headings of the same level within the block and turn them into steps, the content within converted into the content of the step itself (nesting allowed).

```
::::: steps

## Heading that turns into step 1

Content

## Heading that turns into step 2

:::::

### Examples and Research

Mintlify and other SSGs have this, but it's a custom development in most cases. I developed it for Tinybird docs. Example: https://www.tinybird.co/docs/forward

### Alternative Solutions

_No response_

### Additional Context

_No response_

### How important is this feature to you?

Nice to have

theletterf:
+CC @bmorelli25 @alaudazzi

reakaleek:
Related discussion: https://github.com/elastic/docs-builder/discussions/754

PR #1092, opened 2025-04-10
Add stepper directive

reakaleek:
Closes https://github.com/elastic/docs-builder/issues/1074

## Preview

https://docs-v3-preview.elastic.dev/elastic/docs-builder/pull/1092/syntax/stepper

## Result
https://github.com/user-attachments/assets/8a4eab18-744c-43f1-9cd8-d0a088b3cf9d

### DIFF
--- a/docs/syntax/stepper.md
+++ b/docs/syntax/stepper.md
@@ -1,5 +1,8 @@
 # Stepper
 
+This section introduces the concept of a stepper, a visual representation of sequential steps, 
+often used in tutorials or guides to break down processes into manageable stages.
+
 ## Numbered Stepper
 
 ::::::{tab-set}
--- a/docs/syntax/stepper.md
+++ b/docs/syntax/stepper.md
@@ -1,0 +1,121 @@
+# Stepper
+
+## Numbered Stepper
+
+::::::{tab-set}
+
+:::::{tab-item} Output
+
+::::{stepper}
+
+1. ### Create an index [getting-started-index-creation]
+
+   Create a new index named `books`:
+
+   ```console
+   PUT /books
+   ```
+
+   The following response indicates the index was created successfully.
+
+   :::{dropdown} Example response
+   ```console-result
+   {
+     "acknowledged": true,
+     "shards_acknowledged": true,
+     "index": "books"
+   }
+   ```
+   :::
+
+
+2. ### Add data to your index
+
+   :::{tip}
+   This tutorial uses Elasticsearch APIs, but there are many other ways to [add data to Elasticsearch](#).
+   :::
+
+   You add data to Elasticsearch as JSON objects called documents. Elasticsearch stores these documents in searchable indices.
+
+3. ### Define mappings and data types
+
+   When using dynamic mapping, Elasticsearch automatically creates mappings for new fields by default.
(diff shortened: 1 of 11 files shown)


docs/syntax/stepper.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Numbered Stepper
+
+::::::{tab-set}
+
+:::::{tab-item} Output
+
+::::{stepper}
+
+1. ### Create an index [getting-started-index-creation]
```
Mpdreamz, line 14: Visually I LOVE this! 

As for syntax, can we make it work on `unordered` lists too? I always hate renumbering ordered lists :)

The original requests had it working without creating lists which I really like too:


````markdown
::::: steps

## Heading that turns into step 1

Content

## Heading that turns into step 2

:::::
```

e.g turn any header into a list item so that we normalize the header style in the stepper no matter what level people use. This also has the added benfit we force people into a uniform `stepper` style with headers and step text.

Mpdreamz (changes requested):

docs/syntax/stepper.md
reakaleek: Yeah. I was also thinking about this.

I event think we should go one step more explicit.

like

```markdown
:::stepper

:::step
## Heading

content
:::

:::step

## Heading

content
:::
```

This will give us more flexibility in the future, e.g. using icons.

```markdown
:::stepper

:::step
:icon: rocket
## Heading

content
:::

:::step
:icon: checkmark
## Heading

content
:::
```

WDYT?

reakaleek: Example: 
image

Mpdreamz: Ohh we're cooking now 🧑‍🍳 

How about making the heading arguments to the step taking away any ambiguity of the header sizing?

```markdown
:::{stepper}

:::{step} Heading
:icon: rocket


content
:::

:::{step} Heading
:icon: checkmark

content
:::
::::
```

reakaleek: I like it. It would be hard then to integrate it into the "On this page" nav. But maybe we don't want that anyway.

reakaleek: Not sure.

Mpdreamz: I do think we don't want steps as on this page listings.

Mpdreamz: e.g if steps become large paragraphs it might not be ripe for the `{stepper}` directive to begin with. 

It's likely content not made for a quick steps to follow.

reakaleek: But I guess it still makes sense create anchors for it.. to be able to easily share a step.

Mpdreamz: ++ agreed!

Writers can even control it using

`:name: <anchor>` 

Like most directives.

reakaleek: Changed syntax: https://github.com/elastic/docs-builder/pull/1092/commits/12374249e34b26d9d54d99ac5d0f0ecf8740c20f

Mpdreamz (approved):
I freaking love this addition! LGTM

PR #1093, opened 2025-04-11
BREAKING CHANGE: Don't allow dots (.) in slug

reakaleek:
## Context

Currently, dots (`.`) are allowed in slugs. This means generated anchors also contain slugs.

If you create a header like

```markdown
## 1. Install
```

Then it's anchor will be `#1.-install`.

However, IDE's are expecting the anchor `#1-install`.

image

Reported by @flobernd 

## Changes

By removing the `.` from the allowed chars of Slugify, the correct anchor is generated.

### DIFF
--- a/src/Elastic.Markdown/Myst/SectionedHeadingRenderer.cs
+++ b/src/Elastic.Markdown/Myst/SectionedHeadingRenderer.cs
@@ -34,7 +34,7 @@ public class SectionedHeadingRenderer : HtmlObjectRenderer<HeadingBlock>
 
 		var slug = anchor ?? header.Slugify();
 		if (slug.Contains('$'))
-			slug = HeadingAnchorParser.InlineAnchors().Replace(slug, "").Slugify();
+			slug = HeadingAnchorParser.InlineAnchors().Replace(slug, "");
 
 		_ = renderer.Write(@"<div class=""heading-wrapper"" id=""")
 			.Write(slug)
--- a/src/Elastic.Markdown/Myst/SectionedHeadingRenderer.cs
+++ b/src/Elastic.Markdown/Myst/SectionedHeadingRenderer.cs
@@ -32,11 +32,9 @@ public class SectionedHeadingRenderer : HtmlObjectRenderer<HeadingBlock>
 		var header = obj.GetData("header") as string;
 		var anchor = obj.GetData("anchor") as string;
 
-		var slugTarget = (anchor ?? header) ?? string.Empty;
-		if (slugTarget.Contains('$'))
-			slugTarget = HeadingAnchorParser.InlineAnchors().Replace(slugTarget, "");
-
-		var slug = slugTarget.Slugify();
+		var slug = anchor ?? header.Slugify();
+		if (slug.Contains('$'))
+			slug = HeadingAnchorParser.InlineAnchors().Replace(slug, "").Slugify();
 
 		_ = renderer.Write(@"<div class=""heading-wrapper"" id=""")
 			.Write(slug)
--- a/src/Elastic.Markdown/Myst/SectionedHeadingRenderer.cs
+++ b/src/Elastic.Markdown/Myst/SectionedHeadingRenderer.cs
@@ -32,9 +32,13 @@ public class SectionedHeadingRenderer : HtmlObjectRenderer<HeadingBlock>
 		var header = obj.GetData("header") as string;
 		var anchor = obj.GetData("anchor") as string;
 
-		var slug = anchor ?? header.Slugify();
-		if (slug.Contains('$'))
-			slug = HeadingAnchorParser.InlineAnchors().Replace(slug, "");
+		var slugTarget = (anchor ?? header) ?? string.Empty;
+		if (slugTarget.Contains('$'))
+			slugTarget = HeadingAnchorParser.InlineAnchors().Replace(slugTarget, "");
+
+		var slug = slugTarget.Slugify();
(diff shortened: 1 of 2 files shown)


src/Elastic.Markdown/Myst/SectionedHeadingRenderer.cs
```diff
@@ -32,11 +32,9 @@ protected override void Write(HtmlRenderer renderer, HeadingBlock obj)
...
 		var anchor = obj.GetData("anchor") as string;
 
-		var slugTarget = (anchor ?? header) ?? string.Empty;
-		if (slugTarget.Contains('$'))
-			slugTarget = HeadingAnchorParser.InlineAnchors().Replace(slugTarget, "");
-
-		var slug = slugTarget.Slugify();
+		var slug = anchor ?? header.Slugify();
+		if (slug.Contains('$'))
+			slug = HeadingAnchorParser.InlineAnchors().Replace(slug, "");
```
Mpdreamz, line 37: I do think we want to always sanitize `anchor` or `header` through `Slugify()`. 

What was the usecase for removing it from `anchor`?

Mpdreamz (changes requested):
+1 on removing dots.

src/Elastic.Markdown/Myst/SectionedHeadingRenderer.cs
reakaleek: Sorry, currently working on this. Switched back to draft.

But I was thinking of limiting the blast radius.

If someone creates a custom anchor, they might want to maintain the dot.

Mpdreamz: I think having uniformity in our slugging beats allowing dots in custom anchors. 

The main reason might be version numbers but they look okay without dots too e.g

`#whats-new-in-8.18`

vs

`#whats-new-in-8-18`

I actually prefer the latter visually too.

reakaleek: The slight problem here is that it seems like the default markdown behaviour is to strip the `.` and not replace it with `-`.

Screenshot 2025-04-11 at 11 52 55

Mpdreamz: ahh but that discrepancy exists today already no? 


If someone writes 

```
## Whats new in 8.18
```

We slug to: `whats-new-in-8-18` today? 

```
## Whats new in 8.18 [whats-new-in-8.18]
```

I do think that should also be sanitized the same way to `whats-new-in-8-18`. 

We shouldn't take anchors verbatim and having two sanitation rules is a bit suprising to writers IMO.

Mpdreamz: I can also be easily convinced to update our default slugify to match most IDE editor experience to `#whats-new-in-818` being our sanitized output too.

reakaleek: Currently, the Slugify allows dots.

Hence, the auto generated slug would become `whats-new-in-8.18`.

IMO we for headings we should use the default behaviour, so that authors can utilize the suggested anchor from IDEs.

For custom anchors, I think it's okay to replace `.` with `-`.
But it will break custom anchors with `.` and I know they exist already.

Mpdreamz: > IMO we for headings we should use the default behaviour, so that authors can utilize the suggested anchor from IDEs.

+1 to this. 'default' is maybe a misnomer since there is no default/standard in the spec but this will greatly improve writing experience in IDE's. 

> For custom anchors, I think it's okay to replace . with -.

Not a hill I'm not willing to budge from but our code will become significantly easier if we can just blindly slugify something without having to check its origin.


> But it will break custom anchors with . and I know they exist already.

Aye, worth breaking now though IMO! Will be much harder next week.

flobernd: +1 for uniform handling. I found at least another place that behaves completely different 😅 

1885-1020-max

flobernd: - The numbering gets removed (maybe this is intended?)
- As a consequence, the anchor is generated differently which breaks the navigation functionality (clicking on the navbar link does nothing)

reakaleek: 💩

reakaleek: One problem I just thought of is the authoring experience.

It will be confusing if you set a custom anchor and then it's sanitized.

E.g.

If I add the custom anchor `#whats-new-in-8.0` I would expect it to stay like that.

It could be confusing to see that it changed to `#whats-new-in-8-0`.

Maybe the best way is to actually to discourage dots (`.`) in anchors and make it error?

Mpdreamz: We need to sanitize custom anchors anyhoo e.g `#What's-new-in-8.0` we can not take it verbatim. The sanitization is not surprising to me but thats subjective :)

We could `hint` if custom anchors end up different after sanitization?

reakaleek: @flobernd https://github.com/elastic/docs-builder/pull/1094 should fix the bug you reported.

PR #1113, opened 2025-04-14
Fix parent breadcrumbs in assembler build

Mpdreamz:


src/Elastic.Markdown/Slices/Layout/_Breadcrumbs.cshtml
```diff
@@ -1,5 +1,16 @@
 @using Elastic.Markdown.Helpers
 @inherits RazorSlice<LayoutViewModel>
+@{
+	var targets =
+		!Model.Features.IsPrimaryNavEnabled
+			? Model.Parents.Reverse().Skip(1)
+			: Model.Parents.Reverse();
```
reakaleek, line 7: Super nit:

```suggestion
	var targets =
		Model.Features.IsPrimaryNavEnabled
			? Model.Parents.Reverse()
			: Model.Parents.Reverse().Skip(1);
```

IMO it's easier to read as "If primary nav is enabled... then"

But this is also very subjectiv

reakaleek:
Nice 🎸

PR #1141, opened 2025-04-17
Fix --version and new version update announcements

Mpdreamz:


reakaleek:
Will this also work for `docs-assembler`?

PR #1143, opened 2025-04-17
Retry getting repository links

reakaleek:
Related https://github.com/elastic/docs-internal-workflows/issues/41

### DIFF
--- a/src/Elastic.Markdown/Links/CrossLinks/CrossLinkFetcher.cs
+++ b/src/Elastic.Markdown/Links/CrossLinks/CrossLinkFetcher.cs
@@ -56,9 +56,7 @@ public abstract class CrossLinkFetcher(ILoggerFactory logger) : IDisposable
 
 	protected async Task<LinkIndexEntry> GetLinkIndexEntry(string repository, Cancel ctx)
 	{
-		var linkIndex = await FetchLinkIndex(ctx);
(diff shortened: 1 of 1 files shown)


bmorelli25 (commented):
Great idea 👍

reakaleek:
Unsure if this is needed or even fixes the problem.

Let's wait for the new link-index-updater function and see if it still happens.

src/Elastic.Markdown/Links/CrossLinks/CrossLinkFetcher.cs
```diff
@@ -87,6 +82,31 @@ protected async Task<LinkReference> Fetch(string repository, string[] keys, Canc
 		throw new Exception($"Repository found in link index however none of: '{string.Join(", ", keys)}' branches found");
 	}
 
+	// See https://github.com/elastic/docs-internal-workflows/issues/41
+	private async Task<Dictionary<string, LinkIndexEntry>> GetRepositoryLinksWithRetry(string repository, Cancel ctx)
+	{
+		const int maxAttempts = 3;
+		var attempts = 0;
+		while (true)
```
Mpdreamz, line 90: Use the same attempts >= guard as when we throw?


While true loops should be reserved for long running operations

Mpdreamz (changes requested):

reakaleek:
I'm closing this. The error did not occur anymore since I refactored the link-index-updater lambda.

PR #1148, opened 2025-04-17
Improve syntax quick ref

marciw:
* Add lables to make DOs and DON'Ts clearer and more accessibility
* Add link from syntax index to quick ref


marciw:
@elastic/docs-engineering no rush, but thanks in advance for reviewing when you get a chance

bmorelli25 (approved):
✨

PR #1162, opened 2025-04-21
Update syntax image examples to use `images` path

kilfoyle:
As a convention our images are stored in directories named `images`, but our syntax instructions use `img` in the example image paths. This caused confusion for one developer so far, so I think it's worth updating our guide.


colleenmcginnis (commented):
@kilfoyle I think you'll also need to rename the [`docs/syntax/img`](https://github.com/elastic/docs-builder/tree/kilfoyle-patch-1/docs/syntax/img) directory to `docs/syntax/images` and then update any other references to images outside this Markdown file to also point to `images` instead of `img`.

Mpdreamz (approved):
LGTM thanks @kilfoyle !

PR #1163, opened 2025-04-21
Use substitution values in image alt and title text

### LINKED ISSUE #814: Unused substitution key hints displayed if key only used in image alt text
### Describe the bug

After #779 was fixed, I'm still seeing some unused substitution key hints when a key is only used in image alt text. 

### Expected behavior

I should only get hints when a substitution key is actually not used.

### Steps to reproduce

Examples:

* docs-content: `reports-app` ([hint](https://github.com/elastic/docs-content/actions/runs/13992987645/job/39180983455#step:12:591) / [key usage](https://github.com/elastic/docs-content/blob/main/explore-analyze/find-and-organize/reports.md?plain=1#L18))
  * Note: The hint about `release-date` being unused is valid! 
* kibana: `rules-ui` ([hint](https://github.com/elastic/kibana/actions/runs/13993027947/job/39181114208?pr=215260#step:12:38) / [key usage](https://github.com/elastic/kibana/blob/main/docs/reference/connectors-kibana/cases-action-type.md?plain=1#L20))

### Tooling

- [x] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

colleenmcginnis:
Fixes #814

This attempts to use substitution values in image `alt` and `title` text. I've noticed some keys being flagged as "unused" when they are only used in `alt` or `title` text ([example](https://github.com/elastic/docs-builder/issues/814)).  I've also been building the site locally and using the HTML files in the `.artifacts/assembly/` directory to look for content problems (including unresolved sub keys) and there were sooo many in `alt` and `title` text that it was a bit difficult to track content fixes.

(I hope it's ok that I picked up this issue. During spacetime I'm trying to get more familiar with how docs-builder works so I can make tiny fixes like this one as needed in the future. 🪐 )


src/Elastic.Markdown/Myst/Directives/ImageBlock.cs
```diff
@@ -64,9 +70,10 @@ public class ImageBlock(DirectiveBlockParser parser, ParserContext context)
 	public override void FinalizeAndValidate(ParserContext context)
 	{
 		Label = Prop("label", "name");
-		Alt = Prop("alt");
-		Align = Prop("align");
+		Alt = (Prop("alt") ?? "{undefined}").ReplaceSubstitutions(context);
```
reakaleek, line 73: Could you please explain what this is doing?

Will this leave a `{undefined}` in the alt property?

Generally, there is valid use case for an empty `alt=""` property. See https://www.w3.org/WAI/tutorials/images/decorative/

Not sure if this still given with this condition.

colleenmcginnis: I ~~stole~~ borrowed from [`TabSetBlock.cs`](https://github.com/elastic/docs-builder/blob/main/src/Elastic.Markdown/Myst/Directives/TabSetBlock.cs#L52) 🙃  but if an empty string is better, let's do that!

reakaleek: Ah interesting. I also don't understand what it's doing in TabSet :D

Mpdreamz: Uhh I can't think of any. Must have been a leftover from frenzy coding 😸

reakaleek (approved):

reakaleek (approved):
LGTM.

Only nit: Make it possible to set an empty `alt` tag.

src/Elastic.Markdown/Myst/Directives/ImageBlock.cs
colleenmcginnis: I'm having some trouble figuring out the right syntax. Just changing `"{undefined}"` to `""` is the only thing I could get to work, but I don't think it makes sense to replace subs in an empty string.... Any hints? 😏

reakaleek: Maybe something like:

```
Prop("alt")?.ReplaceSubstitutions(context) ?? "";
```

colleenmcginnis: Ahhh the `?` after `Prop("alt")` was the missing piece. Thanks!

src/Elastic.Markdown/Myst/InlineParsers/DiagnosticLinkInlineParser.cs
```diff
@@ -63,38 +63,41 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
...
+				height = width;
+			else if (!height.EndsWith('%'))
+				height += "px";
+
+			attributes.AddProperty("width", width);
+			attributes.AddProperty("height", height);
+
+			title = title[..matches.Index];
+		}
+		link.Title = (title ?? "{undefined}").ReplaceSubstitutions(context);
```
reakaleek, line 100:
```suggestion
		link.Title = title?.ReplaceSubstitutions(context);
```

Mpdreamz (approved):

reakaleek:
> @reakaleek thanks for picking this up. 🙂

No, thank **you** for the contribution!

PR #1176, opened 2025-04-28
Refactor domain model out of Elastic.Markdown.csproj

Mpdreamz:
Only a first step to isolate components in smaller dedicated libraries. 

This allows us to e.g ship smaller AOT compiled lamdba binaries and should make it easier to create dedicated tooling in the future without always having to take a dependency on everything in `Elastic.Markdown`. That project really should only hold our authoring components.

Also a first step towards a more layered cleaner architecture as we shift out of our first release towards GA.

Again this is not complete by any means some pieces require a bit more work to untangle and create cleaner object dependency graphs.


reakaleek (approved):
😮 🥇

PR #1180, opened 2025-04-29
Publish ECS 9.0.0 documentation

mjwolf:
Update assembler config to publish ECS v9.0.0 documentation as the current version.


mjwolf:
I would like this to publish the `v9.0.0` tag in the ecs repo. I'm not completely sure if I've done it correctly, or if publishing tags are supported. If tags can't be published, I can change this to publish the `9.0` branch too.

I also don't have permission to add a label to this PR, to make the "required-labels / check-labels" check happy here.

reakaleek:
Hi @mjwolf,

Thank you for your contribution! To be honest, we don't have a proper current/next strategy yet.
There could be major changes. The changes look fine to me, but I'm not entirely sure about the consequences.

Let's wait for @Mpdreamz or @bmorelli25's opinion.

Mpdreamz:
The change looks good to me and config wise I don't expect this to change. 

Our current/next strategy is defined in the sense we provide the following options:

- current & next are equal
- current & next follow a branching model. 
  - here we 'specialize' by being on the DRA and have a variable for all repositories that branch when releases are cut.
  - Other products that have their own versioning scheme or might sometimes lag behind company releases (like ECS). 


What's not defined is what repositories will use what strategy. 

We have a techical blocker in that we are not currently setup to publish from `current` branches. 

We currently require `links.json` to be build on the main integration branches: https://github.com/elastic/docs-builder/blob/main/.github/workflows/preview-build.yml#L165-L172

Our workflow is currently only triggering on `main` and `master`. We need a Github App to be able to run more dynamically on folks different `current` branches.

Mpdreamz:
This is now correctly failing because `ecs` has not yet build their docs on their `9.0` branch.

image


https://github.com/elastic/ecs/pull/2476 should now be ready to be merged upstream (and backported to `9.0`).

Mpdreamz (approved):
This LGTM!

Did a full assemble production build locally and hit no errors.

PR #1183, opened 2025-04-29
Max width layout

reakaleek:
- **Full width layout PoC**
- **Also set max width for other elements**
- **Fixes and fine-tuning**
- **Cleanup**
- **Fix isolated build header position**
- **Formatting**
- **Prettier**
- **Limited max width PoC**


Mpdreamz (approved):
Approving to unblock.

PR #1197, opened 2025-04-29
Dynamic max-width based on 90ch width for the content

Mpdreamz:
https://github.com/user-attachments/assets/39c63550-f546-4d7f-845a-c642451ac3ec

Instead of having variable gaps this now refocusses on having a dynamic `max-layout-width` based on ensuring theres enough space for the side bars + 90characters of content. 

This normalizes margins and creates calmer reading experience (IMO of course :))

The header menu and footer also follow `max-layout-width` now ensuring they match the content boundaries. 

This also opens up the possibility to create a dedicated ultra large media experience if we ever want to go beyond `90ch` or create a bit more margin for wide screens.


reakaleek:
Making the markdown content and the "On this page" nav stick together follows an existing UI pattern.

And I think it makes sense to visually group the main content and the "on this page" table of contents. Both are showing the content of the current page.

image

image

image

reakaleek (approved):

PR #1198, opened 2025-04-29
Slight on this page margin adjustments

Mpdreamz:
image

Separate PR to #1197 since this might be controversial 😸 

* Ensure the page indicator aligns hard left (no more padding). This makes the marging in between the two vertical borders and the content exactly the same. 

* Move edit / submit an issue back up again since this often gets lost and visually anchors them in the same place for each page.


reakaleek:
https://github.com/elastic/docs-builder/pull/1197#issuecomment-2840120790

PR #1208, opened 2025-05-02
Update navigation.md with cross_links docs

theletterf:
This adds documentation on the cross_links directive in docset.yml files.


Mpdreamz (approved):
ty!

PR #1221, opened 2025-05-07
Update definition-lists.md

lcawl:
When I try to create a definition list with less than three spaces, it doesn't render locally.

### Preview

https://docs-v3-preview.elastic.dev/elastic/docs-builder/pull/1221/syntax/definition-lists

### DIFF
--- a/docs/syntax/definition-lists.md
+++ b/docs/syntax/definition-lists.md
@@ -1,4 +1,4 @@
-# Definition Lists
+# Definition lists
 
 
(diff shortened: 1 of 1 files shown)


docs/syntax/definition-lists.md
```diff
@@ -1,7 +1,7 @@
 # Definition Lists
```
lcawl, line 1:
```suggestion
# Definition lists
```

reakaleek (approved):

PR #1232, opened 2025-05-09
Add install scripts

theletterf:
This PR adds install scripts for docs-builder (*nix and Windows PowerShell). The scripts:

- Download the latest release.
- Unzip the binary in a tmp folder.
- Grant the binary execution permissions.
- Move the binary to a system/custom folder.
- Add the folder to path (Windows only).

The scripts overwrite the binary if one is found.

The PR also adds the scripts as release artifacts that can be invoked with a one-liner (documented).


.github/workflows/release.yml
```diff
@@ -152,13 +152,32 @@ jobs:
             .artifacts/publish/docs-builder/release/*.zip
             .artifacts/publish/docs-assembler/release/*.zip
 
+      # Add install scripts to artifacts directory (only on Linux runner to avoid duplication)
```
reakaleek, line 155: What's the reason for copying it to the `.artifacts/publish/scripts/` folder?

reakaleek (approved):
LGTM. Thank you for the contribution.

Just some questions.

install.sh
```diff
@@ -0,0 +1,29 @@
...
+  read choice
+  case "$choice" in
+    y|Y ) echo "Updating docs-builder..." ;;
+    n|N ) echo "Installation aborted."; exit 0 ;;
+    * ) echo "Invalid choice. Installation aborted."; exit 1 ;;
+  esac
+fi
+
+# Download the latest macOS binary from releases
+curl -LO https://github.com/elastic/docs-builder/releases/latest/download/docs-builder-mac-arm64.zip
```
reakaleek, line 17: Should we add something more dynamic?

like identifying the OS and arch?

theletterf: That would be ideal, especially for Linux. Let me give that a try based on cat /etc/os-release

.github/workflows/release.yml
reakaleek: Does it need to be copied anywhere?

AFAICS, you are uploading the script directly to the release.

Not sure if it's even necessary to upload it to the release.

The most common pattern I see is just sharing the raw github link.

e.g. https://raw.githubusercontent.com/elastic/docs-builder/43a5981b93cfaab58cf497ab477d14284f5c1227/install.sh

theletterf: Good point! I'll remove this release logic

reakaleek (approved):
❤️

PR #1236, opened 2025-05-09
Remove tab from install docs

theletterf:
macOS and Linux instructions are the same now. Also added links to the zip files for manual install.


Mpdreamz (approved):
LGTM!

Should we also create shortened links for the scripts? That way the one liner fits in the `90ch` we have available :)

theletterf:
@Mpdreamz Yes please! Could you? Currently in my car. :D

PR #1238, opened 2025-05-09
Add `id` attribute to `products` list frontmatter

reakaleek:
Add `id` attribute

so instead of

```md
---
products:
  - apm
---
```

use

```md
---
products:
  - id: apm
---
```

This makes the model more flexible and gives us the ability to add properties in the future if requirements are added. Just in case.

### DIFF
--- a/docs/_docset.yml
+++ b/docs/_docset.yml
@@ -13,9 +13,6 @@ subs:
   serverless-short: Serverless
   ece:   "Elastic Cloud Enterprise"
   eck:   "Elastic Cloud on Kubernetes"
-  
-products:
-  - id: elasticsearch
 
 features:
   primary-nav: false
--- a/docs/_docset.yml
+++ b/docs/_docset.yml
@@ -15,7 +15,7 @@ subs:
   eck:   "Elastic Cloud on Kubernetes"
   
 products:
-  - elasticsearch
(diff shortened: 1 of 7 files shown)


reakaleek:
Depends on the outcome of https://github.com/elastic/docs-content/pull/1336#issuecomment-2866161343

Mpdreamz (changes requested):
Can we support reading both `strings` and the `map`? 

That way we can minimize the syntax until the need arrises to expand product with more properties.

I don't really see any use case now, do you have any in mind?

reakaleek:
> Can we support reading both `strings` and the `map`?
> 
> That way we can minimize the syntax until the need arrises to expand product with more properties.
> 
> I don't really see any use case now, do you have any in mind?

I also don't see an immediate reason for extending it.

Generally, I made it a habit to use properties like that in YAML because it makes it so much easier to add properties in the future. Which follows the Open/Closed principle.

Also, I think it semantically makes it clearer for the user that you need to enter an ID in the list. Which removes some cognitive load.

Mpdreamz (approved):
Cool cool, I also came at it more from an authoring point of view where adding a list of keywords is easier then arrays of objects (bcuz yaml 😸)

Mpdreamz (approved):

PR #1240, opened 2025-05-09
Add random query string

reakaleek:
## Context

@florent-leborgne reported that when clicking the "View previous versions" in the landing page, he would be redirected back to elastic.co/docs.

We identified that it's caused by a local cached 301 redirect response.

This might affect public users as well.

## Changes

To avoid a locally cached 301 response redirecting to back to elastic.co/docs I used a random query string `?lang=en` so it does not look totally random like `?xyz`.


Mpdreamz (approved):
neat!

PR #1241, opened 2025-05-09
Ensure docs-builder builds to `.artifacts/assembly/<content-source>` and support a flag on serve-static to choose

Mpdreamz:


Mpdreamz:
Yes will add docs in a follow up PR!

I am actually eager to discuss cleaning up our CLI commands during EAH too :)

This was building inside `.artifacts/bin/debug/` inside `.artifacts/assembly` and thus `serve-static` was serving stale data.

PR #1244, opened 2025-05-12
Add shortened links to install commands

theletterf:
This adds shortened ela.st script to install commands. It also makes the first curl command silent.

### DIFF
--- a/docs/contribute/locally.md
+++ b/docs/contribute/locally.md
@@ -35,7 +35,7 @@ This guide uses the first option. If you'd like to clone the repository and buil
    Run this command to download and install the latest version of `docs-builder`:
 
    ```sh
-   sudo curl -L https://raw.githubusercontent.com/elastic/docs-builder/refs/heads/main/install.sh | sh
+   sudo curl -sL https://ela.st/docs-builder-install | sh
    ```
    This downloads the latest binary, makes it executable, and installs it to your user PATH.
 
@@ -58,7 +58,7 @@ To download and install the binary file manually, refer to [Releases](https://gi
    Run this command to download and install the latest version of `docs-builder`:
 
    ```powershell
-   iex (New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/elastic/docs-builder/refs/heads/main/install.ps1')
+   iex (New-Object System.Net.WebClient).DownloadString('https://ela.st/docs-builder-install-win')
    ```
 
    This downloads the latest binary, makes it executable, and installs it to your user PATH.
--- a/docs/contribute/locally.md
+++ b/docs/contribute/locally.md
@@ -35,7 +35,7 @@ This guide uses the first option. If you'd like to clone the repository and buil
    Run this command to download and install the latest version of `docs-builder`:
 
    ```sh
-   curl -sL https://ela.st/docs-builder-install | sudo sh
+   curl -sL https://ela.st/docs-builder-install | sh
    ```
    This downloads the latest binary, makes it executable, and installs it to your user PATH.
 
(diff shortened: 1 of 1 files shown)


docs/contribute/locally.md
```diff
@@ -35,7 +35,7 @@ This guide uses the first option. If you'd like to clone the repository and buil
    Run this command to download and install the latest version of `docs-builder`:
 
    ```sh
-   sudo curl -L https://raw.githubusercontent.com/elastic/docs-builder/refs/heads/main/install.sh | sh
+   sudo curl -sL https://ela.st/docs-builder-install | sh
```
jamietanna, line 38: Aside, this is the wrong way round, you want the `sudo` to be for executing it, not downloading it:

```suggestion
   curl -sL https://ela.st/docs-builder-install | sudo sh
```

jamietanna, line 38: :stop_sign: If I'm honest, I'm rather concerned about this being a recommendation:

- generally as a `curl | sh`
- then also throwing in `sudo` here
- the fact that this is a short URL that could be modified to point to any random location
- not teaching Elasticians to audit the scripts they're executing with superuser privileges

Looking up this pattern generally results in many articles such as https://lukespademan.com/blog/the-dangers-of-curlbash/ and https://security.stackexchange.com/questions/213401/is-curl-something-sudo-bash-a-reasonably-safe-installation-method and https://macarthur.me/posts/curl-to-bash/

docs/contribute/locally.md
```diff
@@ -35,7 +35,7 @@ This guide uses the first option. If you'd like to clone the repository and buil
    Run this command to download and install the latest version of `docs-builder`:
 
    ```sh
-   sudo curl -L https://raw.githubusercontent.com/elastic/docs-builder/refs/heads/main/install.sh | sh
+   curl -sL https://ela.st/docs-builder-install | sudo sh
```
jamietanna, line 38: Just to flag [my comment](https://github.com/elastic/docs-builder/pull/1244#discussion_r2084155287) (that was collapsed due to referring to the old line) still stands

docs/contribute/locally.md
theletterf: I understand the concern (and thanks for the suggestion, I've already applied it). It's a popular critique. However:

- We need a convenient, fast way of installing and updating docs-builder until we've packages set up.
- We control the [ela.st](http://ela.st/) links. Should Elasticians not trust Elastic links? Will the same happens with packages?
- We're not asking folks to run these instructions for CI/CD.

If the concern is sudo, I think we could work around that by adding the file to a PATH that is not part of the system (/.local/bin?).

docs/contribute/locally.md
theletterf: Would using a different, user-level path, thus removing the need for sudo help here?

theletterf: That and not using URL shorteners?

docs/contribute/locally.md
theletterf: Another option would be to distribute the install scripts with the release ZIPs and let users use them if they want.

docs/contribute/locally.md
theletterf: I've removed sudo from the command, as the user will be prompted anyway when required. Let's continue the conversation in the other comment.

docs/contribute/locally.md
theletterf: I've removed `sudo` from the command, as the user will be prompted interactively anyway.

reakaleek:
Can you add us to the ela.st links as maintainers?

Thank you!

🚀

theletterf:
Sure! Doing that now

theletterf:
@reakaleek @Mpdreamz You're now co-owners of the ela.st links!

PR #1245, opened 2025-05-12
Throw warning when applies_to is empty

### LINKED ISSUE #609: empty applies tag strangeness
### Describe the bug

if I just put this in the front matter:

```
applies_to:
```

I get this:

Image

IMO we rarely need this, because we only include `stack` labels alongside deployment type labels in limited circumstances, and usually with version info. 

if we _did_ want it, we wouldn't want the little empty blob representing a generic product label.

### Expected behavior

CI error on empty `applies_to` tag

### Steps to reproduce

front matter of 
```
---
applies_to:
---
```

### Tooling

- [x] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

theletterf:
Fixes #609 (and updates the docs) by returning a warning and null for the `ApplicableTo` record if the `applies_to` frontmatter value is null or has spaces. This leaves the `all` behavior as is.


theletterf:
@reakaleek Build's green!

Mpdreamz (approved):
LGTM! 

Small semantic change but any breakages will be born out of unintentionally defaulting to `all`

PR #1262, opened 2025-05-16
Test for irregular whitespaces

theletterf:
Irregular whitespaces, which are often inserted by mistake when using code editors, can break Markdown formatting. A way of mitigating this in VS Code is to use the [Fix Irregular Whitespace](https://marketplace.visualstudio.com/items?itemName=karlito40.fix-irregular-whitespace) extension, which detects and immediately edits these invisible characters.

This PR adds some irregular whitespace checking logic to the Markdown parser.


theletterf:
@Mpdreamz It seems we've tons of irregular whitespaces in docs-content... smoke tests are failing. I could confirm this by running a search from VSCode. https://github.com/elastic/docs-builder/actions/runs/15064731599/job/42346927067?pr=1262

Whether all those instances are problematic is to be seen. We could perhaps switch to a hint instead of a warning.

Mpdreamz (changes requested):
Great addition @theletterf !

We should make this part of the markdown parser since it more efficiently reads the files and exposes the contents as `Span<T>`.

That way we don't do double reads. Happy to ammend this PR to move these checks over there.

Making it a HINT first seems like a good idea while we go over all the repositories to fix it.

theletterf:
Thanks! Happy if you'd do both changes, so I can learn 🙇🏻

Mpdreamz (approved):
LGTM!

PR #1273, opened 2025-05-19
Complete locally.md docs and add macros

theletterf:
Edit to serve commands (assuming the installer has been used), as well as adding stepper and tabs, and content for step five.


reakaleek (approved):
Stepper ❤️

PR #1279, opened 2025-05-20
Add a new job to handle push event matching

Mpdreamz:
Introduce a "match" job to process push event-specific logic and outputs, ensuring proper handling of event types. This includes conditional steps for different event contexts and prepares outputs for downstream "build" job dependencies.

### DIFF
--- a/.github/workflows/preview-build.yml
+++ b/.github/workflows/preview-build.yml
@@ -10,6 +10,13 @@ on:
     branches:
       - main
       - master
+      # TODO remove these need to be added to individual repositories
+      - '\d+.\d+.\d+'
+      - '\d+.\d+'
+      - '\d+.x'
+    tags:
+      - 'v?\d+.\d+.\d+'
+      - 'v?\d+.\d+'
   workflow_call:
     inputs:
       strict:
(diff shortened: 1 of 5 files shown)


reakaleek (commented):
We could change `on.pull_request_target` to `on.pull_request` to test if it's working on this PR. 

Without affecting other repos.

.github/workflows/preview-build.yml
```diff
@@ -44,6 +51,37 @@ permissions:
...
+    outputs:
+      content-source-match: ${{ steps.event-check.outputs.content-source-match != '' && steps.event-check.outputs.content-source-match || steps.match.outputs.content-source-match }}
+      content-source-name: ${{ steps.event-check.outputs.content-source-name != '' && steps.event-check.outputs.content-source-name || steps.match.outputs.content-source-name }}
+    steps:
+      - name: Not a push event
+        id: event-check
+        if: contains(fromJSON('["merge_group", "pull_request", "pull_request_target"]'), github.event_name)
+        run: |
+          echo "content-source-match=true >> $GITHUB_OUTPUT
+          echo "content-source-name=next >> $GITHUB_OUTPUT
```
reakaleek, line 69: IIUC, these are just some "random" values so that the build runs on PRs.

Can we add a comment saying that. (Assuming my assumption is true)

Mpdreamz:
@reakaleek I think this is now ready for a spin or our `main` and temporary `9.0` branches. 

I can then follow up with making `build` depend on `match`'s output.

reakaleek (approved):

PR #1294, opened 2025-05-21
Fix NRE in LinkIndexProvider while adding a new branch to existing repository

Mpdreamz:


src/infra/docs-lambda-index-publisher/LinkIndexProvider.cs
```diff
@@ -74,7 +88,11 @@ public async Task Save()
 			ContentType = "application/json",
 			IfMatch = _etag // Only update if the ETag matches. Meaning the object has not been changed in the meantime.
 		};
-		_ = await s3Client.PutObjectAsync(putObjectRequest);
-		logger.LogInformation("Successfully saved link index to s3://{bucketName}/{key}", bucketName, key);
+		var putReponse = await s3Client.PutObjectAsync(putObjectRequest);
+		if (putReponse.HttpStatusCode == HttpStatusCode.OK)
+			logger.LogInformation("Successfully saved link index to s3://{bucketName}/{key}", bucketName, key);
+		else
+			// TODO retry?
```
reakaleek, line 95: The message is returned to the queue if this fails and then retried 10 times until it's sent to the DLQ.

reakaleek: The main reason why it would fail is if the loaded link index in memory is out-of-date with the object in s3

Mpdreamz: Ahh in that case we should throw an exception right so its caught and put on the queue?

reakaleek: Oh, yes!

reakaleek (approved):

PR #1298, opened 2025-05-21
[Fix] Clarify path specification for the serve command

alexandra5000:
This PR clarifies how to specify the path when using the `serve` command. The updated wording makes it easier for users to understand how to correctly point to the `docset.yml` file.


Mpdreamz (approved):
thank you @alexandra5000 !

PR #1302, opened 2025-05-21
Refactor LinkIndexProvider into own project

reakaleek:
So we can reuse it in assembler to get the link index.


src/Elastic.Documentation.LinkIndex/LinkIndexProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+
+using System.Net;
+using Amazon.Runtime;
+using Amazon.S3;
+using Amazon.S3.Model;
+using Elastic.Documentation.Links;
+
+namespace Elastic.Documentation.LinkIndex;
+
+public interface ILinkIndexProvider
```
Mpdreamz, line 13: Is it worth splitting this into `ILinkIndexReadProvider` and have `ILinkIndexProvider` inherit from that? 

Most usages do not need access to the `Save()` operation.

src/Elastic.Documentation.LinkIndex/LinkIndexProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+{
+	Task<LinkReferenceRegistry> GetLinkIndex(Cancel cancellationToken = default);
```
Mpdreamz, line 15: Let's settle this once and for all 😸 

The `link-index.json` at the root is the 'registry' of all the `links.json` files we publish.


* `links.json` == `DocumentationSetLinks` || `RepositoryLinks`
* `links-index.json` == `LinkRegistry` 


That way we can update the type names and methods to be the same and them being unique enough from each other to be able to differentiate. 

e.g

```
Task<LinkReferenceRegistry> GetLinkIndex(Cancel cancellationToken = default);
Task<LinkReference> GetLinkReference(string key, Cancel cancellationToken = default);
```

```
Task<LinkRegistry> GetLinkRegistry(Cancel cancellationToken = default);
Task<RepositoryLinks> GetRepositoryLinks(string key, Cancel cancellationToken = default);
```

Ideally `links-index.json` should be renamed to `links-registry.json` too would be too invasive. 


Names are just a suggestion of course!

src/Elastic.Documentation.LinkIndex/LinkIndexProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+	Task SaveLinkIndex(LinkReferenceRegistry registry, Cancel cancellationToken = default);
+	Task<LinkReference> GetLinkReference(string key, Cancel cancellationToken = default);
+	string GetLinkIndexPublicUrl();
```
Mpdreamz, line 18:
```suggestion
	string LinkIndexPublicUrl { get; }
```

Mpdreamz (changes requested):
Nice! 

I added some refactoring suggestions but this LGTM 👍

src/Elastic.Documentation.LinkIndex/LinkIndexProvider.cs
reakaleek: Thank you for the nice suggestions. Unfortunately, I did all in one commit.

https://github.com/elastic/docs-builder/pull/1302/commits/18309d9ce9be51fb5576b1e0d47d703725afddfa

Mpdreamz (approved):
Fantastic! this unblocks us from using the registry as a build manifest 🚀

PR #1307, opened 2025-05-22
Fix `npm run watch`

reakaleek:
## Context

Related issue https://stackoverflow.com/questions/79615955/cannot-read-properties-of-undefined-reading-input-error

The error only happens when using the `npm run watch` command.

## Changes

- downgrade tailwindcss (related issue: https://github.com/parcel-bundler/parcel/issues/10161)
- downgrad parcel (segmentation fault caused by https://github.com/parcel-bundler/parcel/releases/tag/v2.15.0)


src/Elastic.Markdown/Assets/styles.css
```diff
@@ -287,3 +287,7 @@ body {
 		@apply bg-grey-20 rounded-2xl;
 	}
 }
+
+body {
+	background-color: green;
+}
```
reakaleek, line 293:
```suggestion
```

Mpdreamz (approved):

PR #1316, opened 2025-05-26
Also write the used link-index to the filesystem and add it as part of the new CheckoutResult record

reakaleek:
## Changes

- Write the `link-index.shapshot.json` file to the checkout dir while cloning repositories with `clone-all`
- During `build-all` copy the `link-index.snapshot.json` to the `/docs` output dir.

### DIFF
--- a/src/tooling/docs-assembler/Sourcing/RepositorySourcesFetcher.cs
+++ b/src/tooling/docs-assembler/Sourcing/RepositorySourcesFetcher.cs
@@ -102,7 +102,7 @@ public class AssemblerRepositorySourcer(ILoggerFactory logger, AssembleContext c
 					checkouts.Add(RepositorySourcer.CloneRef(repo.Value, gitRef, fetchLatest));
 				}, c);
 			}).ConfigureAwait(false);
-		await File.WriteAllTextAsync(
+		await context.WriteFileSystem.File.WriteAllTextAsync(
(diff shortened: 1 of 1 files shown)


src/tooling/docs-assembler/Sourcing/RepositorySourcesFetcher.cs
```diff
@@ -91,8 +102,23 @@ await Task.Run(() =>
...
+			ctx
+		);
+		return new CheckoutResult
+		{
+			Checkouts = checkouts,
+			LinkRegistrySnapshot = linkRegistry
+		};
 	}
+
+	public async Task WriteLinkRegistrySnapshot(LinkRegistry linkRegistrySnapshot, Cancel ctx = default) => await File.WriteAllTextAsync(
```
Mpdreamz, line 117:
```suggestion
	public async Task WriteLinkRegistrySnapshot(LinkRegistry linkRegistrySnapshot, Cancel ctx = default) => await context.WriteFileSystem.File.WriteAllTextAsync(
```

Mpdreamz (changes requested):
One nit, use the filesystems on context not the `System.IO.*` classes directly.

src/tooling/docs-assembler/Sourcing/RepositorySourcesFetcher.cs
reakaleek: Also changed it here: https://github.com/elastic/docs-builder/pull/1316/commits/6c2f39d8b6410682fc8988bce98e23bcc6f38b49

PR #1320, opened 2025-05-27
Add exporter options to assembler

Mpdreamz:
This adds the following export options to assembler. 

* `html`, the default renders HTML
* `llm`, only half implemented renders the markdown as markdown (with substitutions resolved)
* `elasticsearch`, writes documentation files to Elasticsearch (currently requires `ELASTIC_URL` and `ELASTIC_API_KEY` to be set)

These can be specified as:

```
./docs-assembler repo build-all --exporters llm,elasticsearch
```

In which case no html will be emitted, only the llm text files and Elasticsearch.

Submitting this PR  early so I can more cleanly follow up with a more complete LLM and Elasticsearch export implementation without the infrastructure noise :)


reakaleek (approved):
🚀

PR #1334, opened 2025-06-02
Add `docs-assembler deploy [plan|apply]` commands

reakaleek:
## Changes

This gives us the ability to analyze the plan and identify which files are being uploaded, deleted or skipped

## Usage

```shell
docs-assembler repo clone-all --environment edge
docs-assembler repo build-all --environment edge
docs-assembler deploy plan --environment edge --s3-bucket-name elastic-docs-v3-website-edge --out out.plan 

# ... Do something else with the plan? E.g. optimize for cache invalidation.

docs-assembler deploy apply --environment edge --s3-bucket-name elastic-docs-v3-website-edge --plan-file out.plan
```

## Example Output

```shell
info ::Program               :: Total files to sync: 14525
info ::Program               :: Total files to delete: 0
info ::Program               :: Total files to add: 0
info ::Program               :: Total files to update: 1725
info ::Program               :: Total files to skip: 12800
info ::Program               :: Plan written to out.plan
```

```json
{
  "add": [
    { 
      "local_path": "/path-to-local/docs/add.md", 
      "destination_path": "/docs/add.md"
    }
  ],
  "update": [
    { 
      "local_path": "/path-to-local/docs/update.md",
      "destination_path": "/docs/update.md"
    }
  ],
  "skip": [
    { 
      "local_path": "/path-to-local/docs/skip.md",
      "destination_path": "/docs/skip.md"
    }
  ],
  "delete": [
    { 
      "destination_path": "/docs/delete.md"
    }
  ]
}
```

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -16,7 +16,7 @@
     <PackageVersion Include="AWSSDK.Core" Version="4.0.0.2" />
     <PackageVersion Include="AWSSDK.SQS" Version="4.0.0.1" />
     <PackageVersion Include="AWSSDK.S3" Version="4.0.0.1" />
-    <PackageVersion Include="Moq" Version="4.20.72" />
+    <PackageVersion Include="FakeItEasy" Version="8.3.0" />
     <PackageVersion Include="Elastic.Ingest.Elasticsearch" Version="0.11.3" />
   </ItemGroup>
   <!-- Build -->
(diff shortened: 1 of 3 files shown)


Mpdreamz:
> If there is a change in a static file the content-hash should change. This means the link tag content also changes and therefore it will basically do a full sync because all the ETags change.

I guess my question also was if in that case instead of writing all files out into the plan we write a special marker instead (`fullsync: true`) or something like that. But even if all content needs to be synced it makes sense to still do a partial update so that we don't have to upload all images.

reakaleek:
> > If there is a change in a static file the content-hash should change. This means the link tag content also changes and therefore it will basically do a full sync because all the ETags change.
> 
> I guess my question also was if in that case instead of writing all files out into the plan we write a special marker instead (`fullsync: true`) or something like that. But even if all content needs to be synced it makes sense to still do a partial update so that we don't have to upload all images.

Ah, I see. Sorry for misunderstanding. Makes sense!

PR #1345, opened 2025-06-05
Publish beats docs from the `9.0` branch

colleenmcginnis:
⚠️  Wait until https://github.com/elastic/beats/pull/44682 + backport are merged

Update assembler config to publish Beats documentation from the `9.0` branch. 

cc @belimawr

### DIFF
--- a/src/tooling/docs-assembler/assembler.yml
+++ b/src/tooling/docs-assembler/assembler.yml
@@ -65,6 +65,8 @@ references:
   apm-aws-lambda:
   apm-k8s-attacher:
   beats:
+    current: "9.0"
+    next: main
   cloud-on-k8s:
   cloud:
     current: master


src/tooling/docs-assembler/assembler.yml
```diff
@@ -65,6 +65,8 @@ references:
   apm-aws-lambda:
   apm-k8s-attacher:
   beats:
+    current: "9.0"
+    next: main
```
belimawr, line 69: Can we have `next` as `9.1`? That would ensure we don't publish `main` again by accident and that once `v9.1.0` is released, there won't be any unreleased features in our docs.

belimawr (approved):

src/tooling/docs-assembler/assembler.yml
Mpdreamz: You didn't :)

Mpdreamz:
Ahh automerge! :) Will wait with releasing this and ensuring production reflects assembler.yml till monday.

PR #1358, opened 2025-06-12
Allow specifying which version to install

bmorelli25:
We have a couple folks that are blocked because the current release of docs-builder is broken. They're unsure how to go about installing a previous version. This fix allows a specific version to be specified if necessary but defaults to latest.


theletterf:
Very nice!

PR #1359, opened 2025-06-12
Add new product to product list

theletterf:
This adds the EDOT Cloud Forwarder `product` to the list and the docs. It's a different EDOT family of ingest solutions.

See https://github.com/elastic/edot-cloud-forwarder-aws (slated to launch for 9.1 in its first iteration for AWS).


Mpdreamz (approved):
LGTM

PR #1365, opened 2025-06-13
Add cloud-enterprise legacy mapping

reakaleek:


colleenmcginnis (approved):
Approved (with the caveat that we might want `4.0` to be `4.0+`, but it might more sense to make a decision and apply that decision to all versioning systems at once) !

PR #1389, opened 2025-06-18
Move new version dropdown to sidebar and add missing behaviour from the old dropdown

reakaleek:
## Changes

- Move the new EUI dropdown to the sidebar
- Add "View all versions" link to version dropdown
- Add "View previous docs versions" CTA link

## Notes

The new dropdown is still hidden behind a feature flag.

We can enable after it was QAed.

## Recording

https://github.com/user-attachments/assets/1bd53aa9-d67f-455f-b3df-4a639228f67b


Mpdreamz (approved):
Lgtm! The empty side nav sits a bit wrong with me but I have no suggestions on how to improve as of now. 

Either way nothing to block this PR over!

Nice work !

PR #1404, opened 2025-06-20
Fix flaky static initialization error

reakaleek:
As seen in https://github.com/elastic/docs-builder/actions/runs/15775018706/job/44467383605?pr=1403#step:5:96


Mpdreamz (changes requested):
This is sadly not a long term fix, since the initialization might happen on any test following this pattern which is all. 

Its static so that it will be shared amongs every instance of the type. Each `[Fact]` method will instantiate a new instance of the containing type. 

Removing `static` alone would thus be enough to ensure each gets a non shared instance.

The static/lazy stuff might be a premature optimalization on my end.

PR #1406, opened 2025-06-20
Replace APM agents with OpenTelemetry in landing page

theletterf:
This replaces the "APM agents" box of the https://www.elastic.co/docs with "OpenTelemetry", linking to the EDOT reference documentation. Strategically, this is where we want to go; it's a small step in a long walk of changes we have to make. 

Whether it's the right time to do so right now can be debated, of course. I would argue that it's more symbolic than anything, with very limited impact on clicks. Symbolically, it's significant.

Before going forward, I'd like to hear from @colleenmcginnis @bmorelli25.


bmorelli25:
Wow. It's happening!

I don't have strong opinions on the timing here. I know you've been chatting with the relevant product and eng folks. If the sense you're getting is that now is the right time to start making these changes, then I'm for it 👍

Issue #1395, opened 2025-06-18
[Feature Request]: Add `doc-preview-comment.yml` functionality to the doc build


colleenmcginnis:
### Prerequisites

- [x] I have searched existing issues to ensure this feature hasn't already been requested
- [x] I have tested using the latest version of docs-builder

### What problem are you trying to solve?

In https://github.com/elastic/docs-content/pull/1341, @bmorelli25 added a GitHub action to the docs-content repo that adds a comment with links from each changed file to the built location in the deployment preview to make it easier for reviewers.

### Proposed Solution

We should consider adding this functionality to the doc build so it can be used across repos without adding the maintenance burden of managing individual workflows in each repo.

### How important is this feature to you?

Nice to have

shainaraskas:
https://github.com/elastic/elasticsearch/pull/129779: waiting on eng to fix the labels
https://github.com/elastic/cloud-on-k8s/pull/8696: waiting for an [assist with merging](https://elastic.slack.com/archives/C8RU9J1C2/p1750437231211759) 
https://github.com/elastic/ecctl/pull/730: cp apps should get around to this but will keep an eye out

PR #1416, opened 2025-06-23
Add preview links comment to PRs

reakaleek:
Closes https://github.com/elastic/docs-builder/issues/1395

## Changes

When the upload to s3 completes, comment on the PR with a list of changed files.

## Related Issues

- Depends on linked PRs in https://github.com/elastic/docs-builder/issues/1395 to be merged. (Adding the `pull-requests: write` permission)

### DIFF
--- a/.github/workflows/preview-build.yml
+++ b/.github/workflows/preview-build.yml
@@ -239,7 +239,7 @@ jobs:
           
       - name: Comment on PR
         continue-on-error: true
-        if: inputs.disable-comments != 'true' && env.MATCH == 'true' && steps.deployment.outputs.result && steps.check-files.outputs.all_changed_files
+        if: startsWith(github.event_name, 'pull_request') && inputs.disable-comments != 'true' && env.MATCH == 'true' && steps.deployment.outputs.result && steps.check-files.outputs.all_changed_files
         uses: actions/github-script@v7
         env:
           ALL_CHANGED_FILES: ${{ steps.check-files.outputs.all_changed_files }}


.github/workflows/preview-build.yml
```diff
@@ -231,6 +237,70 @@ jobs:
...
+            const links = changedMdFiles.map(toLink)
+            
+            const body = [
+              title,
+              ...links.slice(0, 20).map(i => `- ${i}`),
+            ]
+            
+            if (links.length > 20) {
+              body.push('');
+              body.push(`<sub>There are ${links.length - 20} more links...</sub>`);
```
shainaraskas, line 276: do these additional links get exposed anywhere?

.github/workflows/preview-build.yml
```diff
@@ -112,6 +117,7 @@ jobs:
         id: check-files
         uses: tj-actions/changed-files@2f7c5bfce28377bc069a65ba478de0a74aa0ca32 # v46.0.1
         with:
+          safe_output: false # set to false because we are using an environment variable to store the output and avoid command injection.
```
Mpdreamz, line 120: How does this work? won't this still emit unsanitzed files names on standard out? 

Is setting this unsafe because of the action or because of possible usages of the `steps.check-files.outputs.all_changed_files` in follow up scripts?

reakaleek: I got this from the [examples section](https://github.com/tj-actions/changed-files?tab=readme-ov-file#examples-).

The "Get all changed files without escaping unsafe filename characters" example. 

But I might have copy-pasted to soon. I guess this is not needed. Thanks for noticing!

reakaleek, line 120:
```suggestion
```

reakaleek: https://github.com/elastic/docs-builder/pull/1416/commits/6c50add5b2affa7ccb375aaf7306e78250c3f360

.github/workflows/preview-build.yml
```diff
@@ -1,7 +1,7 @@
 name: preview-build
 
 on:
-  pull_request_target:
+  pull_request:
```
v1v, line 4: IIUC, this will not let you use `write` permissions scopes. Is that the expected behaviour?

bmorelli25 (approved):

.github/workflows/preview-build.yml
reakaleek: https://github.com/elastic/docs-builder/pull/1416/commits/73066e1924ed232d89768c1974fa3cf1d0bfd244

PR #1417, opened 2025-06-23
Finish LLM text exporter

Mpdreamz:
- Reorganize Elastic.Markdown so that code is grouped by purpose not type

Thought this was a prerequisite for LLM text by utilizing markdigs round trip serializer as a base of our own. However that heavily relies on parsing with Trivia something we can not do because it breaks list continuations. See https://github.com/elastic/docs-builder/pull/435

We now manually parse includes and re-evaluate substitutions on the full included files. Adding the exporter does not add much overhead in addition to the HTML exporter

This emits a `filename.md` next to each `filename/index.html` 

image

In addition it emits a `llm.zip` that can be used to download everything at once.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,6 +17,9 @@ updates:
       day: 'monday'
       time: '08:00'
     groups:
+      eui:
+        patterns:
+          - '@elastic/eui*'
       tailwindcss:
         patterns:
           - 'tailwindcss'


theletterf:
Niceeee! How does this work? Do we generate the final Markdown from an intermediate representation/AST? It's important that the final file we produce has the same content as the rendered HTML, that is, resolved substitutions, etc.

PR #1385, opened 2025-06-17
Add EDOT SDK release notes

theletterf:
This adds EDOT SDK release notes. To be merged once all tocs are live.

PR #1418, opened 2025-06-23
Update `NavigationPrefixChecker` to skip phantom paths during link resolution.

Mpdreamz:
We need to do this on both branches as otherwise it requires a certain order for phantom branches to not be validated.

Fixes: https://github.com/elastic/elastic-otel-node/actions/runs/15825603447/job/44606696330

The check was failing unrelated to anything `elastic-otel-node`. Our check relied on `elasticsearch://reference` being checked before `docs-content://reference` but in the current iteration of the links.json checks this did not happen because our dictionaries are unordered (as they should be).

### DIFF
--- a/src/tooling/docs-assembler/Links/NavigationPrefixChecker.cs
+++ b/src/tooling/docs-assembler/Links/NavigationPrefixChecker.cs
@@ -98,7 +98,7 @@ public class NavigationPrefixChecker
 
 			// Todo publish all relative folders as part of the link reference
 			// That way we don't need to iterate over all links and find all permutations of their relative paths
-			foreach (var (relativeLink, _) in linkReference.Links)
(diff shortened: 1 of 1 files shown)


theletterf (approved):

cotti (approved):

theletterf:
@cotti @Mpdreamz Seems like the problem now cascades down one step:

```
Error: 'elasticsearch' defines: 'reference/' that 'docs-content://reference/index.md resolving to '/docs/reference/' conflicts with 

	The following errors and warnings were found in the documentation

Error: 'elasticsearch' defines: 'reference/' that 'docs-content://reference/index.md resolving to '/docs/reference/' conflicts with 
NOTE: docs-content
```

See https://github.com/elastic/docs-builder/pull/1385

Mpdreamz:
That PR is out of date, needs an update :)

PR #1423, opened 2025-06-24
add version number has to hx-get to break cache on version updates

Mpdreamz:


src/Elastic.Markdown/Myst/Renderers/HtmxLinkInlineRenderer.cs
```diff
@@ -25,7 +32,8 @@ protected override void Write(HtmlRenderer renderer, LinkInline link)
 				return;
 			}
 
-			var url = link.GetDynamicUrl != null ? link.GetDynamicUrl() : link.Url;
+			var url = link.GetDynamicUrl?.Invoke() ?? link.Url;
```
reakaleek, line 35: I'm not sure if we want this here. 

we only want to set it in the `hx-get` attribute.

otherwise, clicking on the link will contain the querystring

reakaleek (commented):
I think it's still missing for the navigation links.

src/Elastic.Documentation.Site/Htmx.cs
```diff
@@ -24,8 +31,9 @@ public static string GetHxAttributes(
 		string? hxIndicator = HxIndicator
 	)
 	{
+		var hxGetUrl = targetUrl.Contains('?') ? $"{targetUrl}&v={VersionHash}" : $"{targetUrl}?v={VersionHash}";
```
reakaleek, line 34: It seems like the targetUrl may already contain a fragment and this might created a malformed url.

PR #1458, opened 2025-06-26
Fix crosslinks

reakaleek:
## Context

Crosslinks resolved to an absolute URL including the domain. The logic in our parser renderer did not consider this.

## Changes

Also add htmx attributes to crosslinks.


reakaleek:
Oh, wait. We are doing this so the link is clickable in docs-builder build.

Mpdreamz (approved):

Issue #1464, opened 2025-06-27
[Feature Request]: Automatically add anchors to stepper items, and add them to the ToC


theletterf:
### Prerequisites

- [x] I have searched existing issues to ensure this feature hasn't already been requested
- [x] I have tested using the latest version of docs-builder

### What problem are you trying to solve?

The Stepper component is great, but I wish the steps could be deeplinked through anchors somehow, and appeared in the ToC, with a level beneath the heading acting as a parent for the component.

### Proposed Solution

Modify the Stepper component to:

- Add anchors to the step titles.
- Make the step titles appear in the ToC as children of the parent heading.

### How important is this feature to you?

Nice to have

PR #1472, opened 2025-06-30
Add semantic headings to stepper component

theletterf:
This PR updates the Stepper component so that:

- Step titles are rendered as an `<Hn>`  heading, where `n` is the level of the preceding heading + 1.
- Heading thus appear in the table of content if they're of level 3.
- Step titles are *not* included in the ToC if they're nested inside another directive to avoid weird situations, such as competing headings in tabs using steppers and headings hidden inside a dropdown. Strange, but could happen.

Fixes https://github.com/elastic/docs-builder/issues/1464

### Example of dynamic headings (h4 after an h3)

headings

### Example of H3 step titles appearing in ToC

Screenshot 2025-06-30 at 10 57 55


theletterf:
@reakaleek Good point. For now I would keep the number size the same.

Issue #1466, opened 2025-06-27
[docs] Document how to check for build errors locally


colleenmcginnis:
Add an optional section to the [Contributing locally](https://elastic.github.io/docs-builder/contribute/locally/) about checking for build errors locally before pushing up to GitHub (this is the same as serving the site, just without the word `serve`). While this step is not required when contributing locally, it would be helpful to have it documented so we can point contributors to it if/when they ask.

PR #1474, opened 2025-06-30
Add step for building the docs

theletterf:
Fixes https://github.com/elastic/docs-builder/issues/1466

Notice that the preview process is now left last, as we want to encourage local docs builds and error checking.


theletterf:
Build errors — ironic! :D I'll fix those now. Also the whitespace anomalies if I can.

colleenmcginnis:
>Notice that the preview process is now left last, as we want to encourage local docs builds and error checking.

I'm not sure that I agree with this. I think it is valuable to run the preview _while_ writing so the contributors to see the changes they're making. Then the can check for build errors before committing/pushing to GitHub. 🤔

PR #1488, opened 2025-07-01
Add `ech` alias for `ess` and add hint to prefer `ech`

reakaleek:


src/Elastic.Markdown/Myst/FrontMatter/ApplicableTo.cs
```diff
@@ -78,11 +78,15 @@ public record DeploymentApplicability
...
 
+	[YamlMember(Alias = "ech")]
+	public AppliesCollection? Ech { get; set; } // ess alias, preferred
+
 	public static DeploymentApplicability All { get; } = new()
 	{
 		Ece = AppliesCollection.GenerallyAvailable,
 		Eck = AppliesCollection.GenerallyAvailable,
 		Ess = AppliesCollection.GenerallyAvailable,
+		Ech = AppliesCollection.GenerallyAvailable,
```
Mpdreamz, line 89: I don't think this should be a new property, rename `Ess` to `Ech` and have both `ess` and `ech` assign to `Ech`. 

Specifying both should be an error.

PR #1494, opened 2025-07-01
[docs] Fixing broken link in what books were migrated

eedugon:
The modified link was incorrect. I have updated it to the main page of the `asciidocalypse` repository.

But I still have a concern... because we say:

> A full list of the content that was migrated to the V3 format is available here

But in the `asciidocalypse` repo we don't have such a list (or I can't find it), but the leftovers (migrated content not used), if I'm not mistaken.

Where is the actual list that we should be offering?

### DIFF
--- a/.github/workflows/preview-build.yml
+++ b/.github/workflows/preview-build.yml
@@ -43,12 +43,17 @@ on:
         type: string
         default: 'false'
         required: false
+      disable-comments:
+        description: 'Disable comments'
+        type: boolean
+        default: false
+        required: false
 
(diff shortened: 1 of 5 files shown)


reakaleek:
I think what you are referring to is https://github.com/elastic/asciidocalypse/tree/main-archive/docs#readme

eedugon:
thanks @reakaleek ! That was exactly it! Good to know that we have all original content in the `main-archive` branch.
I've updated the PR to point to that link now.

PR #1499, opened 2025-07-02
preview-build: Add path-pattern-ignore

reakaleek:
So we can ignore some files and so that the files are not included in the comment. 
(I'm not 100% sure if the files will actually be excluded from the `changed_files` output. But that is my assumption)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,6 +17,11 @@ updates:
       day: 'monday'
       time: '08:00'
     groups:
+      eslint:
+        patterns:
+          - 'eslint'
+          - '@eslint/*'
+          - 'typescript-eslint'
       eui:
         patterns:
           - '@elastic/eui*'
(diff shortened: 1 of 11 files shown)


leemthompo:
Sorry it's two separate things:

- any .md file under under [/docs/reference/query-languages/esql/kibana](https://github.com/elastic/elasticsearch/tree/main/docs/reference/query-languages/esql/kibana)

And any random README file under docs

Because we're doing this centrally I assumed you'd have to do it all in one place, how would we configure this per repo?

reakaleek:
> Sorry it's two separate things:
> 
> * any .md file under [/docs/reference/query-languages/esql/kibana](https://github.com/elastic/elasticsearch/tree/main/docs/reference/query-languages/esql/kibana)
> * and any random README file under docs
> 
> Because we're doing this centrally I assumed you'd have to do it all in one place, how would we configure this per repo?

You can change the configuration here: https://github.com/elastic/elasticsearch/blob/main/.github/workflows/docs-build.yml

And then you can actually do:

```
with:
  path-pattern: docs/**
  path-pattern-ignore: |
    docs/reference/query-languages/esql/kibana/*.md
    docs/**/README.md
```

PR #1511, opened 2025-07-02
add ECCTL to release notes nav

shainaraskas:
related to https://github.com/elastic/docs-content-internal/issues/47

- adds ECCTL to the release notes nav
- bumps current for ECCTL
- depends on https://github.com/elastic/ecctl/pull/732


src/tooling/docs-assembler/assembler.yml
```diff
@@ -94,7 +94,7 @@ narrative:
 #  cloud:
```
bmorelli25, line 94: @shainaraskas I think this is the wrong file. @reakaleek we should remove these comments to avoid confusion.

This is defined here: https://github.com/elastic/docs-builder/blob/e55ea1bbadd3db4d23dd632100a7a887cd39557f/src/Elastic.Documentation.Configuration/versions.yml#L26

It looks like we already have 1.15.0 marked as current.

bmorelli25: We had to move versioning out of `assembler.yml` as it is needed for individual repo builds as well.

reakaleek: Aw! Clean up will follow.

shainaraskas: sorry, my b everyone.

bmorelli25: Updated in `450b51b`

reakaleek: Does 1.13 link anywhere?

bmorelli25: Discussed on zoom. Fixed in https://github.com/elastic/docs-builder/pull/1511/commits/3a9965ea8920e18e882696b4d355f89600cae2b2

PR #1514, opened 2025-07-03
Optimize lazy load navigation

reakaleek:
## Context

follow-up from https://github.com/elastic/docs-builder/pull/1505

Currently we are create a .nav.html file for every page that exists.
However, there are many duplicates as the navigation for sub pages are probably the same.

## Changes

- Only create a single .nav.html partial file for each cached navigation. This makes it possible to cache the navigation in browser.

### DIFF
--- a/src/Elastic.Documentation.Site/Navigation/INavigationHtmlWriter.cs
+++ b/src/Elastic.Documentation.Site/Navigation/INavigationHtmlWriter.cs
@@ -21,7 +21,7 @@ public interface INavigationHtmlWriter
 }
 public record NavigationRenderResult
 {
-	public static NavigationRenderResult Empty => new()
+	public static NavigationRenderResult Empty { get; } = new()
 	{
 		Html = string.Empty,
 		Id = "empty-navigation" // random id


src/Elastic.Documentation.Site/Navigation/INavigationHtmlWriter.cs
```diff
@@ -10,11 +10,30 @@ public interface INavigationHtmlWriter
...
+	string Id { get; init; }
+}
+
+public record OkNavigationRenderResult : INavigationRenderResult
+{
+	public required string Html { get; init; }
+	public required string Id { get; init; }
+}
+
+public record EmptyNavigationRenderResult : INavigationRenderResult
```
Mpdreamz, line 35: This can be a static property on `OkNavigationRenderResult` ? 

e.g `OkNavigationRenderResult.Empty` then we can just have a single record `NavigationRenderResult` without an interface.

Mpdreamz (changes requested):

src/Elastic.Documentation.Site/Navigation/INavigationHtmlWriter.cs
reakaleek: https://github.com/elastic/docs-builder/pull/1514/commits/64966753aa8eb7150b3cedb9e43f2ea4e556b306

src/Elastic.Documentation.Site/Navigation/INavigationHtmlWriter.cs
```diff
@@ -10,11 +10,23 @@ public interface INavigationHtmlWriter
...
 
 	async Task<string> Render(NavigationViewModel model, Cancel ctx)
 	{
 		var slice = _TocTree.Create(model);
 		return await slice.RenderAsync(cancellationToken: ctx);
 	}
 }
+public record NavigationRenderResult
+{
+	public static NavigationRenderResult Empty => new()
```
Mpdreamz, line 24:
```suggestion
	public static NavigationRenderResult Empty { get; } = new()
```

Mpdreamz (changes requested):

PR #1525, opened 2025-07-07
Load elastic nav asynchronously

reakaleek:


reakaleek:
> LGTM, just a q? will this create layout shifts?

It might, but it already does IMO.

When you load a page freshly, you always see the navigation blinking.

Mpdreamz:
SGTM 👍

reakaleek:
> SGTM 👍

The layout shift actually got worse with this..

That's probably the reason why we didn't add it in the first place.

But since it also happens with `async` in prod.. I will try to fix the min-height thing.

PR #1526, opened 2025-07-07
Optimize preview build permissions

reakaleek:
## Changes

- For each job only add the needed permissions following the principle of least privilege.
- Move the comment logic into it's own job.
- Move the check files into it's own job so its output can be consumed by the 'build' and 'comment' job.


Mpdreamz (changes requested):
Should check depend on match?

PR #1539, opened 2025-07-08
Allow dangling toc definitions in navigation.yml

Mpdreamz:
This allows documentation sets to use `toc.yml` files for grouping purposes without them being declared in `navigation.yml` as long as their parent navigation element is. 

This leniency allows for e.g `docs-content` to partition docs folders for other teams without those teams having to worry about the larger navigation.


src/tooling/docs-assembler/Navigation/GlobalNavigationHtmlWriter.cs
```diff
@@ -32,13 +32,25 @@ private bool TryGetNavigationRoot(
 		navigationRootSource = null;
 		if (!assembleSources.TocTopLevelMappings.TryGetValue(navigationSource, out var topLevelMapping))
```
Mpdreamz, line 33: Long term all of this will go away and we can just query the navigation itself..

reakaleek (approved):
LGTM.

I tested it locally and it fixes the missing navigation in opentelemetry.

PR #1541, opened 2025-07-08
Add `stack` and `stack_next` variables to `assembler.yml`

colleenmcginnis:
I'm working on updating the docs release checklist for Elastic Stack releases. In the current draft, I'm recommending that the docs release coordinator touches this file twice:

* **The day after feature freeze** (after the new version branch is created): The docs release coordinator will update `stack_next` to the version of the upcoming release. The idea is that we'll be publishing those docs to staging (right?) so we'll be able to catch any build errors (like any broken cross-repo links) ahead of the release date. Does this make sense?
* **The day of the release**: The docs release coordinator will update `stack` to the new version and update `stack_next` to `main` (I don't _think_ any Stack-versioned repos still use `master`).

Let me know what you think of this approach. I'm very open to suggestions.

### DIFF
--- a/src/tooling/docs-assembler/assembler.yml
+++ b/src/tooling/docs-assembler/assembler.yml
@@ -17,7 +17,7 @@ environments:
       auth: nPocPUG0wiH68jsVeyRSxA
       preview: env-507
       cookies_win: x
-  edge: 
+  edge:
     uri: https://d34ipnu52o64md.cloudfront.net
     path_prefix: docs
     content_source: current
@@ -31,7 +31,8 @@ environments:
     path_prefix: docs
 
 named_git_references:
-  stack: &stack 9.0
+  stack: &stack_current 9.0
+  stack_next: &stack_next 9.1
   cloud-hosted: ms-120
 
 ###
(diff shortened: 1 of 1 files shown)


src/tooling/docs-assembler/assembler.yml
```diff
@@ -32,6 +32,7 @@ environments:
 
 named_git_references:
   stack: &stack 9.0
```
reakaleek, line 34: This could because something like 
```suggestion
  stack: &stack
    current: 9.0
    next: 9.1
```

And then you only do 

```yaml
beats: *stack
```

reakaleek:
@colleenmcginnis what happens when the `9.1` branch is created?

Is it an exact snapshot of `main`? Or are the additional changes.

src/tooling/docs-assembler/assembler.yml
colleenmcginnis: I couldn't get something like that to work locally. Does this suggestion work for you?

reakaleek: You're right. It doesn't work for me as well.. Although I think it should. 

Probably some quick on how we desrialize the config file.

Mpdreamz:
I guess my main concern is that staging is either `main` or `Major.Minor`.

After feature freeze there would be no environment for `main`. 

With the edge environment always being `main` we can at least always inspect the main integration branch of all repositories.

PR #1563, opened 2025-07-15
feature/variable mutation operators

Mpdreamz:
- **Add support for version context variables**
- **Add support for variable operator syntax**


reakaleek (approved):
nice

PR #1568, opened 2025-07-16
Add ConfigurationExporter to export all configurations to a central location.

Mpdreamz:
In preparation of separating configuration from releases.

> ls -al .artifacts/assembly/config/

```
drwxr-xr-x  5 mpdreamz  staff    160 Jul 16 16:01 ./
drwxr-xr-x  5 mpdreamz  staff    160 Jul 16 16:01 ../
-rw-r--r--  1 mpdreamz  staff   2909 Jul 14 11:24 assembler.yml
-rw-r--r--  1 mpdreamz  staff   3715 Jul  7 12:40 legacy-url-mappings.yml
-rw-r--r--  1 mpdreamz  staff  27434 Jul 16 15:36 navigation.yml
```

This also centralize our reading of configuration files in one location to make it easier to swap it out for something remote later.

All configuration files now live under `/config` in our codebase so external contributors do not need to hunt for these files.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -16,7 +16,6 @@
     <PackageVersion Include="AWSSDK.Core" Version="4.0.0.2" />
     <PackageVersion Include="AWSSDK.SQS" Version="4.0.0.1" />
     <PackageVersion Include="AWSSDK.S3" Version="4.0.0.1" />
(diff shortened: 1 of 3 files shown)


reakaleek (approved):
do we also need the versions.yml config?

Mpdreamz:
yes we do!

Issue #1434, opened 2025-06-24
Decouple assembler config from docs-builder release process


bmorelli25:
## Summary

We need to be able to release a new version of the docs without having to release a new version of docs-assembler. This will allow us to move quickly and release multiple version updates in quick succession—both of which are often necessary on release days.

If we don't do this, it will require docs eng to be around during releases (and no one wants that 😉 ).

## Requirment

- [x] Enable writers to release a new version of the docs without needing to release a new version of docs-builder
- [x] Enable writers to add a new content set without needing to release a new version of docs-builder

bmorelli25:
We need to also explore decoupling adding new content sets from the release process. See [this slack thread](https://elastic.slack.com/archives/C07APH4RCDT/p1751570820757369) (sorry internal link) for any example of why we want to be able to add new content without creating a new release.

bmorelli25:
We will go live with this **after 9.1** so as to not make a big change right before the release.

bmorelli25:
@Mpdreamz can you pls scope this with the remaining work needed to decouple version bumps from the release process.

bmorelli25:
@colleenmcginnis updated our internal docs release process documentation in https://github.com/elastic/dev/pull/3224 & https://github.com/elastic/docs-content-internal/pull/111. I've updated the task list for this issue to include updating those internal docs with the new release process (post-decoupling).

reakaleek:
@Mpdreamz will add context:
- what has been done already
- what needs to be done

Mpdreamz:
Upate release process doc in https://github.com/elastic/docs-content-internal/pull/378

Mpdreamz:
Update release templates in dev https://github.com/elastic/dev/pull/3321

PR #1580, opened 2025-07-18
feature/config folder arg

Mpdreamz:
This allows the build to source the configuration prior to cloning and building

docs-assembler repo init-config
docs-assembler repo clone-all
docs-assembler repo build-all

On production we can lock the configuration to a commit using `--git-ref sha`.

I think we want to automate opening PR's to update that sha to the last commit on our `config` folder. We can then merge it once we know staging is still good.

ConfigurationFileProvider will look at

- {pwd}/config, development
- {app_data}/config, (where init-config clones too)
- embedded sources

We also now log the configuration source and in case of app_data also the git reference.

### DIFF
--- a/src/Elastic.Documentation.Configuration/ConfigurationFileProvider.cs
+++ b/src/Elastic.Documentation.Configuration/ConfigurationFileProvider.cs
@@ -4,13 +4,25 @@
 
 using System.IO.Abstractions;
 using Microsoft.Extensions.DependencyInjection;
+using NetEscapades.EnumGenerators;
 
 namespace Elastic.Documentation.Configuration;
+
+[EnumExtensions]
+public enum ConfigurationSource
+{
+	Local,
+	Checkout,
+	Embedded
+}
 
 public class ConfigurationFileProvider
 {
 	private readonly IFileSystem _fileSystem;
 	private readonly string _assemblyName;
+
+	public ConfigurationSource ConfigurationSource { get; private set; } = ConfigurationSource.Embedded;
+	public string? GitReference { get; }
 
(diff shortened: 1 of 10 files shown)


src/Elastic.Documentation.Configuration/ConfigurationFileProvider.cs
```diff
@@ -4,14 +4,26 @@
...
 using System.IO.Abstractions;
 using Microsoft.Extensions.DependencyInjection;
+using NetEscapades.EnumGenerators;
 
 namespace Elastic.Documentation.Configuration;
 
+[EnumExtensions]
+public enum ConfigurationSource
+{
+	Local,
```
reakaleek, line 14: Can we use local and put the config in elastic/docs-internal-workflows?

reakaleek (approved):

src/Elastic.Documentation.Configuration/ConfigurationFileProvider.cs
Mpdreamz: We could, as source of truth or as a copy?

I think we want the source of truth to remain in docs-builder.

Mpdreamz: I think config should live where its read. 

If it lives in elastic/docs-internal-workflows we'd need multiple copies for each environment. 

We can do a lot more PR validations from this repository too IMO.  Let's discuss today :)

bmorelli25:
Sounds good. Closes https://github.com/elastic/docs-builder/issues/1434.

PR #1596, opened 2025-07-21
Better approach to set the title on directives in LLM markdown output

reakaleek:
## Changes

- Set missing title for Stepper StepBlock.
- Refactor how the LlmDirectiveRenderer determines if it needs to render a title attribute


reakaleek:
> Would prefer `IBlockTitle` but LGTM

https://github.com/elastic/docs-builder/commit/72250cb924fd9ef25a979c325aeeefecb9214c49

Issue #1605, opened 2025-07-22
Mutation operators do not work with version variables in code blocks and links


colleenmcginnis:
### Describe the bug

[Mutation operators](https://elastic.github.io/docs-builder/syntax/version-variables/#formatting) do not work with version variables in code blocks.

### Expected behavior

I'd expect `{{version.stack | M.M}}` to be replaced with `9.0` in code blocks with `subs=true`.

### Steps to reproduce

Here's what I tried:

````md
To get the manifests just run:

{{version.stack | M.M}}

```sh subs=true
curl -L -O https://raw.githubusercontent.com/elastic/beats/{{version.stack | M.M}}/deploy/kubernetes/auditbeat-kubernetes.yaml
```
````

Rendered result:

<img width="857" height="200" alt="Image" src="https://github.com/user-attachments/assets/919d033f-298d-4581-a7db-feaa821cce21" />

### Tooling

- [x] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

PR #1606, opened 2025-07-23
Fix mutation operators not working in links and code blocks

theletterf:
Fixes https://github.com/elastic/docs-builder/issues/1605

Also adds tests for mutation operators.


Mpdreamz (changes requested):
One more change to ensure we don't needlessly roundtrip already parsed enums

Mpdreamz (approved):

theletterf:
OK, I think it's ready now :)

Issue #1593, opened 2025-07-21
Custom TOC/nav links


bmorelli25:
## Summary

Add the ability to customize the TOC with links to pages that don't exist within the current content set. The goal is be able to interweave internal (elastic.co/docs/*) and external links to content into the navigation structure of a content set. This will allow us to, for example, link to API docs from within a narrative TOC.

If the content being linked to is within a different content set, the TOC should update to reflect the move to a new content set.

theletterf:
Agree with rediscussing this. Intuitively, it feels like we should go from 10 main sections to perhaps five. Solutions could contain (should contain?) also Reference and Troubleshooting docs, for example.

theletterf:
Agree. I think the external links path is the simplest, also to avoid crosslink resolution in the nav:

- You define an external link like this:

```
toc:
  - link: https://elastic.co/docs/something
```

It would render like a nav item, but with an external link indicator, like we added here: https://github.com/elastic/docs-builder/pull/1096

theletterf:
The way this is being implemented is to allow crosslinks in the nav. In that sense, a doc will only exist once in nav, but we'll be able to link to it from other navs.

reakaleek:
deploy to production already

PR #1613, opened 2025-07-25
Add external links to navigation

theletterf:
Fixes https://github.com/elastic/docs-builder/issues/1593

### DIFF
--- a/docs/_docset.yml
+++ b/docs/_docset.yml
@@ -147,3 +147,5 @@ toc:
           - folder: baz
             children:
               - file: qux.md
+  - link: https://github.com/elastic/docs-builder
+    title: GitHub repository
--- a/docs/configure/content-set/navigation.md
+++ b/docs/configure/content-set/navigation.md
@@ -159,6 +159,39 @@ toc:
   - toc: solutions
 ```
 
+#### External links
+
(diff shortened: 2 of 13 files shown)


theletterf:
@bmorelli25 See the GitHub repository link in https://docs-v3-preview.elastic.dev/elastic/docs-builder/pull/1613/configure/content-set/navigation

leemthompo:
oooooh I immediately want to use this to add a link to the Elasticsearch reference docs in the Elasticsearch solution nav 😍 

would this be a valid use case?

theletterf:
@Mpdreamz @bmorelli25 After melting Claude's brain, it looks like adding crosslinks validation and rendering to the toc is too much. LLMs being unable to figure this out suggests that crosslinks depend deeply on the overall architecture, but correct me if I'm wrong. 

For a quick solution, I'd rather show external links as they appear now and, if they start with `elastic.co/docs`, open them as if they were normal toc links (no icon, no _blank target). How does that sound?

theletterf:
@Mpdreamz I love that. Any hint on how to achieve that though?

theletterf:
CSS issue solved for this particular PR, @Mpdreamz! We shouldn't abuse this system though, I agree.

PR #1618, opened 2025-07-28
Search or Ask AI beta webcompontent

reakaleek:
## Changes

Adds the "Search or Ask AI" button to the isolated build, but only works locally when you provide the correct GCP credentials.

## Follow ups

Testing. The initial testing set up I created, felt overcomplicated and required a lot of patches. I will follow-up with a clean test setup.

## Notes

Suggestions are hard coded.

For the MVP we can come up with better hard coded values.

In the future we want to have a suggestions API or something like that.

## Preview 

https://github.com/user-attachments/assets/5328229e-b404-4b40-a4c4-6c78e4ffa576


Mpdreamz (approved):
Great work on the custom GCP id token generation !

Issue #1655, opened 2025-07-30
[Feature Request]: Pass ALT text value as title automatically for inline images


florent-leborgne:
### Prerequisites

- [x] I have searched existing issues to ensure this feature hasn't already been requested
- [x] I have tested using the latest version of docs-builder

### What problem are you trying to solve?

Is there a way we can automatically pass image ALT values as title values as well? Example:
`![Hide missing values](../images/charts-gaps-fill-hide.png "title =50%")` shows this when hovering an image:

<img width="720" height="427" alt="Image" src="https://github.com/user-attachments/assets/ab4b22a1-fafa-45fd-b253-672a25682aee" />

`<img src="/elastic/docs-content/pull/2210/explore-analyze/images/charts-gaps-fill-hide.png" width="50%" height="50%" alt="Hide missing values" title="title">`

- We must resize our images in many cases, so if we [start from the example in the internal docs](https://elastic.github.io/docs-builder/syntax/images/#inline-image-sizing), we all use `"title =<the-size>"` so we'll see a proliferation of "title" titles.

Alternatively, if we know we don't want to do this or cannot, let us know as well so we can plan to check these "title" values in existing docs and improve our accessibility/SEO

### Proposed Solution

- Automatically pass the ALT value as title if possible
- Update the internal docs: the "inline image sizing" should focus on the size parameter and exclude the dummy title, if that is even possible to specify the size without a title

### How important is this feature to you?

Nice to have

PR #1656, opened 2025-07-30
Automatically use alt text as title for images when no title is provided

theletterf:
Fixes https://github.com/elastic/docs-builder/issues/1655

When adding images to documentation, users often need to specify sizing parameters (e.g., `=50%`) in the title attribute. This creates a redundancy issue where the alt text needs to be duplicated as the title before adding the size parameters, resulting in Markdown like:

`Description`

This PR modifies the `ImageBlock` class to automatically use the alt text as the title value when no explicit title is provided. This allows users to:

1. Skip the redundant title text when only specifying size parameters:
   `Description`

2. Not have to duplicate information when using image directives.

## Changes

- Modified the `FinalizeAndValidate` method in `ImageBlock.cs` to use alt text as the title when title is not specified
- Updated the documentation to explain this new behavior
- Updated test cases to reflect the new HTML output that includes title attributes

### DIFF
--- a/docs/syntax/images.md
+++ b/docs/syntax/images.md
@@ -29,6 +29,8 @@ Or, use the `image` directive.
 :::
 ```
 
+Note: When using the image directive, if you specify an `:alt:` value but no `:title:` value, the alt text will automatically be used as the title.
+
 :::{image} /syntax/images/observability.png
 :alt: Elasticsearch
 :width: 250px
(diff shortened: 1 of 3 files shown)


Mpdreamz (approved):
Great QoL improvement 👍

theletterf:
@Mpdreamz @reakaleek To make things simpler, I've made it so that Alt text always overrides the title, whether it's present or not. In most cases, the titles we have are not useful.

florent-leborgne:
This is already super great! Is it possible to also check if the current title is plain "title" and in that case overwrite it with the alt value too? :D 
Kind of a safety net

This could also be something we instead flag with a linter for example, if we don't want to bake it in docs-builder :)

PR #1664, opened 2025-07-31
Use Aspire as distributed service orchestrator, start on integration tests

Mpdreamz:
- **Initial infra work for assembler integration tests**
- **Move to Aspire**
- **Move integration tests over to aspire**

Support `--skip-private-repositories` on `clone-all` `build-all` to ensure we can run an assemble build token less. This allows anyone to bootstrap our integration tests even outside the company.


reakaleek (approved):
💅

PR #1666, opened 2025-08-01
API lambda adjustments

reakaleek:


src/api/Elastic.Documentation.Api.Lambda/Program.cs
```diff
@@ -5,20 +5,22 @@
...
 var builder = WebApplication.CreateSlimBuilder(args);
 
 builder.Services.AddAWSLambdaHosting(LambdaEventSource.RestApi, new SourceGeneratorLambdaJsonSerializer<LambdaJsonSerializerContext>());
-builder.Services.AddElasticDocsApiUsecases(Environment.GetEnvironmentVariable("APP_ENVIRONMENT"));
+builder.Services.AddElasticDocsApiUsecases(Environment.GetEnvironmentVariable("ENVIRONMENT") ?? "dev");
 
 var app = builder.Build();
 
-var v1 = app.MapGroup("/v1");
+var v1 = app.MapGroup("/docs/_api/v1");
```
Mpdreamz, line 18: We need a few reserved namespaces in the markdown build :). I have `/docs/api` for the OpenAPI too.

src/infra/docs-lambda-index-publisher/lambda.DockerFile
```diff
@@ -31,5 +31,4 @@ RUN arch=$TARGETARCH \
     && if [ "$arch" = "amd64" ]; then arch="x64"; fi \
     && echo $TARGETOS-$arch > /tmp/rid
 
-RUN dotnet publish src/infra/docs-lambda-index-publisher -r linux-x64 -c Release
-
+RUN dotnet publish src/infra/docs-lambda-index-publisher -r $(cat /tmp/rid) -c Release
```
Mpdreamz, line 34: Do we need to do the same here: https://github.com/elastic/docs-builder/blob/main/src/infra/docs-lambda-index-publisher/lambda.DockerFile ?

reakaleek: Hehe, I unintentionally changed the wrong file..

I will keep as is, because it works.

PR #1667, opened 2025-08-01
Add info comment to preview build

reakaleek:
## Context

As an attempt to centralize https://github.com/elastic/elasticsearch/pull/132257

## Changes

Add a step to the preview build that comments about versioning.

## Usage

```diff
name: docs-build

on:
  push:
    branches:
      - main
  pull_request_target: ~
  merge_group: ~

jobs:
  docs-preview:
    uses: elastic/docs-builder/.github/workflows/preview-build.yml@main
    with:
      path-pattern: docs/**
      path-pattern-ignore: docs/changelog/**/*.yaml
+     enable-cumulative-comment: true
    permissions:
      deployments: write
      id-token: write
      contents: read
      pull-requests: write
```


Mpdreamz (changes requested):
It'd be nice if we can rename disable-comments to disable-changes-comment, might be hard unless we support both for a while.

Mpdreamz (approved):

.github/workflows/preview-build.yml
```diff
@@ -438,3 +444,58 @@ jobs:
                 body:body.join('\n'),
               });
             }
+      - name: Comment on docs changes about versioning requirements
+        if: inputs.enable-cumulative-comment == 'true'
```
leemthompo, line 448:
```suggestion
        if: inputs.enable-cumulative-comment == 'true'
        && needs.check.outputs.any_modified == 'true'
```

@Mpdreamz I'm sure it's not as simple as that 😆 

Had forgotten about this again until I looked at my TODOs, would be nice to get this change in for elasticsearch tho

reakaleek: The whole job already has the condition to only trigger on changes

Mpdreamz: doh missed that!

Issue #1308, opened 2025-05-22
[Feature Request]: Support substitutions in inline code


eedugon:
### Prerequisites

- [x] I have searched existing issues to ensure this feature hasn't already been requested
- [x] I have tested using the latest version of docs-builder

### What problem are you trying to solve?

At the moment we don't support substitutions in inline code blocks, like:

\`{{my-sub}}\`.

In some cases we generate sections of the pages from snippets that receive certain substitutions, and we should be able to play with the format of those substitutions.

There's an example of this need in this PR: https://github.com/elastic/docs-content/pull/1457

### Proposed Solution

Implement something similar as with the normal code blocks that support the subs=true option.

### How important is this feature to you?

Nice to have

PR #1705, opened 2025-08-08
Add subs support for inline code snippets

theletterf:
Fixes https://github.com/elastic/docs-builder/issues/1308

This adds the following notation for inline code subs:

```
{subs}`This is an inline snippet with {{version.stack}} a sub`
```


theletterf:
@Mpdreamz Set this as a `{subs}` role, WDYT?

Mpdreamz (approved):
Love it!

PR #1731, opened 2025-08-13
Reapply #1708 Add ElasticsearchGateway and UI adjustments"

Mpdreamz:
- **Reapply "Add ElasticsearchGateway and UI adjustments (#1708)" (#1730)**
- **move to later runtime deps image**

PR #1736, opened 2025-08-13
Add Search/AI Assistant to dev enviroment.

Mpdreamz:
This also exposes the proxy to `docs-builder serve-static`.

It now also wires up the configuration for our dotnet aspire project.

If the secrets have not been configured yet aspire will prompt to provide them

image

Make sure `Save to user secrets` is checked:

image

After which you'll need to restart the aspire application again. 

Once `DocsBuilderServeStatic` is started (it waits for clone-all and build-all) it will have the experimental documentation search bar enabled. 

The search is still static (pending #1731) but now the AI responses are real.

image

### DIFF
--- a/config/assembler.yml
+++ b/config/assembler.yml
@@ -30,6 +30,8 @@ environments:
     uri: http://localhost:4000
     content_source: current
     path_prefix: docs
+    feature_flags:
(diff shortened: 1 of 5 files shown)


cotti (approved):
Awe-_aspiring_. LGTM

Issue #1751, opened 2025-08-20
Tooltip missing lifecycle status text for unavailable features


leemthompo:
### Describe the bug

<img width="720" height="210" alt="Image" src="https://github.com/user-attachments/assets/018a40be-25e7-4238-8a23-81e37962a2f3" />

**Description:**
Tooltip shows incomplete text - missing lifecycle status at the beginning of the sentence.

**Root cause:** `lifecycleFull` variable is empty/null when `ProductLifecycle.Unavailable`, causing the tooltip template to render with missing text.

## Remedy

The issue is that `lifecycleFull` isn't being set properly for the `Unavailable` lifecycle state, so the tooltip template renders with an empty string at the beginning.

https://github.com/elastic/docs-builder/blob/main/src/Elastic.Markdown/Myst/Components/ApplicableToComponent.cshtml#L253

I nominate @theletterf to fix this.

### Expected behavior

**Expected:** "Unavailable on Elastic Cloud Serverless unless otherwise specified."
**Actual:** "on Elastic Cloud Serverless unless otherwise specified."

### Steps to reproduce

Read this issue

### Tooling

- [x] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

PR #1752, opened 2025-08-20
Fix "unavailable" tooltip text missing in applicability badges

theletterf:
This adds the missing `unavailable` ProductLifecycle value, mapped to `Not available`.

Fixes https://github.com/elastic/docs-builder/issues/1751


Mpdreamz:
It's gone because I removed it :)

PR #1777, opened 2025-08-25
Warn when a heading contains an inline `applies_to`

colleenmcginnis:
Warns when a heading contains an inline `applies_to`.

@theletterf and I paired today on adding a warning when we detect inline `applies_to` syntax in the heading text. We tried a couple different approaches. This one works, but I'm not sure if it's the best place to check for this.

Open questions for @elastic/docs-engineering:

* Is this an appropriate place to check for `applies_to` syntax in a heading?
* Does it make sense to use regex to check?
* Is there a way to add more context (like the line number and highlighting the text that's causing the error) to the warning message? Right now it just looks like this:
  Screenshot 2025-08-25 at 12 34 10 PM


Mpdreamz:
I just implemented it without looking at @cotti's review but ended up doing the same thing. NICE!

image

Mpdreamz (approved):

PR #1782, opened 2025-08-26
diff validate improvements

Mpdreamz:
- Normalize path handling between this command and others. Path is now specified using `--path`
- local file changes check could potentially pop unrelated changes back if nothing was stashed in `docs` folder to begin with.
- `RedirectFile` path localization is now part of the constructor.
- `_snippets` folders are now excluded from the redirect file checks.


src/Elastic.Documentation/Extensions/IFileInfoExtensions.cs
```diff
@@ -42,4 +49,20 @@ public static bool IsSubPathOf(this IDirectoryInfo directory, IDirectoryInfo par
 
 		return false;
 	}
+
+	/// Checks if <paramref name="directory"/> has parent directory <paramref name="parentName"/>
+	public static bool HasParent(this IDirectoryInfo directory, string parentName)
+	{
+		if (directory.Name == parentName)
```
cotti, line 56: nit: `directory.Name.Equals(parentName, StringComparison.OrdinalIgnoreCase)`

Issue #1436, opened 2025-06-24
Support `applies_to` in additional directives/components


bmorelli25:
## Summary

Add custom support for `applies_to` tags in the following components:

* Tabs
* Dropdowns
* Admonitions

### Tabs

Automatically convert a single `applies_to` badge into a tab title, so users can immediately see which version or platform each tab applies to without expanding it.

#### Behavior

If the `applies_to:` prop is not null:

* Set the `applies_to` **text** as the tab title
* Add the `applies_to` badge to the top of the tab content
* Error if a tab title exists

#### Limitations

* Multiple `applies_to` tags per tab not supported as tab titles would get too long
* This is a short term fix. Some tabs will need need `applies_to` tab sections each release which would lead to way too many tabs in the widget.

#### Syntax

```
::::{tab-set}
:::{tab-item}
applies_to: stack: beta 9.0`
This is where the content for tab #1 goes.
:::
:::{tab-item}
applies_to: stack: ga 9.1`
This is where the content for tab #2 goes.
:::
:::{tab-item}
applies_to: serverless: ga
This is where the content for tab #3 goes.
:::
::::
```

#### Output

<img width="714" height="194" alt="Image" src="https://github.com/user-attachments/assets/dd81c9cf-2699-47f3-925e-0d955b27823e" />

<img width="686" height="174" alt="Image" src="https://github.com/user-attachments/assets/33ed7105-cc25-4193-baf4-652d9e99e575" />

### Drop downs

Make version/platform context visible in dropdown titles so users can make informed choices before expanding the content.

#### Behavior

If the `applies_to:` prop is not null:

* Prepend the **text** of the `applies_to` badge to the dropdown title
* Add the `applies_to` badge to the top of the dropdown content

#### Limitations

* Multiple `applies_to` tags per dropdown not supported as titles would get too long

#### Syntax

```
:::{dropdown} Rules are failing due to number of alerts
applies_to: stack: all
Dropdown content
:::
:::{dropdown} Rules are failing due to number of alerts
applies_to: eck: all
Dropdown content
:::
```

#### Output

<img width="692" height="252" alt="Image" src="https://github.com/user-attachments/assets/72852a12-5f40-4fa0-9756-44d0bbf4085f" />

### Admonitions

Allow authors to add applies_to badges to admonitions so users can quickly identify whether the content is relevant to their deployment environment. Admonitions are a little different from tabs and dropdowns as all context is available to the user without having to click on anything.

#### Behavior

[...]

PR #1791, opened 2025-08-28
#1436 Support applies_to in additional directives/components

siamakp-elastic:
Implements: 
    https://github.com/elastic/docs-builder/issues/1436


bmorelli25 (changes requested):
We want to move forward with a text-based implementation initially. I've updated  https://github.com/elastic/docs-builder/issues/1436 accordingly.

PR #1794, opened 2025-08-29
Fixes #63 - Breadcrumbs issue in new docs pages

siamakp-elastic:


reakaleek:
IIUC, we want to show the full breadcrumbs.

I still see an ellipsis in https://docs-v3-preview.elastic.dev/elastic/docs-builder/pull/1794/testing/breadcrumbs/level-3/level-4/level-5/level-5

How does the current change fix the issue?

PR #1796, opened 2025-09-01
Add AI/LLM files to .gitignore

theletterf:
This adds common AI agent context / instructions files to .gitignore, so they can be used in AI-powered tools / IDEs.


theletterf:
Hmm, I agree with @reakaleek, let's create on a common file.

PR #1816, opened 2025-09-02
Sync URL search params with search term

reakaleek:


reakaleek:
> I wonder if these need to be linkable? The popover is a navigation aid. E.g the query can happen on ALL the pages.
> 
> We should discuss the full UX of this and whether we want a generic /search page or if we should forward that to the generic site search page?
> 
> If we do want them to be linked we need to take care these are not susceptible to SEO poisoning.
> 
> We will need to update our parent's robots.txt which might be hard/messy if these `q=?` can appear on all the pages.
> 
> https://www.elastic.co/robots.txt

IIUC, the `<link rel="canonical" href="..." />` should avoid this problem.

Mpdreamz:
Not a bad idea! Should never conflict with our slugs.

PR #1823, opened 2025-09-03
Use environment aware elasticsearch index names

reakaleek:
## Changes

Add the environment name to the index name, allowing us to have separate indexes for each environment.


Mpdreamz (approved):
LGTM apart from the compile error.

Issue #1552, opened 2025-07-11
[Feature Request]: Change how the `title` in the `head` is defined


colleenmcginnis:
### Prerequisites

- [x] I have searched existing issues to ensure this feature hasn't already been requested
- [x] I have tested using the latest version of docs-builder

### What problem are you trying to solve?

We've noticed that some pages are being marked as duplicates by search engines even when they are not duplicates. The SEO team found that some pages have the same `title` in the `head` that in the AsciiDoc build system had different titles. For example, the `title` on both of these pages are the same (`Configure Kerberos | Beats`):

* https://www.elastic.co/docs/reference/beats/filebeat/configuration-kerberos
* https://www.elastic.co/docs/reference/beats/metricbeat/configuration-kerberos

I _think_ the second part of the `title`, `Beats`, is coming from the title of the first page in the docset based on [this](https://github.com/elastic/docs-builder/blob/main/src/Elastic.Markdown/Page/Index.cshtml#L17) and this:

<img width="1663" height="452" alt="Image" src="https://github.com/user-attachments/assets/72a30f3f-9196-45fb-b046-b437169e49a1" />

### Proposed Solution

I'm not sure _how_ we should define the `title` instead, but in the example above using `Filebeat` and `Metricbeat` instead of `Beats` would help make the `title`s unique. Happy to brainstorm other options! 

### How important is this feature to you?

Important

reakaleek:
What are your thoughts on adding Filebeat or Metricbeat actually to the title?

I just saw the same pattern for the ["Configure Filebeat" page](https://www.elastic.co/docs/reference/beats/filebeat/configuring-howto-filebeat). The navigation_title is just "Configure", but the actual title is "Configure Filebeat".

In the same manner, the "Configure Kerberos" title could be "Configure Kerberos for Filebeat" and the navigation_title remains "Configure Kerberos".

This would also help the user directly identify for what the page is. Right now you have to look at the navigation or breadcrumbs to understand that it's for a specific beat.

colleenmcginnis:
>What are your thoughts on adding Filebeat or Metricbeat actually to the title?

>This would also help the user directly identify for what the page is about. Right now you have to look at the navigation or breadcrumbs to understand that it's for a specific beat.

I'd be fine with that. It wouldn't be difficult to make those changes to the existing pages, but I wonder if we could/should implement some kind of check to ensure we don't end up with more duplicate titles in the future.

cotti:
Implementation notes:

[...]

PR #1832, opened 2025-09-05
Implements #1552: Change how the title in the head is defined.

siamakp-elastic:


src/Elastic.Markdown/DocumentationGenerator.cs
```diff
@@ -99,6 +99,36 @@ public async Task ResolveDirectoryTree(Cancel ctx)
...
 		await DocumentationSet.Tree.Resolve(ctx);
 		_logger.LogInformation("Resolved tree");
+		ReportDuplicateTitles(DocumentationSet.Tree.ResolvedMarkdownFiles);
+	}
+
+	private void ReportDuplicateTitles(List<MarkdownFile> files)
+	{
+		// Create a dictionary where keys are the titles
+		// and values are files with that title
+		var titleMap = new Dictionary<string, List<MarkdownFile>>(StringComparer.OrdinalIgnoreCase);
```
Mpdreamz, line 109: `HashSet<string>` will suffice we care about something being duplicate not necessary how many times since the diagnostics reporting will take care of that.

I think it makes more sense to implement this as `IMarkdownExporter` implementations (maybe `ValidationMarkdownExporter`). 

That way it works for both `docs-builder` and `docs-assembler`. Here the duplicate file checking is per `DocumentationSet` not overall.

src/Elastic.Markdown/DocumentationGenerator.cs
```diff
@@ -99,6 +99,36 @@ public async Task ResolveDirectoryTree(Cancel ctx)
...
+		foreach (var file in files)
+		{
+			if (string.IsNullOrWhiteSpace(file.Title))
```
Mpdreamz, line 112: This should be reported as an error.

PR #1836, opened 2025-09-05
[docs] Add more guidance on `products` metadata

### LINKED ISSUE #1818: [docs] Improve docs on using `products` metadata
While reviewing https://github.com/elastic/apm-agent-dotnet/pull/2650, I realized how little information we have in the docs-builder docs about how to use `products` metadata. 

In addition to the [list on the frontmatter docs page](https://elastic.github.io/docs-builder/syntax/frontmatter/#products), we should also add:

* Documentation on using products in `docset.yml` to apply metadata to every page in a doc set.
* Guidelines on how to choose which `products` to include. Some factors to consider:
  * Think about how users will use this data via search[^1] (for example, if a user was searching for information on Elastic Cloud Serverless and filters to pages that are tagged with `cloud-serverless`, does it make sense to include a page from the APM .NET agent docs?).
  * Avoid just duplicating information in `applies_to` (for example, just because a page describes functionality that is applicable to a specific deployment method doesn't mean the page is _about_ that deployment method).
* Anything else? 

[^1]: Note: I'm not sure if `products` are going to be used differently in the new search experience that's being developed. @KOTungseth and @reakaleek might be able to clarify.

cc @theletterf

colleenmcginnis:
Fixes https://github.com/elastic/docs-builder/issues/1818

Adds more guidance on `products` metadata. I left a couple code comments inline where I'm not sure what we want to recommend.


docs/_snippets/products-list.md
```diff
@@ -0,0 +1,51 @@
+`products` is a list of objects, each object has an `id` field.
+
+```yaml
+products:
+  - id: apm
+```
+
+Each product ID is mapped to a full product name.
```
theletterf, line 8: Where's the mapping defined? Shall we point to the source code?

docs/configure/content-set/navigation.md
```diff
@@ -49,6 +49,11 @@ Optional key. Defaults to `false`. When enabled turns soft line endings in the m
 
 ### `external_hosts`
 
+% Or remove this section altogether?
```
theletterf, line 52: +1

docs/configure/content-set/navigation.md
```diff
@@ -107,7 +123,7 @@ The TOC in principle follows the directory structure on disk.
   - folder: subsection
 ```
 
-If a folder does not explicitly define `children` all markdown files within that folder are included automatically 
+If a folder does not explicitly define `children` all markdown files within that folder are included automatically
```
theletterf, line 126:
```suggestion
If a folder does not explicitly define `children` all Markdown files within that folder are included automatically
```

docs/syntax/frontmatter.md
```diff
@@ -43,43 +43,28 @@ See [](./applies.md)
...
-| `search-ui`                                 | Search UI                                     |
-| `security`                                  | Elastic Security                              |
-| `winlogbeat`                                | Winlogbeat                                    |
+
+:::{include} /_snippets/products-list.md
+:::
+
+`products` can also be defined in the [`docset.yml` file](/configure/content-set/navigation.md#products).
+If you define `products` in a page's Markdown file and the `docset.yml` file also includes `products`, docs-builder will combine the two lists.
+You can _not_ override doc set level `products` at the page level.
```
theletterf, line 52: Since this seems to be repeated, shall we add it to the include?

docs/syntax/frontmatter.md
```diff
@@ -43,43 +43,28 @@ See [](./applies.md)
...
+
+:::{tip}
+`products` is distinct from `applies_to`, which is used to indicate feature availability and applicability.
```
theletterf, line 55:
```suggestion
The purpose of the `products` field is different from `applies_to`: the former provides metadata to search engines, while the latter indicates feature availability and applicability.
```

docs/syntax/frontmatter.md
```diff
@@ -43,43 +43,28 @@ See [](./applies.md)
...
+
+% Is this true?
+For example, while a page might contain content that is _applicable_ to Elastic Cloud Serverless projects,
+it might not be _about_ Elastic Cloud Serverless. In that case, you would include the `serverless` key in `applies_to`,
+but _not_ include the `cloud-serverless` ID in `products`:
```
theletterf, line 60: Could we perhaps replace this with a small table of When to use vs When not to use?

theletterf (commented):
Thanks! Some suggestions and comments as a recent `products` user.

docs/syntax/frontmatter.md
reakaleek: This makes sense.

But the original reason was that `products` are only used for the search.elastic.co filters. At the time of implementation, we did not know how `applies_to` and `products` relate to.

Not sure if we even know this today. (this statement is actually a good explanation) But yes, there were ongoing discussions that they are similar. same same but different.

reakaleek (approved):

docs/syntax/frontmatter.md
colleenmcginnis: The general idea is repeated, but the content is not identical: one is on the frontmatter page calling out that it's also possible to define products in the `docset.yml` file and the other in on the navigation page calling out that it's also possible to define in the frontmatter.

docs/syntax/frontmatter.md
theletterf: @colleenmcginnis Examples, I would say. Just a suggestion though, this is not blocking the PR.

docs/syntax/frontmatter.md
```diff
@@ -43,43 +43,28 @@ See [](./applies.md)
...
+
+:::{include} /_snippets/products-list.md
+:::
+
+`products` can also be defined in the [`docset.yml` file](/configure/content-set/navigation.md#products).
+If you define `products` in a page's Markdown file and the `docset.yml` file also includes `products`, docs-builder will combine the two lists.
+You can _not_ override doc set level `products` at the page level.
+
+:::{tip}
+The purpose of the `products` field is different from `applies_to`: the former provides metadata to search engines, while the latter indicates feature availability and applicability.
```
florent-leborgne, line 55:
```suggestion
The purpose of the `products` field is different from `applies_to`: the former provides metadata to elastic.co search engines, while the latter indicates feature availability and applicability.
```
This is just about our own searches, isn't it?

docs/syntax/frontmatter.md
florent-leborgne: This is a tricky example. I would add the corresponding product(s) in all cases.

Meaning that "A page with content that applies to Elastic Cloud Serverless should also be identified with the "cloud-serverless" product ID", to be searchable as such using the elastic.co search. I think this is true in most cases. If the content applies to "Stack", then you'd want product ids to identify which stack components are relevant here, etc.

lcawl: > Meaning that "A page with content that applies to Elastic Cloud Serverless should also be identified with the "cloud-serverless" product ID", to be searchable as such using the elastic.co search. 

In that case wouldn't we want that to be derived automatically from the `applies_to` at some point, rather than requiring the duplication? Then the `products` would only need to include the level of product detail (components of the stack, etc) that aren't covered in the other metadata

florent-leborgne: But overall this seems logical, if we mark content as applicable to a context, why wouldn't we allow users to find that content when selecting that context through facets?

cotti (approved):

Mpdreamz:
@cotti can you ensure this gets merged in?

docs/syntax/frontmatter.md
colleenmcginnis: I just removed this for now. We'll have to revisit it as a part of the cumulative docs initiative.

PR #1853, opened 2025-09-08
Use POST request to check if the API is available

reakaleek:
GET requests are cached on the edge. which means, once a user with VPN gets a 200 OK, all users in the region get a 200 OK, which is not the intended behaviour

POST requests are not cached


Mpdreamz (approved):
Why POST and not just the GET (while we investigate the HEAD issue on fastly?)

PR #1871, opened 2025-09-10
Add nav and assembler changes for EDOT docs move

theletterf:
Contributes to https://github.com/elastic/opentelemetry-dev/issues/1044


config/navigation.yml
```diff
@@ -333,7 +333,26 @@ toc:
       # https://github.com/elastic/opentelemetry/blob/main/docs/reference/toc.yml
       - toc: opentelemetry://reference
         path_prefix: reference/opentelemetry
-        # Children include all EDOT distribution docs.
+        children:
+          - toc: elastic-agent://reference/edot-collector
+            path_prefix: reference/opentelemetry/edot-collector
+          - toc: opentelemetry://reference/edot-sdks
+            path_prefix: reference/opentelemetry/edot-sdks
+            children:
```
colleenmcginnis, line 341: I think the problem is that the [`reference/toc.yml`](https://github.com/elastic/opentelemetry/blob/remove-migrated-docs/docs/reference/toc.yml#L23) file in the opentelemetry repo includes the `reference/edot-sdks/index.md` file. This means that when you define `toc: opentelemetry://reference` here, you're already including a page at path `reference/opentelemetry/edot-sdks` so you can't add another toc (`opentelemetry://reference/edot-sdks`) using that `path_prefix`.

Note: We ran into a similar problem with APM agents. That's why we ended up with the `path_prefix` [`reference/apm-agents`](https://github.com/elastic/docs-builder/blob/main/config/navigation.yml#L371) instead of `reference/apm/agents` (which would better align with the pattern of all other `path_prefix`es in that section).

If you want the final paths for the intro page and all the individual EDOT SDK repos to follow the same pattern (`reference/edot-sdks/*`), I was able to get this to work locally by:

1. In the opentelemetry, renaming the directory [`docs/reference/edot-sdks/`](https://github.com/elastic/opentelemetry/tree/remove-migrated-docs/docs/reference/edot-sdks) to `docs/intros/edot-sdks/`.
2. Removing [this line from `docs/reference/toc.yml`](https://github.com/elastic/opentelemetry/blob/remove-migrated-docs/docs/reference/toc.yml#L23).
3. Adding `- toc: intros/edot-sdks` to [`docs/docset.yml`](https://github.com/elastic/opentelemetry/blob/remove-migrated-docs/docs/docset.yml#L32) (after `- toc: reference`).
    * Note: The `docs/docset.yml` is only used to build the PR previews in each repo. It has no impact on the production site. So you really just need to add this extra page somewhere so it can be built in previews to both test that the syntax/links are valid and preview the rendered page.

This is what the **production site** will look like: 

Screenshot 2025-09-15 at 4 24 13 PM

This is what the **PR previews** will look like:

Screenshot 2025-09-15 at 4 27 47 PM

config/navigation.yml
```diff
@@ -333,7 +333,26 @@ toc:
...
+                path_prefix: reference/opentelemetry/edot-sdks/python
+              - toc: elastic-otel-node://reference
+                path_prefix: reference/opentelemetry/edot-sdks/node
+              - toc: elastic-otel-php://reference
+                path_prefix: reference/opentelemetry/edot-sdks/php
+              - toc: apm-agent-android://reference
+                path_prefix: reference/opentelemetry/edot-sdks/android
+              - toc: apm-agent-ios://reference
+                path_prefix: reference/opentelemetry/edot-sdks/ios
+              - toc: apm-agent-dotnet://reference
```
colleenmcginnis, line 354:
```suggestion
              - toc: elastic-otel-dotnet://reference
```

config/navigation.yml
theletterf: I ended up putting SDK docs inside subfolders and defining different nav paths. I'll deal with this through redirects.

reakaleek (approved):

Issue #1826, opened 2025-09-04
[Feature Request]: Remove console title highlighting


lcawl:
### Prerequisites

- [x] I have searched existing issues to ensure this feature hasn't already been requested
- [x] I have tested using the latest version of docs-builder

### What problem are you trying to solve?

The console codeblock support added in https://github.com/elastic/docs-builder/pull/375 "highlights the first line using a new custom highlighter for console code examples where the first line is the HTTP API call".

Per feedback from @drewdaemon, this formatting is unexpected where codeblocks contain multiple API calls.

### Proposed Solution

Since we did not have this kind of title highlighting in the old docs (e.g. https://www.elastic.co/guide/en/elasticsearch/reference/8.19/search-your-data.html) I think it's worth considering its removal.

### Examples and Research

Per @bmorelli25 "there are 1756 console code blocks in the ES repo. 293 are multi-API snippets".

### Additional Context

IMO the more important feature related to Console-specific code blocks that continues to be missing from Docs V3 is https://github.com/elastic/docs-builder/issues/55 / https://github.com/elastic/docs-builder/issues/225

### How important is this feature to you?

Nice to have

reakaleek:
Is fixing the appearance for multiple API calls also desired? Or do we just want to remove it?

PR #1895, opened 2025-09-16
Fix appearance of multiple API calls in console codeblock

reakaleek:
Closes https://github.com/elastic/docs-builder/issues/1826

## Changes

- Identify and render multiple API call headers in `console` codeblocks

## Input

````markdown

```console
PUT my-index-000001
{
  "mappings": {
    "enabled": false <1>
  }
}

PUT my-index-000001/_doc/session_1
{
  "user_id": "kimchy",
  "session_data": {
    "arbitrary_object": {
      "some_array": [ "foo", "bar", { "baz": 2 } ]
    }
  },
  "last_updated": "2015-12-06T18:20:22"
}

GET my-index-000001/_doc/session_1 <2>

GET my-index-000001/_mapping <3>
```
1. The entire mapping is disabled.
2. The document can be retrieved.
3. Checking the mapping reveals that no fields have been added.

````

## Output

image

## How to test
1. Copy the input to a markdown file somewhere in `./docs/**`
2. Run `dotnet run --project src/tooling/docs-builder -- serve`
3. Validate the the output in the page you added the codeblock.


lcawl (approved):
Output looks like a good solution to me!

leemthompo:
++ this is a great quick solution!

I wonder if removing the borders would make it 100% clear that this is one big (copyable) block, might be overthinking this though 

For reference, that's how it was handled in the old docs 
 
Screenshot 2025-09-17 at 10 01 04

reakaleek:
> ++ this is a great quick solution!
> 
> I wonder if removing the borders would make it 100% clear that this is one big (copyable) block, might be overthinking this though (and probably not worth the investment)
> 
> For reference, that's how it was handled in the old docs
> 
> Screenshot 2025-09-17 at 10 01 04

@leemthompo I made the borders darker and dotted to make the segments feel less separated.

I still think there is benefit, readability wise. I also tried removing the borders, but IMO the spacing didn't look right.

PR #1900, opened 2025-09-17
Combine applies_to badges with the same key

reakaleek:


cotti (approved):

florent-leborgne:
Currently testing. This is what I'm seeing right now:
image

I think that in order to be shippable standalone and still make sense (rest of the changes will be made separately), we'd need at least the following tweaks made as part of this same PR:

- Include the ellipsis on the badge in case of multiple states
- Also only in case of multiple states (if preview or beta exist in addition to GA), the badge should include "GA"
- Bold the 1st line of each "state" in the tooltip, and optionally add a separator between each state section

I assume you'd rather keep tooltip information ordering for later but if those are simple to change, I'm also happy to provide input on this here.

@reakaleek Thanks for working on this!

florent-leborgne (approved):
Thanks @reakaleek! Just tested and looks good from what I could see :D

Issue #1689, opened 2025-08-05
[Feature Request]: Create llms.txt


reakaleek:
### Prerequisites

- [x] I have searched existing issues to ensure this feature hasn't already been requested
- [x] I have tested using the latest version of docs-builder

### What problem are you trying to solve?

See https://llmstxt.org/

### Proposed Solution

See https://llmstxt.org/

### How important is this feature to you?

Nice to have

PR #1928, opened 2025-09-22
Add llms.txt template and generator

theletterf:
Fixes https://github.com/elastic/docs-builder/issues/1689

The approach here is to create a template based on https://llmstxt.org/ recommendations, and then append a number of generated H2 sections after the boilerplate. The sections should contain a list, ordered following the same navigation order in the site, of the first level sections of each category, with a Markdown link to their .md files and, after a colon, their description, extracted from the description: frontmatter of those pages. This process should happen after assembly, more or less around the time we create the sitemap.xml. 

For clarity. If, for example, we have `## Solutions` as the category (https://www.elastic.co/docs/solutions), then I would expect the list to contain the first three subsections you see when opening it, and no more. The goal is for an LLM to consume this file and get an overview of the first level of navigation within each section.

> **LLM disclosure: All commits in this PR have been coauthored by Claude using Cursor Enterprise**

### DIFF
--- a/src/Elastic.Markdown/Exporters/LlmMarkdownExporter.cs
+++ b/src/Elastic.Markdown/Exporters/LlmMarkdownExporter.cs
@@ -6,6 +6,7 @@ using System.IO.Abstractions;
 using System.IO.Compression;
 using System.Text;
 using Elastic.Documentation.Configuration;
+using Elastic.Documentation.Configuration.Products;
 using Elastic.Markdown.Helpers;
 using Markdig.Syntax;
 
@@ -147,21 +148,11 @@ public class LlmMarkdownExporter : IMarkdownExporter
 		if (!string.IsNullOrEmpty(sourceFile.Url))
 			_ = metadata.AppendLine($"url: {context.BuildContext.CanonicalBaseUrl?.Scheme}://{context.BuildContext.CanonicalBaseUrl?.Host}{sourceFile.Url}");
 
-		var configProducts = context.BuildContext.ProductsConfiguration.Products.Select(p =>
-		{
-			if (context.BuildContext.ProductsConfiguration.Products.TryGetValue(p.Value.Id, out var product))
-				return product;
-			throw new ArgumentException($"Invalid product id: {p.Value.Id}");
-		});
-		var frontMatterProducts = sourceFile.YamlFrontMatter?.Products ?? [];
-		var allProducts = frontMatterProducts
-			.Union(configProducts)
-			.Distinct()
-			.ToList();
-		if (allProducts.Count > 0)
+		var pageProducts = GetPageProducts(sourceFile.YamlFrontMatter?.Products);
+		if (pageProducts.Count > 0)
 		{
 			_ = metadata.AppendLine("products:");
-			foreach (var item in allProducts.Select(p => p.DisplayName).Order())
+			foreach (var item in pageProducts.Select(p => p.DisplayName).Order())
 				_ = metadata.AppendLine($"  - {item}");
 		}
 
@@ -172,6 +163,9 @@ public class LlmMarkdownExporter : IMarkdownExporter
 
 		return metadata.ToString();
 	}
+
+	private static List<Product> GetPageProducts(IReadOnlyCollection<Product>? frontMatterProducts) =>
+		frontMatterProducts?.ToList() ?? [];
 }
 
 public static class LlmMarkdownExporterExtensions
--- a/build/Targets.fs
+++ b/build/Targets.fs
@@ -134,7 +134,7 @@ let private runTests (testSuite: TestSuite) _ =
     let testFilter =
         match testSuite with
         | All -> []
-        | Unit -> ["--filter"; "FullyQualifiedName~.Tests"]
+        | Unit -> ["--filter"; "FullyQualifiedName~.Tests|FullyQualifiedName~AuthoringTests"]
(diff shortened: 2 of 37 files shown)


src/Elastic.Markdown/Myst/Renderers/LlmMarkdown/LlmBlockRenderers.cs
```diff
@@ -39,19 +39,38 @@ public static void RenderBlockWithIndentation(LlmMarkdownRenderer renderer, Mark
...
+	/// Converts relative URLs to absolute URLs using BuildContext.CanonicalBaseUrl for better LLM consumption.
+	/// For documentation URLs, converts them to .md format for consistency with LLM-generated files.
 	/// </summary>
 	public static string? MakeAbsoluteUrl(LlmMarkdownRenderer renderer, string? url)
+	{
+		if (renderer.BuildContext.CanonicalBaseUrl == null)
+			return url;
+
+		// For documentation URLs (starting with /docs/), convert to .md format for LLM consistency
+		if (!string.IsNullOrEmpty(url) && url.StartsWith("/docs/", StringComparison.Ordinal))
```
reakaleek, line 51: I'm not sure if this is the best way to identify an internal URL.

Let me do some digging on this.

src/Elastic.Markdown/Myst/Renderers/LlmMarkdown/LlmInlineRenderers.cs
```diff
@@ -43,6 +43,43 @@ protected override void Write(LlmMarkdownRenderer renderer, LinkInline obj)
...
+	/// </summary>
+	private static string? ProcessLinkUrl(LlmMarkdownRenderer renderer, LinkInline obj, string? url)
+	{
+		// Check if this is a crosslink (similar to HtmxLinkInlineRenderer)
+		var isCrossLink = (obj.GetData("isCrossLink") as bool?) == true;
+
+		// Also check if this is a resolved crosslink to Elastic docs by URL pattern
+		// This handles cases where crosslinks have been resolved but lost their metadata
+		var isElasticDocsUrl = !string.IsNullOrEmpty(url) &&
+							  url.StartsWith("https://www.elastic.co/docs/", StringComparison.Ordinal);
```
reakaleek, line 58: is the `.StartsWith` correct here?

Shouldn't we compare with `.StartsWitch("/docs/")`?

Otherwise, I think the isCrossLink check is unnecessary, because crosslinks should also start with `https://www.elastic.co/docs/`

reakaleek: The thing is.. crosslinks are rendered as absolute links (e.g. `https://www.elastic.co/docs/..`, because we need an absolute link in isolated builds.

internal links are relative (e.g. `/docs/...`

reakaleek (approved):

theletterf:
@Mpdreamz Do we have our blessing? 🙏🏻

src/Elastic.Markdown/Exporters/LlmMarkdownExporter.cs
```diff
@@ -64,6 +90,17 @@ public static string ConvertToLlmMarkdown(MarkdownDocument document, BuildContex
...
+	{
+		var defaultOutputFile = fileContext.DefaultOutputFile;
+		var fileName = Path.GetFileNameWithoutExtension(defaultOutputFile.Name);
+		if (fileName != "index")
+			return false;
+
+		var root = fileContext.BuildContext.OutputDirectory;
+		return defaultOutputFile.Directory!.FullName == root.FullName;
+	}
+
```
Mpdreamz, line 103:
```suggestion
	private static bool IsRootIndexFile(MarkdownExportFileContext fileContext)
	{
		var defaultOutputFile = fileContext.DefaultOutputFile;
		var fileName = Path.GetFileNameWithoutExtension(defaultOutputFile.Name);
		if (fileName != "index")
			return false;

		var root = fileContext.BuildContext.OutputDirectory;
		return defaultOutputFile.Directory!.FullName == root.FullName;
	}
```

src/Elastic.Markdown/Myst/Renderers/LlmMarkdown/LlmInlineRenderers.cs
```diff
@@ -43,6 +43,8 @@ protected override void Write(LlmMarkdownRenderer renderer, LinkInline obj)
 		}
 		renderer.Writer.Write(")");
 	}
+
+
```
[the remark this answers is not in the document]
theletterf: Fat fingers :) let me fix this

Mpdreamz (approved):
Spot a few cases for improvements but not blocking this PR lets get v1 of llms.txt out the door :) its been too long.

src/Elastic.Markdown/Exporters/LlmMarkdownExporter.cs
theletterf: @Mpdreamz This change seems to base the llms.txt file on the index.html page again instead of using the stringified boilerplate. See example output here: https://gist.github.com/theletterf/ebae8d20bd369aa643fe7c7318b97ad3

We should either revert it or find a way of using the boilerplate in https://github.com/elastic/docs-builder/pull/1928/files#diff-1e28335bf33fdd740a27b988223294330c7e1e510c86a0db70215cd4165a18e7R19-R42

PR #1933, opened 2025-09-23
Include proxy only in DEBUG builds not in publish artifacts

Mpdreamz:


Mpdreamz:
The default configuration for `run` and `watch` is debug so no changes should be necessary:

https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-run

PR #1941, opened 2025-09-24
Applies switch compontent

reakaleek:
Part of https://github.com/elastic/docs-builder/issues/1436

## Changes

Adds a new `applies-switch` and `applies-item` compontent.

⚠️ I intentionally duplicated the tabs logic and did not reuse it so we can easier change the appearance and behaviour without affecting the current tabs.

## Usage

```markdown 
::::{applies-switch}

:::{applies-item} { ece:, ess: }
Content for ECE and ECH
:::

:::{applies-item} serverless:
Content for Serverless
:::

::::
```

## Screenshot

image


reakaleek:
@colleenmcginnis

fixed in https://github.com/elastic/docs-builder/pull/1941/commits/e2cbae5a129b2831dedf60ddfc73f01af7516e27

Mpdreamz (approved):
Heavily in favor of a new directive instead of grafting this onto regular tabs. 

This allows us to improve the UX in the future and ensure the applies to in this directive are visually distinctive. 

I can live with them being not visually identifiable as applies to information in the interim.

Mpdreamz:
I do think when used to distinguish between deployment types it hides an content organization smell and might go against @florent-leborgne simplification initiative. 

By making it a separate control if we need to constrain this to version/availability dimensions only it will be easier to update.

colleenmcginnis:
Thanks @reakaleek! This approach makes sense to me -- it will allow us start standardizing how we structure this kind of information in the source files ahead of the 9.2 release while giving us the flexibility to update what the rendered result looks like when we have dedicated design support.

If merging this can wait until tomorrow, I'd like @florent-leborgne to have a chance to comment on the approach and the name of the directives (`applies-switch` and `applies-item`). I think the names are important to get right at this stage before we start using them across a dozen or more repos. 

>I do think when used to distinguish between deployment types it hides an content organization smell and might go against @florent-leborgne simplification initiative.

Tabs used to distinguish between deployment types are mostly in the content that the admin docs team owns, and I think @shainaraskas has a similar view on whether we should use this in those cases. 🙂

PR #1944, opened 2025-09-25
Avoid crashing update-redirects when CLI output is empty

cotti:
`update-redirects` should verify the output at each stage - an empty string is not parsed as a valid empty JSON response. We should fail the workflow step, but not crash.

### DIFF
--- a/src/services/Elastic.Documentation.Assembler/Deploying/Redirects/AwsCloudFrontKeyValueStoreProxy.cs
+++ b/src/services/Elastic.Documentation.Assembler/Deploying/Redirects/AwsCloudFrontKeyValueStoreProxy.cs
@@ -31,7 +31,10 @@ public class AwsCloudFrontKeyValueStoreProxy(IDiagnosticsCollector collector, IL
 		if (string.IsNullOrEmpty(eTag))
 			return;
 
-		var existingRedirects = ListAllKeys(kvsArn);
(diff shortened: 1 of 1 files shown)


src/services/Elastic.Documentation.Assembler/Deploying/Redirects/AwsCloudFrontKeyValueStoreProxy.cs
```diff
@@ -51,7 +54,14 @@ private string DescribeKeyValueStore(string kvsName)
 		try
 		{
 			var json = CaptureMultiple("aws", "cloudfront", "describe-key-value-store", "--name", kvsName);
```
reakaleek, line 56: Can we have a retry mechanism here? It seems like this is a temporary error that only happens occasionally.

reakaleek: I see.

Kinda unrelated.. but I'm starting to wonder how do we handle paging for describe-key-valiue-store?

cotti: We are now logging as warnings when exceptions occur in an execution attempt loop. They shouldn't be errors since later attempts might be successful.

PR #1949, opened 2025-09-26
Add `:applies_to:` prop to dropdown directive

reakaleek:
## Changes

- Adds the `:applies_to:` prop to the dropdown directive
- Fixes the prop parsing (It didn't handle `:` in the value correctly)
- Add additional tests

## Screenshot

image

### DIFF
--- a/docs/syntax/dropdowns.md
+++ b/docs/syntax/dropdowns.md
@@ -50,3 +50,55 @@ Dropdown content
 ::::
 
 :::::
+
+## With applies_to badge
+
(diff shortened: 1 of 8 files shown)


Mpdreamz:
Would this look better as inline 'real' badges?

reakaleek:
> I love it :)

I also like it more. But I think the original implementation/design is more consistent with the new applies switch. 

Hence, I will keep it for now.. after the admonitions PR, I might create a PR to show how it would like like in all.

Mpdreamz (approved):

PR #1951, opened 2025-09-29
[ci] Use same drive as checkouts for TMP on Windows

Mpdreamz:
Temporary while we wait for https://github.com/parcel-bundler/parcel/pull/10095


reakaleek (approved):

reakaleek (changes requested):
Also needs to be added to release.yml

Mpdreamz:
Good catch updated!

PR #1961, opened 2025-09-29
Add `:applies_to:` prop to admonitions

reakaleek:
## Changes
- Add `:applies_to:` prop handling for adminitions
- Change styling to make it work with applies_to

## Screenshot

image

### DIFF
--- a/docs/syntax/admonitions.md
+++ b/docs/syntax/admonitions.md
@@ -150,26 +150,31 @@ It can *span* multiple lines and supports inline formatting.
 
 :::::
 
-## Collapsible admonitions
+## Applies to information
 
-:::{warning}
-Collapsible admonitions are deprecated. Do not use them. Use [dropdowns](./dropdowns.md) instead.
-:::
-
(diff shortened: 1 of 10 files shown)


Mpdreamz (approved):
I do think the styling of the applies switcher is the exception. 

Can discuss later though not blocking the pr.

colleenmcginnis:
>What would be the expectation for the tooltip or future popover in this case?

I'm not a UX designer, but my gut tells me that making the badge interactive in admonitions is less problematic than with tabs/dropdowns because the admonition itself is not interactive.

florent-leborgne:
^^ What @colleenmcginnis said.
This is something we can rationalize in the future as the cumulative docs initiative will hopefully also lead to output improvements/simplifications in that area.

florent-leborgne (approved):
LGTM, thanks!

shainaraskas:
@colleenmcginnis perhaps good enough for now? 🤷

PR #1965, opened 2025-09-30
Fix running authoring tests in CI

reakaleek:
## Changes
- prefix all authoring tests with `AuthoringTests`.
- Add prefix filter
- Fix failing applies_to tests.


Mpdreamz (approved):
Thank you for remembering and doing this!

PR #1978, opened 2025-10-01
Add more documentation for `docs-builder` CLI and its basic building blocks

Mpdreamz:
- Document all `CLI` commands on `docs-builder`. 
- Document basic building blocks that define `docs-builder`'s architecture.
  - Documentation Set
  - Assembler Builds
  - Navigation, docset.yml, toc.yml, navigation.yml and how they all relate
  - Linking, inbound and outbound link validation
  - Distributed documentaton
  - Link Service, Link Catalog and Link Index files. (cc @elastic/docs-engineering for updated terminology).

### DIFF
--- a/.github/workflows/preview-build.yml
+++ b/.github/workflows/preview-build.yml
@@ -404,6 +404,7 @@ jobs:
             !cancelled() 
             && github.repository != 'elastic/docs-builder'
             && steps.docs-build.outputs.skip != 'true'
+            && steps.docs-build.outcome != 'skipped'
             && (
               steps.deployment.outputs.result 
               || (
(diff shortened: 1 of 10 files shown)


docs/cli/assembler/assembler-bloom-filter-lookup.md
```diff
@@ -0,0 +1,17 @@
+---
```
reakaleek, line 1: This comment is not particularly regarding this file.

But I think we should keep it more high-level. This kind of documentation can go out of sync.

and anyone should be able to lookup the options with `--help`.

IMO we should focus more on use cases rather than reference content.

But this is just my opinion. Not a blocker.

Mpdreamz: I tend to agree with the out of dateness. 

We do need some hard reference docs like this in `docs-builder` so we can hopefully automate its generation in the future and leverage the settings directive as a real test case.

I will merge this as is to keep a record of it and will ask Claude to create a follow up. WDYT?

reakaleek: SGTM

reakaleek (approved):

PR #2014, opened 2025-10-07
Refine image markdown syntax for size attributes

florent-leborgne:
There needs to be a space before the = sign otherwise the specified size is ignored

### DIFF
--- a/docs/syntax/images.md
+++ b/docs/syntax/images.md
@@ -80,8 +80,8 @@ For inline images, the alt text always overrides any title specified in the Mark
 Image sizing is specified through the title argument. You can specify just the size without needing to provide a redundant title:
 
 ```markdown
-![alt](img.png "=WxH")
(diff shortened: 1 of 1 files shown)


docs/syntax/images.md
reakaleek, line 1: IMO, that's a bug

florent-leborgne: I figured you'd say this, but in case you don't get to fix this quickly due to priorities, we can still update these docs so that they correspond to the current behavior? I'll open an issue in parallel

reakaleek: There is always time for small bugs.

Let me try if copilot can fix this.

florent-leborgne: I created https://github.com/elastic/docs-builder/issues/2016

florent-leborgne:
I'll close this since the fix is already underway

PR #2023, opened 2025-10-08
Make Curator 8 a legacy version

shainaraskas:
Curator 9 was just released, so we're working on returning the Curator 8 docs to the asciidoc system (because we don't support major version releases right now)  https://github.com/elastic/docs-content-internal/issues/353 

This PR adds curator 8 to the legacy version array.


colleenmcginnis:
I'm not sure if you're planning to do this separately, but you'll also need to update the [`versions.yml` file](https://github.com/elastic/docs-builder/blob/main/config/versions.yml#L32):

```diff
   curator:
-    base: 8.0
+    base: 9.0
     current: 9.0.0
```

| Before | After |
|---|---|
| Screenshot 2025-10-20 at 5 18 47 PM | Screenshot 2025-10-20 at 5 20 40 PM |

colleenmcginnis (approved):

PR #2060, opened 2025-10-16
Center images when not using 100% width

theletterf:
This PR tweaks our CSS so that images not taking the whole width of the container will be centered, as it happens in the Elastic Blog, for example.

— 

I used Claude Sonnet 4.5 with Claude CLI 2.0 to author this pull request.


theletterf:
@Mpdreamz That's an inline image. I'd argue we don't need / want inline images to be aligned at all, no?

Mpdreamz:
I went crosseyed 
image

The one above is pushed by the content :)

Mpdreamz (approved):

PR #2091, opened 2025-10-22
Add synonyms.yml

cotti:
Adding `synonyms.yml` before the rest of the changeset for synonyms as code, so we can have tests running properly on the branch as files are acquired from `main` .


Mpdreamz (approved):

Mpdreamz:
@shainaraskas we run a lowercase filter in the analyzer so you are correct these need to be lowercased

Mpdreamz (changes requested):
lowercase all terms

cotti:
> do you want more synonyms? thinking of elastic cloud + elasticsearch service being synonyms, for example
> 
> also wondering if capitalization matters here

Good call. These are the ones that were already set up in the UI, but we can always use more as we go.

config/synonyms.yml
```diff
@@ -0,0 +1,24 @@
...
+  - [ "opentelemetry", "otel" ]
+  - [ "eck", "elastic cloud on kubernetes" ]
+  - [ "ece", "elastic cloud enterprise" ]
+  - [ "elv2", "elastic license v2" ]
+  - [ "kql", "kibana query language" ]
+  - [ "ccr", "cross-cluster replication", "cross cluster replication" ]
+  - [ "esaas", "elastic stack as a service" ]
+  - [ "knn", "k-nearest neighbors" ]
+  - [ "ech", "elastic cloud hosted", "ess" ]
+  - [ "aws", "amazon" ]
```
shainaraskas:
```suggestion
  - [ "aws", "amazon" ]
  - [ "gcp", "google cloud platform" ]
```

shainaraskas (commented):
here are some other acronyms for you. I probably have some blind spots but hopefully this is a good start

Mpdreamz (approved):

PR #2103, opened 2025-10-23
Add CF repos to UpdateCLI action

theletterf:
This adds the Cloud Forwarder repos for AWS and Azure to the UpdateCLI configuration, so we can keep those versions updated.


theletterf:
@v1v Would this be the right config?

reakaleek (approved):

.github/updatecli/values.d/scm.yml
```diff
@@ -1,5 +1,8 @@
 scm:
   enabled: true
   owner: elastic
-  repository: docs-builder
+  repositories: |
+    docs-builder
+    edot-cloud-forwarder-aws
+    edot-cloud-forwarder-azure
```
v1v: This is not yet available; let me request IT.

BTW, we will eventually move away from our GH app and use a new approach for managing ephemeral tokens, see https://github.com/elastic/ci-gh-actions/pull/3

v1v (approved):
LGTM, however wait for IT to onboard the new repository

.github/updatecli/values.d/scm.yml
constanca-m: I don't see the [edot-cloud-forwarder-gcp repo](https://github.com/elastic/edot-cloud-forwarder-gcp)

v1v:
it is resolved and actually I can see it's listed as one of the repositories that our GH app ahs been installed.

You likely need to modify the GitHub action create-token to include the repository too

PR #2104, opened 2025-10-23
Applies_to - Mention new supported components in more cumulative docs docs

florent-leborgne:
Didn't want to completely force applies-switch over tabs for now because of ranges, etc. so I went with a tip snippet.

Fixes: https://github.com/elastic/docs-projects/issues/512


colleenmcginnis (approved):
👀

Issue #1660, opened 2025-07-30
[Feature Request]: Allow overrides of "default" versioning scheme


shainaraskas:
### Prerequisites

- [x] I have searched existing issues to ensure this feature hasn't already been requested
- [x] I have tested using the latest version of docs-builder

### What problem are you trying to solve?

In certain areas where docs were not migrated from an initial source, the version dropdown is auto-populating with stack versioning info. 

Examples:
* ECK (should be in the 3.x area): https://www.elastic.co/docs/deploy-manage/deploy/cloud-on-k8s/manage-deployments
* Cloud (no version #s at all): https://www.elastic.co/docs/reference/cloud/cloud-hosted/regions

### Proposed Solution

Some override method

A nice automated way would perhaps be to look at the products or applies_to data - if a `stack` product/tag is not found, fall back to the first listed product for the "current" listing in the dropdown

### Alternative Solutions

throw the dropdown in a fire where it belongs, because it doesn't work for pages with multiple products/version models like [this one](https://www.elastic.co/docs/deploy-manage/security)

### How important is this feature to you?

Critical

PR #2110, opened 2025-10-27
Fix/versioning dropdown

### LINKED ISSUE #2018: EDOT SDK version dropdown not showing the right version
https://github.com/elastic/docs-builder/issues/1637 should have fixed the issue with version dropdowns not showing the right version for certain products, but this still happens. For example:

```
 edot-dotnet:
    display: 'Elastic Distribution of OpenTelemetry .NET'
  edot-java:
    display: 'Elastic Distribution of OpenTelemetry Java'
  edot-node:
    display: 'Elastic Distribution of OpenTelemetry Node'
  edot-php:
    display: 'Elastic Distribution of OpenTelemetry PHP'
  edot-python:
    display: 'Elastic Distribution of OpenTelemetry Python'
```

are showing Stack versions in the dropdowns. For example:

https://www.elastic.co/docs/release-notes/edot/sdks/java

cotti:
Closes #2018, may solve a portion of #1660

Taking into account feedback and study cases provided by @shainaraskas, this PR handles the following:

### Versioning system inference priority refinements

Pages in `docs-content` now rely primarily on frontmatter metadata before legacy page matching.
- If there is at least one direct product reference, use the first one
- If there is a product or deployment applicability set, use the first match, starting from product
- Fallback to legacy page mapping and further existing methods

Question: Should this be applied to all repositories from the get-go, or should we phase this change?

### Dropdown rendering

If a page has a versioning system that doesn't warrant version selection, don't render it.

### DIFF
--- a/docs/_snippets/applies-switch.md
+++ b/docs/_snippets/applies-switch.md
@@ -1,0 +1,11 @@
+::::{applies-switch}
+
+:::{applies-item} stack:
+Content for Stack
+:::
+
+:::{applies-item} serverless:
+Content for Serverless
+:::
+
+::::
--- a/docs/testing/index.md
+++ b/docs/testing/index.md
@@ -99,3 +99,10 @@ const foo = "bar"; <1>
 ```
 
 1. This is a JavaScript code block.
+
+
+:::{include} /_snippets/applies-switch.md
+:::
+
+:::{include} /_snippets/applies-switch.md
+:::
--- a/src/Elastic.Markdown/Myst/Directives/AppliesSwitch/AppliesSwitchBlock.cs
+++ b/src/Elastic.Markdown/Myst/Directives/AppliesSwitch/AppliesSwitchBlock.cs
@@ -22,15 +22,12 @@ public class AppliesSwitchBlock(DirectiveBlockParser parser, ParserContext conte
 
 	private int _index = -1;
 
-	// For simplicity, we use the line number as the index.
-	// It's not ideal, but it's unique.
-	// This is more efficient than finding the root block and then finding the index.
 	public int FindIndex()
 	{
 		if (_index > -1)
 			return _index;
 
-		_index = Line;
+		_index = GetUniqueLineIndex();
 		return _index;
 	}
 }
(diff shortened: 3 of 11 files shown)


src/Elastic.Documentation.Configuration/Versions/VersionInference.cs
```diff
@@ -29,11 +43,59 @@ public VersioningSystem InferVersion(string repositoryName, IReadOnlyCollection<
...
 	}
+
+	private VersioningSystem? VersioningFromApplicability(ApplicableTo applicableTo)
+	{
+		VersioningSystem? versioning = null;
+		if (applicableTo.ProductApplicability is not null)
+		{
+			versioning = applicableTo.ProductApplicability switch
+			{
+				{ ApmAgentAndroid: not null } => VersionsConfiguration.VersioningSystems[VersioningSystemId.ApmAgentAndroid],
```
Mpdreamz: Maybe we should have an exhaustive `ProductFromApplicability` and have a unit test to ensure all known products are handled.

Mpdreamz:
We need these rules to have `all` pages be able default to a versioning scheme/product. We can leverage it more places than just showing the version picker (e.g indexing to ES). 

1. product in frontmatter 
2. product in applicable_to frontmatter
3. product in docset.yml 
4. legacy url mapping (Treated as legacy).
5. docset git repository name.

Rule number 2 is the preferred way to be explicit, we always ask folks to include an `applies_to` section if appropriate.

We need rule number 1 to provide an override but we should keep the tagging at a minimum if `applies_to` already suffices.

We need option 4 for as long as docs-content has pages that would not be covered by rules 1,2,3
We need option 5 when onboarding new documentation sets. Smaller projects need not care about this at all. 

In the case of multiple products:

- 2 > pick the first non `stack` product
- 2 > Error if no unique product (error message says make it explicit by setting product in frontmatter)
- N > Error if no unique product (error message says make it explicit by setting product in frontmatter).


This may be less simplistic to implement but it puts a lot less cognitive overhead for most documentation sets. 

New projects just have to ensure `assembler.yml, products.yml and versioning.yml` is correct and don't need to tag every page with something.

shainaraskas:
@Mpdreamz we might want to chat about this synchronously tomorrow, because a lot of the terms that you're using aren't 100% clear to me (we say product a lot but it might multiple things in our metadata). I also think we differ in opinion about how reliable these data points are, and what to prioritize. I worry that, if I am understanding your order, we'll be breaking more dropdowns than we fix with this change because of how that `products` frontmatter was initially populated.

PR #2111, opened 2025-10-27
Introduce substitutions for previous versions

### LINKED ISSUE #2092: [Feature Request]: Provide a version substitution for latest 8.19.x
### Prerequisites

It would be great to have the latest 8.x stack release (actually 8.19.x) in a version substitution similar to {{version.stack}}.

We use and refer to that version as 8.19.x a lot in the upgrade docs, where recommending users to upgrade to 8.19.latest (even if they are already running 8.19) to get the latest version of the upgrade assistant, when explaning the upgrade paths, etc.

This isn't critical at all but our docs would look and feel better with this.

### Proposed Solution

Introduce a variable to point to the latest version of the previous major and update it after every new release (we will be shipping 8.19.x releases for at least 2 more years I think).

### Examples and Research

_No response_

### Alternative Solutions

_No response_

### Additional Context

_No response_

Nice to have

cotti:
Closes #2092

### DIFF
--- a/src/Elastic.Documentation.Configuration/Builder/ConfigurationFile.cs
+++ b/src/Elastic.Documentation.Configuration/Builder/ConfigurationFile.cs
@@ -4,11 +4,11 @@
 
 using System.IO.Abstractions;
 using DotNet.Globbing;
+using Elastic.Documentation.Configuration.LegacyUrlMappings;
 using Elastic.Documentation.Configuration.Products;
 using Elastic.Documentation.Configuration.Suggestions;
 using Elastic.Documentation.Configuration.TableOfContents;
 using Elastic.Documentation.Configuration.Versions;
-using Elastic.Documentation.Configuration.LegacyUrlMappings;
 using Elastic.Documentation.Links;
 using Elastic.Documentation.Navigation;
 using YamlDotNet.RepresentationModel;


Mpdreamz (changes requested):
I am not in love with with the `previous` moniker.

Could this be a mutation? e.g


```
{{ .elasticsearch | version_legacy }}
```


Ideally we can support version and base also as mutations 


```
{{ .elasticsearch | version }}
{{ .elasticsearch | version_base }} ?
```

shainaraskas:
>I am not in love with with the `previous` moniker.

`legacy` is not ideal either. it's a legacy-ish docs system, but the products are still in active development. while this term is good enough for me, I could see it being confusing for contributors and maybe sending the wrong message for users looking under the hood.

likewise, while I appreciate that this is a straightforward way to make this change, it's not going to be robust long term (i.e. once we hit a new major). the info we actually want here is "the last minor of the previous major", which will break down when we move to stack 10 and 9.x docs are accessible in a way other than traversing mapped pages. I'd rather just see this as another piece of info stored in versions.yml ... we can avoid adding it until it's needed, product by product.

alt: we just store this as a sub in docset.yml. again, we don't need this for all products afaik (maybe @colleenmcginnis can correct me if she sees a big need for this in ingest tool land)

cotti:
I see. Yeah, the need for the patch version means the initial premise here of picking this information from legacy url mappings isn't good enough.

We can add this information to `versions.yml`, is this worth pursuing this change?

shainaraskas:
@cotti I think this is mostly a nice to have feature but not critical. might be worth doing if it's trivially easy

Mpdreamz:
I am closing this, stale and not bubbled back up as a priority. 

@eedugon if this is very much still top of mind, mind creating a ticket of a simplified stack only previous version var? 

There are many ways we can query for this (without having the read elastic/docs even :))

PR #2172, opened 2025-11-06
Fix release

reakaleek:
## Context

The release is failing because it's trying to zip docs-assembler. Which doesn't exist anymore.

## Changes

Remove instructions to zip docs-assembler artifacts.


Mpdreamz (approved):
Eek! I thought i thought i caught em all.

PR #2197, opened 2025-11-12
Add SearchUsecase result logs

reakaleek:


src/api/Elastic.Documentation.Api.Core/Search/SearchUsecase.cs
```diff
@@ -2,32 +2,37 @@
...
-		return new SearchResponse
+		var response = new SearchResponse
 		{
 			Results = results,
 			TotalResults = totalHits,
 			PageNumber = request.PageNumber,
 			PageSize = request.PageSize,
 		};
+
+		logger.LogInformation("Search completed: {search.result.page_size} results for query '{search.query.text}' (Page {search.result.page_number}): {search.result.urls}",
```
Mpdreamz: I am not sure if squeezing search.result.urls in the log message is going to be useful. 

We can make it a log attribute/property only see `[LogProperties]` when using the compile time log emitters.

https://andrewlock.net/customising-the-new-telemetry-logging-source-generator/#adding-properties-to-all-logs-using-enrichment

PR #2207, opened 2025-11-14
Fix OTel histogram metrics

reakaleek:


Mpdreamz (approved):
Cc @stevejgordon this might be something the otel collector could do automatically?

Issue #1838, opened 2025-09-08
[Feature Request]: Add preview hovercards for internal links


theletterf:
### Prerequisites

- [x] I have searched existing issues to ensure this feature hasn't already been requested
- [x] I have tested using the latest version of docs-builder

### What problem are you trying to solve?

GitHub's documentation show hovercards with a preview of internal links when hovering over an internal docs link. The preview looks as follows:

<img width="1062" height="560" alt="Image" src="https://github.com/user-attachments/assets/326f89ce-f9a4-45bb-af7f-3d73f17b5a9a" />

I think we should add this to our docs, too.

### Proposed Solution

We could generate a page-metadata.json file at build that we could then access/read through the hovercard JS code. The JSON file would map links to their metadata.

### Examples and Research

GH seems to use an API for this, with calls to https://docs.github.com/api/article/meta?pathname=<path>

cotti:
## AI triage

_Automated investigation against the current codebase. A human still needs to confirm before any code change._

### Original concern

Add GitHub-style hover preview cards (showing the target page's title/excerpt) when hovering over internal documentation links.

### Findings

- `src/Elastic.Markdown/Myst/Renderers/HtmxLinkInlineRenderer.cs:59-64` — internal/cross links today only get a native browser `title=""` attribute (plain tooltip) when the Markdown link carries a title; there is no preview-card UI anywhere in the link rendering path.
- `src/Elastic.Markdown/Myst/Renderers/HtmxLinkInlineRenderer.cs:50-51` — same-set and cross links already emit `preload=...` (htmx prefetch), which fetches the target HTML on hover/mousedown for fast navigation. This is prefetch-for-speed, not a rendered card, but it confirms a hover hook already exists on links.
- `src/Elastic.Markdown/IO/MarkdownFile.cs:66` and `:76` — every page exposes `Title` and `Description`, so the per-page metadata a card would show **does exist at build time** for same-repo pages.
- `src/Elastic.Markdown/Myst/InlineParsers/DiagnosticLinkInlineParser.cs:323-334` — the link parser already reads `markdown.Title` (and heading text for anchors) to auto-fill link text, proving target title is reachable during the build.
[...]

PR #2222, opened 2025-11-17
[ON week experiment] Add tooltips to the nav

leemthompo:
- vibe-coded pre-ON week testeroo
- probably totally unviable
- there were some refactoring in docs-builder in meantime and I merged main fixing conflicts as much as possibl
- just pushing to see what CI says at this point 😄 

## 📹 

nav-tooltips

### Implementation Summary

Added `NavigationTooltip` property across the navigation system to display hover tooltips for navigation items:

- **Interface Change**: Added `NavigationTooltip` property to `IDocumentationFile` interface
- **Implementation Pattern**:
  - **Leaf nodes**: Delegate to their `Model.NavigationTooltip`
  - **Node items**: Delegate to their `Index.NavigationTooltip`
  - **Root/container items**: Return `null`

### Files Modified

**Core Implementation**:
- `src/Elastic.Markdown/IO/MarkdownFile.cs` - Tooltip logic using description from YAML frontmatter, strips markdown and escapes quotes
- `src/Elastic.Documentation.Navigation/IDocumentationFile.cs` - Added `NavigationTooltip` property to interface

**Navigation Classes** (9 files):
- `src/Elastic.Documentation.Navigation/Assembler/SiteNavigation.cs`
- `src/Elastic.Documentation.Navigation/Isolated/Leaf/CrossLinkNavigationLeaf.cs`
- `src/Elastic.Documentation.Navigation/Isolated/Leaf/FileNavigationLeaf.cs`
- `src/Elastic.Documentation.Navigation/Isolated/Node/DocumentationSetNavigation.cs`
- `src/Elastic.Documentation.Navigation/Isolated/Node/FolderNavigation.cs`
- `src/Elastic.Documentation.Navigation/Isolated/Node/TableOfContentsNavigation.cs`
- `src/Elastic.Documentation.Navigation/Isolated/Node/VirtualFileNavigation.cs`
- `src/Elastic.ApiExplorer/ApiIndexLeafNavigation.cs`

**Frontend**:
- `src/Elastic.Documentation.Site/Assets/navigation-tooltip.ts` - JavaScript implementation
- `src/Elastic.Documentation.Site/Assets/navigation-tooltip.css` - Tooltip styling

**Tests**:
- `tests/Elastic.Markdown.Tests/Inline/ImagePathResolutionTests.cs`
- `tests/Navigation.Tests/TestDocumentationSetContext.cs`

### Integration

Successfully integrated with the refactored navigation system from main branch after merge conflict resolution.

### DIFF
--- a/src/Elastic.Markdown/IO/MarkdownFile.cs
+++ b/src/Elastic.Markdown/IO/MarkdownFile.cs
@@ -339,7 +339,7 @@ public record MarkdownFile : DocumentationFile, ITableOfContentsScope, IDocument
 		var raw = yaml.Lines.Lines is not null
 			? string.Join(Environment.NewLine, yaml.Lines.Lines)
 			: string.Empty;
-		var fm = ReadYamlFrontMatter(raw);
+		var fm = ReadYamlFrontMatter(raw) ?? new YamlFrontMatter();
 
 		if (fm.AppliesTo?.Diagnostics is not null)
 		{


leemthompo:
> Do we need defaults if its not provided?

Good question @Mpdreamz, currently it just does nothing if there isn't a `description` field, but I'm not sure if it's acceptable UX to have some pages not render a tooltip, as we gradually add descriptions.

A related question is that this introduces a new frontmatter field, `navigation_tooltip` which is theoretically more concise than an SEO description, but seeings as it's hard enough to get `description`s everywhere...that might be a step too far. 😄 

> We might want to back this with querying AI descriptions from our search service if not provided?

I mean if you could do that, that would be amazing, but sounds complex 🚀 

One vague ambition with this was that it would be a tangible incentive to adding descriptions to pages :)

theletterf:
@leemthompo Related? https://github.com/elastic/docs-builder/issues/1838

leemthompo:
Thanks for chiming in @itsalexcm! 

> In navigation, users are usually in a goal-oriented mindset. When they hover an item, they’re not trying to “learn” about it, but deciding whether to click or not. In most cases, they’ll just click and expect the destination page to explain the concept.

I'll push back here though, because the whole point of the tooltip is to give some quick visibility into what the page actually contains, to motivate the decision about whether to click or not. If users are "deciding whether to click or not" when hovering, then providing information to help that decision might be useful.

Having to click through multiple pages just to find the right content is frustrating, especially if tooltips could help users navigate more efficiently on the first try.`navigation_title`s aren't always self-explanatory and the cognitive cost of context-switching between pages is real. _Maybe_ tooltips could reduce that friction.

That said, implementation matters: you'd want them to appear on intentional hover (maybe with a slight delay) rather than aggressive instant tooltips that create noise, plus we do need `description`s that are genuinely helpful in this context.

I'm not sure what the "risk" is, feels like a pretty low-stakes enhancement. We could always ship it as an experiment and see if people actually use them. The feedback would have to be pretty qualitative though, because I don't think it would be easy to get clear telemetry data about stuff like hover times and click throughs that wasn't influenced by tons of other factors.

PR #2229, opened 2025-11-17
Fix/synonym tweaks

### LINKED ISSUE #2159: Fix synonyms in search

cotti:
Closes #2159 

This PR adjusts the following parameters for lexical searches:

- `abstract` and `stripped_body` now use the `synonyms_analyzer`;
- A `MatchPhrasePrefixQuery` for synonym-compatible title searches was added to the lexical retriever, at a high priority - but lower than direct matches;
- A custom tokenizer is now provided to handle whitespace and most token-splitting scenarios, while keeping current use cases in `synonyms.yml` intact;
- Document headings have been added to the available data;
- `abstract`, `stripped_body` and `headings` all have explicit AND/OR variants, with the AND versions having a higher boost

### DIFF
--- a/src/Elastic.Markdown/Exporters/Elasticsearch/ElasticsearchExporter.cs
+++ b/src/Elastic.Markdown/Exporters/Elasticsearch/ElasticsearchExporter.cs
@@ -147,23 +147,15 @@ public abstract class ElasticsearchExporter<TChannelOptions, TChannel> : IDispos
 		{
 		  "analysis": {
 		    "analyzer": {
-		      "combined_analyzer": {
-		    	"tokenizer": "whitespace",
-		    	"filter": [
-		    	  "lowercase",
-		    	  "synonyms_filter",
-		    	  "english_stop"
-				]
-			  },
 		      "synonyms_analyzer": {
-		        "tokenizer": "whitespace",
+		        "tokenizer": "group_tokenizer",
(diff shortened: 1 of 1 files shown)


src/Elastic.Markdown/Exporters/Elasticsearch/ElasticsearchExporter.cs
```diff
@@ -243,17 +246,24 @@ protected static string CreateMapping(string? inferenceId) =>
 		      },
 		      "stripped_body": {
 		        "type": "text",
-		        "search_analyzer": "highlight_analyzer",
```
reakaleek: We need highlight_analyzer here so that the highlighted text comes in a certain way

Mpdreamz: We already specify this highlighter during query time: 

https://github.com/elastic/docs-builder/blob/293a9434377c6d7eb7bf1d2fb2c4f5171dcbd90b/src/api/Elastic.Documentation.Api.Infrastructure/Adapters/Search/ElasticsearchGateway.cs#L150

This overrides the one in the mapping. `synonym_analyzer` seems like a good default.

Mpdreamz (approved):

reakaleek (approved):

Issue #2227, opened 2025-11-17
Incorrect previous/next links


colleenmcginnis:
### Describe the bug

I'm seeing strange previous/next links at the end of each page in the reference, release notes, and extend sections. 

### Expected behavior

<img width="1223" height="797" alt="Image" src="https://github.com/user-attachments/assets/91f87f58-7bc2-484b-8fef-d2c2c7975050" />

### Steps to reproduce

1. Go to https://www.elastic.co/docs/reference/elasticsearch/configuration-reference/auding-settings.
2. Scroll to the bottom to look at the previous/next links.

### Tooling

- [x] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

PR #1995, opened 2025-10-03
Navigation Refactor: Generic Type System, O(1) Re-homing, and improved assembly routines

Mpdreamz:
This PR fundamentally rewrites the navigation system to solve long-standing maintainability issues and enable future extensibility.

## The Problem

The original navigation implementation had four critical issues:

1. **Type erasure everywhere** - All navigation returned `INavigationItem<INavigationModel, INavigationItem>`, forcing runtime type checks and casts throughout the codebase to access model-specific properties
2. **Tight coupling to MarkdownFile** - Navigation properties (URLs, paths, breadcrumbs) duplicated on `MarkdownFile` itself, creating tight coupling between navigation logic and a single model implementation tied to HTML rendering
3. **Mixed concerns** - Configuration parsing mixed with navigation construction, making isolated testing impossible
4. **Unclear architecture** - No documentation of core principles, making contributions difficult and bugs easy to introduce

These issues blocked API documentation integration, made the assembler process opaque, and created a brittle system where navigation concerns leaked into model classes.

### 1. Generic Type System with Covariance

**Before:**
```csharp
INodeNavigationItem<INavigationModel, INavigationItem> node = GetNode();
if (node.Model is MarkdownFile markdown)  // Runtime check on every access
{
    var content = markdown.Content;
}
```

**After:**
```csharp
INodeNavigationItem<MarkdownFile, INavigationItem> node = QueryForMarkdownNodes();
var content = node.Model.Content;  // Static typing, zero casts
```

Made all navigation classes generic over `TModel where TModel : IDocumentationFile`. Enables:
- **Static type safety** - Query for specific types without runtime casts
- **No runtime type checks** - Access model properties directly
- **Compile-time errors** - Type mismatches caught during compilation
- **Extensibility** - Easy to add new documentation types (API docs, generated docs)

### 2. Navigation as Single Source of Truth

**Before:** URLs, paths, and navigation properties duplicated on `MarkdownFile`:
```csharp
public class MarkdownFile : IDocumentationFile
{
    public string Url { get; set; }                    // ❌ Duplicated
    public string Path { get; set; }                   // ❌ Duplicated
    public List<Breadcrumb> Breadcrumbs { get; set; }  // ❌ Duplicated
    // Navigation logic leaking into model class
}
```

**After:** Navigation owns all navigational concerns:
```csharp
public class MarkdownFile : IDocumentationFile
{
    // Only model data, no navigation
    public string Content { get; set; }
    public FrontMatter FrontMatter { get; set; }
}

[...]

PR #2231, opened 2025-11-18
Fix Previous/Next and related navigation glitches after refactor

Mpdreamz:
Relates to: https://github.com/elastic/docs-builder/pull/1995

1. Fixes previous/next being broken on the boundaries of documentationsets (https://github.com/elastic/docs-builder/issues/2227)

2. Also fixes a rendering glitch for the index of DocumentationSets in assembler builds see e.g

https://www.elastic.co/docs/reference/ecs:
image

vs its children which do render correctly when navigated to directly:
https://www.elastic.co/docs/reference/ecs/ecs-using-ecs

3. Ensure root items such as https://www.elastic.co/docs/versions render an empty navigation vs rendering the whole thing.

-------
- **Ensure SiteNavigation implements INavigationTraverable**
- **Ensure we get the current navigation in renderlayout from INavigationTraversable not DocumentationSet.**
- **remove NavigationIndexedByCrossLink**
- **SiteNavigation and DocumentationSet share assignments for INavigationTraversable**

### DIFF
--- a/src/Elastic.Documentation.Navigation/INavigationTraversable.cs
+++ b/src/Elastic.Documentation.Navigation/INavigationTraversable.cs
@@ -14,6 +14,8 @@ public interface INavigationTraversable
 
 	IEnumerable<INavigationItem> YieldAll()
 	{
+		if (NavigationIndexedByOrder.Count == 0)
+			yield break;
 		var current = NavigationIndexedByOrder.Values.First();
 		yield return current;
 		do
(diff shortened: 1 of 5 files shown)


src/Elastic.Documentation.Navigation/INavigationTraversable.cs
```diff
@@ -0,0 +1,97 @@
...
+namespace Elastic.Documentation.Navigation;
+
+public interface INavigationTraversable
+{
+	ConditionalWeakTable<IDocumentationFile, INavigationItem> NavigationDocumentationFileLookup { get; }
+	FrozenDictionary<int, INavigationItem> NavigationIndexedByOrder { get; }
+
+	IEnumerable<INavigationItem> YieldAll()
+	{
+		var current = NavigationIndexedByOrder.Values.First();
```
reakaleek: I'm not sure how likely it is. But could `.First()` be `null`?

src/Elastic.Markdown/HtmlWriter.cs
```diff
@@ -59,18 +60,23 @@ private async Task<RenderResult> RenderLayout(MarkdownFile markdown, MarkdownDoc
 	{
 		var html = MarkdownFile.CreateHtml(document);
 		await DocumentationSet.ResolveDirectoryTree(ctx);
-		var navigationItem = DocumentationSet.FindNavigationByMarkdown(markdown);
+		var navigationItem = NavigationTraversable.GetNavigationItem(markdown);
```
reakaleek: Q: Is this already the "current" navigation item?

What's the difference between `NavigationTraversable.GetNavigationItem` and `NavigationTraversable.GetCurrent`

PR #2267, opened 2025-11-25
Redirect fixes

### LINKED ISSUE #2264: Redirects
## Context

We have a fundamental problem with redirects that already occurred a couple of times and we only noticed by accident. There might be a higher number of failed redirects.

## The Problem

When a user deletes or renames a file, the user must add an entry to the redirects.yml file. The key is the path to the deleted/renamed file and the value is the path to the new file.

The redirect is built using this path. **However**, the path might be different in production because it is wired differently through navigation.yml.

## Request

We need a safe and reliable way to set up redirects.

cotti:
Closes #2264 

This PR fixes the following scenario involving redirect sanitization:

### Empty redirect targets

An incorrect input for the redirect target (such as a cross-link not registered in the link index) output an error, but didn't properly register one at the collector. This caused docs-builder to proceed, yielding an empty string as a resolved path.

At a later point in the assemble process, during key-value store updates, the redirect was being set as an empty string.

### DIFF
--- a/src/services/Elastic.Documentation.Assembler/Building/AssemblerBuilder.cs
+++ b/src/services/Elastic.Documentation.Assembler/Building/AssemblerBuilder.cs
@@ -116,7 +116,7 @@ public class AssemblerBuilder(
 			if (Uri.IsWellFormedUriString(path, UriKind.Absolute)) // Cross-repo links
 			{
 				_ = linkResolver.TryResolve(
-					e => context.Collector.EmitError(path, $"An error occurred while resolving cross-link {path}", e),
+					specificErrorMessage => context.Collector.EmitError(path, $"An error occurred while resolving cross-link {path}", specificErrorMessage),
 					new Uri(path),
 					out uri);
 			}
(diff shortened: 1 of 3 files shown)


src/services/Elastic.Documentation.Assembler/Building/AssemblerBuilder.cs
```diff
@@ -116,7 +116,7 @@ string Resolve(string path)
 			if (Uri.IsWellFormedUriString(path, UriKind.Absolute)) // Cross-repo links
 			{
 				_ = linkResolver.TryResolve(
-					e => _logger.LogError("An error occurred while resolving cross-link {Path}: {Error}", path, e),
+					e => context.Collector.EmitError(path, $"An error occurred while resolving cross-link {path}: {e}", null),
```
reakaleek: I 
```suggestion
					e => context.Collector.EmitError(path, $"An error occurred while resolving cross-link {path}", e),
```

cotti: So a new overload for EmitError to support this? (e is a string in this context)

reakaleek (approved):

PR #2270, opened 2025-11-25
Add docker layer caching for docker builds

reakaleek:


Mpdreamz (approved):
Changes look good to me mind giving the PR description as to why these changes are necessary? 

I've read https://docs.docker.com/build/ci/github-actions/cache/#github-cache and understand what it does not super clear on all the benefits. Is this to speed up the build? A lot of time is simply the AOT publish maybe? 


The default max cache size is 10GB so no qualms there.

reakaleek:
> Changes look good to me mind giving the PR description as to why these changes are necessary?
> 
> I've read https://docs.docker.com/build/ci/github-actions/cache/#github-cache and understand what it does not super clear on all the benefits. Is this to speed up the build? A lot of time is simply the AOT publish maybe?
> 
> The default max cache size is 10GB so no qualms there.

It looks like the caching made the build even slower. 😅 

The main reason why I created this PR was because I saw the `dotnet_install` script failing again in a [release workflow](https://github.com/elastic/docs-builder/actions/runs/19664233691/job/56317198219#step:3:963) and I thought this could be cached to make it more resilient

src/api/Elastic.Documentation.Api.Lambda/Dockerfile
```diff
@@ -5,33 +5,37 @@ ARG TARGETOS
...
+    DOTNET_NOLOGO=true \
+    DOTNET_CLI_TELEMETRY_OPTOUT=true
 
+# Install all system dependencies in a single layer
+RUN rpm --import https://packages.microsoft.com/keys/microsoft.asc && \
+    dnf update -y && \
+    dnf install -y npm git clang tar findutils && \
+    dnf clean all
+
+# Install .NET SDK based on global.json version
```
Mpdreamz: I think we should leave in the comment that this might be temporary:

https://github.com/amazonlinux/amazon-linux-2023/issues/1025

Mpdreamz (changes requested):

reakaleek:
Closing, the problem I tried to solve didn't happen since I opened this PR.

PR #2292, opened 2025-12-01
Implement latest search experience design

reakaleek:
## Recording 
https://github.com/user-attachments/assets/c40d2cf4-1ba3-4a78-ba93-a9d930a347e3


Mpdreamz:
@reakaleek @itsalexcm I am merging this in so I can play with our debounce algorithm. That way we can follow up in isolation easier :)

PR #2299, opened 2025-12-02
github-actions: support for vale validation

v1v:
Use https://github.com/elastic/vale-rules using `pull_request_target`.


theletterf:
@v1v @reakaleek We agreed on rolling this out in waves, starting from two/three code repos first. Is this "feature gate" something we can implement here?

reakaleek:
Closing in favour of https://github.com/elastic/docs-builder/pull/2343

PR #2342, opened 2025-12-12
Enable edot-cf-gcp in updatecli

theletterf:
With the first release of edot-cf-gcp happening, we can uncomment the updatecli logic for the repo.

+CC @constanca-m

### DIFF
--- a/.github/updatecli/updatecli.d/versions.yml
+++ b/.github/updatecli/updatecli.d/versions.yml
@@ -73,7 +73,7 @@ sources:
       versionfilter:
         kind: latest
 
-  latest-edot-cf-gcp-version:
+  latest-docker-edot-cf-gcp-version:
     name: Get latest release version for the edot-cf-gcp
     kind: githubrelease
     transformers:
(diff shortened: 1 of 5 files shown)


.github/updatecli/updatecli.d/versions.yml
```diff
@@ -73,18 +73,18 @@ sources:
...
-  #   transformers:
-  #     - trimprefix: v
-  #   spec:
-  #     owner: elastic
-  #     repository: edot-cloud-forwarder-gcp
-  #     token: '{{ requiredEnv "GITHUB_TOKEN" }}'
-  #     username: '{{ requiredEnv "GITHUB_ACTOR" }}'
-  #     versionfilter:
-  #       kind: latest
+  latest-edot-cf-gcp-version:
```
constanca-m: Can we rename this do `latest-docker-edot-cf-gcp-version`?

And add a new one called `latest-terraform-google-edot-cf-version`? That comes from this [repository](https://github.com/elastic/terraform-google-edot-cloud-forwarder).

theletterf: @constanca-m Both versions will always be in sync? If that's the case, I would just fetch one, unless we expect them to diverge.

In any case, let's wait till the other repo is public before merging this PR. I'll now draft the changes for the other repo.

constanca-m: They won't be in sync, they are different things: one is for docker release, go code; and the other is for terraform module.

Both are private at the moment, and ECF GCP repository will always be private

Mpdreamz (approved):

PR #2376, opened 2025-12-17
Register LinksIndexCrossLinkFetcher in DI container and inject properly

reakaleek:
- **Add link tools to MCP server**
- **Register LinksIndexCrossLinkFetcher in DI container and inject properly**


Mpdreamz (approved):
This is the way :)

Then implementations for the tool commands for this also needs to move to a 'Service' like we did for commands, this ensures we can expose these as cli/MCP/http invocations easier. Will comment on the other PR though :)

PR #2380, opened 2025-12-17
Add legal disclaimer and reposition empty chat components

reakaleek:
## Screenshots

image

image


Mpdreamz (approved):
Do we need the disclaimer on the search page too? It's kinda confusing there IMO.

Mpdreamz:
E.g if the disclaimer gets reiterated at the top before the ai replies I think we can remove it from search.

reakaleek:
I will merge it for now. We can revisit this in a later iteration

PR #2374, opened 2025-12-16
Stricter filename validations

Mpdreamz:
## Summary

Enforces stricter filename validation to ensure consistency across documentation files. Filenames must now:
- Be lowercase
- Start with an alphanumeric character or underscore
- Contain only alphanumeric characters, spaces, dashes, dots, and underscores

This PR adds validation for both the full file path and the filename itself, catching issues like uppercase letters in directory names and invalid starting characters.

## Changes

- Added `FilePathRegex` to validate the full path (lowercase, valid characters)
- Added `FileNameRegex` to validate the filename starts with `[a-z0-9_]`
- Combined both validations in `IsValidFileName`
- Extended test coverage with 140 theory-based test cases

## Test plan

- [x] All `OutputDirectoryTests` pass (140 tests)
- [x] Validation correctly rejects uppercase in paths and filenames
- [x] Validation correctly rejects filenames starting with hyphens, dots, or spaces

## Files requiring updates in other repositories

The following files have leading underscores and need to be renamed before we can make the check strictly alphanumeric (removing underscore from allowed starting characters):

### elasticsearch
- `docs/reference/elasticsearch-plugins/_reimplementing_and_extending_the_analyzers.md`
- `docs/reference/elasticsearch-plugins/_other_command_line_parameters.md`
- `docs/reference/elasticsearch-plugins/_plugins_directory.md`
- `docs/reference/elasticsearch-plugins/_reimplementing_and_extending_the_analyzers_2.md`

### integrations
- `docs/extend/_publish_an_integration.md`

### docs-content
- `deploy-manage/tools/cross-cluster-replication/_failback_when_clustera_comes_back.md`
- `deploy-manage/tools/cross-cluster-replication/_failback_when_clustera_comes_back_2.md`
- `deploy-manage/tools/cross-cluster-replication/_prerequisites_14.md`
- `deploy-manage/tools/cross-cluster-replication/_configure_privileges_for_cross_cluster_replication_2.md`
- `deploy-manage/tools/cross-cluster-replication/_perform_update_or_delete_by_query.md`
- `deploy-manage/tools/cross-cluster-replication/_failover_when_clustera_is_down.md`
- `deploy-manage/tools/cross-cluster-replication/_failover_when_clustera_is_down_2.md`
- `deploy-manage/tools/cross-cluster-replication/_connect_to_a_remote_cluster.md`
- `troubleshoot/observability/apm/_collection_of_diagnostic_information.md`
[...]

PR #2384, opened 2025-12-18
Add plus sign filename filter

theletterf:
https://github.com/elastic/docs-builder/pull/2374/files introduced stricter conditions for file names, but forgot to include plus signs, which are used by changelog tooling when creating fragments. See an example here: https://github.com/elastic/elastic-agent/actions/runs/20338076791

This PR adds the plus sign as an allowed character in paths.


Mpdreamz (approved):
LGTM for now, I don't think we should allow plus signs in filenames 

but lets unblock

theletterf:
@Mpdreamz Seems like CodeQL is stuck?

PR #2396, opened 2025-12-22
Add AI-enriched fields

reakaleek:
Adds LLM-powered enrichment to documentation during indexing, generating metadata optimized for semantic search and RAG.

### What it does

Each document gets enriched with:
- **`ai_rag_optimized_summary`** - Dense technical summary for vector search
- **`ai_short_summary`** - 5-10 word tooltip
- **`ai_search_query`** - Keywords a dev would search
- **`ai_questions`** - Questions this doc answers
- **`ai_use_cases`** - Simple tasks like "bulk index documents"

### How enrichment works

**Hybrid approach:**
- **Cache hit:** Enrich processor applies fields at index time (no LLM call)
- **Cache miss (new or stale content):** Call LLM, store in cache, apply fields inline
- **Backfill:** Documents skipped by hash-based upsert get AI fields via `_update_by_query`

```
Document → Generate enrichment_key(title + body) → Check cache (valid entries only) → Hit? Use enrich processor : Call LLM → Store in cache → Apply inline
```

- **Cache index:** `docs-ai-enriched-fields-cache` stores LLM responses with `enrichment_key` and `prompt_hash`
- **Pre-loaded at startup:** Only entries with **current prompt hash** are loaded as valid

### Enrichment key construction

The enrichment key is a **content-only SHA-256 hash** (no prompt hash):

```
input = title + stripped_body
normalized = regex_replace(input, "[^a-zA-Z0-9]", "").ToLowerInvariant()
enrichment_key = SHA256(normalized)
```

**Why content-only keys?**
- Same content always maps to the same cache entry
- Allows stale entries (old prompt) to be overwritten rather than orphaned
- Prompt versioning handled separately via `prompt_hash` field in cache entries

**Why aggressive normalization?**
- Strips all whitespace, punctuation, markdown syntax
- Only keeps alphanumeric characters, lowercased
- **Result:** Formatting changes don't invalidate the cache, only content changes trigger re-enrichment

**Example:**
```
"# Hello, World!"  →  "helloworld"  →  SHA256  →  "a591a6d40..."
"Hello World"      →  "helloworld"  →  SHA256  →  "a591a6d40..."  (same!)
```

### Prompt versioning (automatic cache invalidation)

Each cache entry stores the `prompt_hash` it was generated with. On startup:

1. Load all cache entries with their `prompt_hash` values
2. Compare each entry's `prompt_hash` to current prompt hash
3. **Matching:** Entry is valid → cache hit
4. **Mismatched:** Entry is stale → treated as non-existent → regenerated

Startup:
  Load 5000 entries: 4200 valid (current prompt), 800 stale (will be refreshed)

Document processing:
  if enrichment_key in valid_entries → cache hit
  else → cache miss (new OR stale) → generate + overwrite

**Result:** 
- Prompt changes trigger automatic, gradual re-enrichment
- No manual version bumping needed
- Stale entries are overwritten (not orphaned) since keys are content-based

### Gradual rollout

To avoid deployment explosions, enrichment is **rate-limited**:

| Setting | Value |
|---------|-------|
| Enrichments per run (new + stale) | 100 |
| Concurrent LLM calls | 10 |
| Retry on 429 | Exponential backoff (5 retries) |

**Timeline to full enrichment:**

| Documents | Per run | Runs needed | At hourly deploys |
|-----------|---------|-------------|-------------------|
| ~12,000 | 100 | 120 runs | **~5 days** |

Cache hits don't count against the limit, so once cached, all documents enrich instantly.

### Usage

```bash
docs-builder assembler index --enable-ai-enrichment
```

### DIFF
--- a/src/Elastic.Markdown/Exporters/Elasticsearch/ElasticsearchMarkdownExporter.Export.cs
+++ b/src/Elastic.Markdown/Exporters/Elasticsearch/ElasticsearchMarkdownExporter.Export.cs
@@ -238,7 +238,7 @@ public partial class ElasticsearchMarkdownExporter
 				return;
 
 			// Store in cache for future runs
-			await _enrichmentCache.StoreAsync(doc.EnrichmentKey, enrichment, ctx);
+			await _enrichmentCache.StoreAsync(doc.EnrichmentKey, doc.Url, enrichment, ctx);
 
 			// Apply fields directly (enrich processor won't have this entry yet)
 			doc.AiRagOptimizedSummary = enrichment.RagOptimizedSummary;
--- a/src/Elastic.Markdown/Exporters/Elasticsearch/Enrichment/ElasticsearchEnrichmentCache.cs
+++ b/src/Elastic.Markdown/Exporters/Elasticsearch/Enrichment/ElasticsearchEnrichmentCache.cs
@@ -37,6 +37,7 @@ public sealed class ElasticsearchEnrichmentCache(
 			"mappings": {
 				"properties": {
 				"enrichment_key": { "type": "keyword" },
+				"url": { "type": "keyword" },
 				"ai_rag_optimized_summary": { "type": "text" },
 				"ai_short_summary": { "type": "text" },
 				"ai_search_query": { "type": "text" },
@@ -71,12 +72,13 @@ public sealed class ElasticsearchEnrichmentCache(
 	/// </summary>
 	public bool Exists(string enrichmentKey) => _validEntries.ContainsKey(enrichmentKey);
 
-	public async Task StoreAsync(string enrichmentKey, EnrichmentData data, CancellationToken ct)
+	public async Task StoreAsync(string enrichmentKey, string url, EnrichmentData data, CancellationToken ct)
 	{
 		var promptHash = ElasticsearchLlmClient.PromptHash;
 		var cacheEntry = new CacheIndexEntry
 		{
 			EnrichmentKey = enrichmentKey,
+			Url = url,
 			AiRagOptimizedSummary = data.RagOptimizedSummary,
 			AiShortSummary = data.ShortSummary,
 			AiSearchQuery = data.SearchQuery,
@@ -224,6 +226,12 @@ public sealed record CacheIndexEntry
 {
 	[JsonPropertyName("enrichment_key")]
 	public required string EnrichmentKey { get; init; }
+
+	/// <summary>
+	/// Document URL for debugging - helps identify which document this cache entry belongs to.
+	/// </summary>
+	[JsonPropertyName("url")]
+	public string? Url { get; init; }
 
 	[JsonPropertyName("ai_rag_optimized_summary")]
 	public string? AiRagOptimizedSummary { get; init; }
--- a/src/Elastic.Markdown/Exporters/Elasticsearch/Enrichment/IEnrichmentCache.cs
+++ b/src/Elastic.Markdown/Exporters/Elasticsearch/Enrichment/IEnrichmentCache.cs
@@ -29,7 +29,11 @@ public interface IEnrichmentCache
 	/// <summary>
 	/// Stores enrichment data in the cache.
 	/// </summary>
-	Task StoreAsync(string enrichmentKey, EnrichmentData data, CancellationToken ct);
+	/// <param name="enrichmentKey">The enrichment key (content hash).</param>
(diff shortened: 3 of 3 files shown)


src/Elastic.Markdown/Exporters/Elasticsearch/Enrichment/ElasticsearchEnrichmentCache.cs
```diff
@@ -0,0 +1,248 @@
...
+
+	/// <summary>
+	/// Number of stale entries found during initialization (for logging only).
+	/// </summary>
+	public int StaleCount { get; private set; }
+
+	public async Task InitializeAsync(CancellationToken ct)
+	{
+		await EnsureIndexExistsAsync(ct);
+		await LoadExistingHashesAsync(ct);
```
Mpdreamz: I don't think we should load all hashes in memory just to check if we can enrich 100-200 documents at a time, the llm client can do a DocExists() call.

src/Elastic.Markdown/Exporters/Elasticsearch/ElasticsearchMarkdownExporter.Export.cs
```diff
@@ -191,4 +204,53 @@ public async ValueTask<bool> FinishExportAsync(IDirectoryInfo outputFolder, Canc
...
+		if (_enrichmentCache.Exists(doc.EnrichmentKey))
+		{
+			// Cache hit - enrich processor will apply fields at index time
+			_ = Interlocked.Increment(ref _cacheHitCount);
+			return;
+		}
+
+		// Check if we've hit the limit for enrichments
+		var current = Interlocked.Increment(ref _enrichmentCount);
+		if (current > _enrichmentOptions.MaxNewEnrichmentsPerRun)
```
Mpdreamz: Move this check above the Exists() check (especially if it does IO to do an exist check).

Mpdreamz (changes requested):
A few more nitpicks :)

It might make sense to have a short lived `EnrichmentChannel` to buffer those `200` enrichments and have them `_bulk` up. Especially if we decide to up that number. 

Or build a `enrichment only` index mode that we can run as an offline/one off process that can run for many hours.

src/Elastic.Markdown/Exporters/Elasticsearch/ElasticsearchMarkdownExporter.cs
```diff
@@ -329,6 +387,83 @@ public async ValueTask StopAsync(Cancel ctx = default)
...
+
+			// Backfill: Apply AI fields to documents that were skipped by hash-based upsert
+			await BackfillMissingAiFieldsAsync(ctx);
+		}
+	}
+
+	private async ValueTask BackfillMissingAiFieldsAsync(Cancel ctx)
+	{
+		// Why backfill is needed:
+		// The exporter uses hash-based upsert - unchanged documents are skipped during indexing.
```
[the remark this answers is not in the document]
reakaleek: https://github.com/elastic/docs-builder/pull/2396/commits/9f7ddbba68b9f5930867b2515d470a233effecad

src/Elastic.Markdown/Exporters/Elasticsearch/Enrichment/ElasticsearchEnrichmentCache.cs
```diff
@@ -0,0 +1,248 @@
...
+		}
+		return (total, stale, scrollId);
+	}
+}
+
+/// <summary>
+/// Document structure for the enrichment cache index.
+/// Fields are stored directly for use with the enrich processor.
+/// </summary>
+public sealed record CacheIndexEntry
```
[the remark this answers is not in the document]
reakaleek: https://github.com/elastic/docs-builder/pull/2396/commits/c61fcdb279b18b12a6254a689fb28da40431c38d

src/Elastic.Markdown/Exporters/Elasticsearch/ElasticsearchMarkdownExporter.Export.cs
reakaleek: The order is intentional. `Exists()` is an in-memory dictionary lookup (as of now). Cache hits don't call the LLM, so they shouldn't count against the limit. If we checked the limit first, we'd block documents that already have cached enrichments. The limit caps LLM calls, not total enrichments.

Mpdreamz (approved):

PR #2441, opened 2026-01-09
Isolate /search to /navigation-search and simplify our webcomponents

Mpdreamz:
## Summary

Separates NavigationSearch and AskAi into independent web components to prepare for adding a real full-text search feature later.

The existing `/search` endpoint was an autocomplete for navigating to pages, not actual search. Renaming it to `/navigation-search` clarifies its purpose and reserves `/search` for future full-text search functionality.

Splitting the components also improves code organization - NavigationSearch and AskAi have different responsibilities and lifecycles, so coupling them together added unnecessary complexity.

## Changes

- Rename API endpoint from `/search` to `/navigation-search`
- Split `<search-or-ask-ai>` into two independent web components:
  - `<navigation-search>` - autocomplete for page navigation
  - `<ask-ai>` - AI assistant modal (Cmd+;)
- Move shared utilities (cooldown, error handling) to `web-components/shared/`

## Test plan

- [x] Navigation search autocomplete works
- [x] Ask AI modal opens with Cmd+; and functions correctly
- [x] Both components hide when API is unavailable

### DIFF
--- a/config/versions.yml
+++ b/config/versions.yml
@@ -3,7 +3,7 @@ versioning_systems:
   # Updates for Stack versions are manual
   stack: &stack
     base: 9.0
-    current: 9.2.3
+    current: 9.2.4
 
   # Using an unlikely high version
   # So that our logic that would display "planned" doesn't trigger
@@ -73,7 +73,7 @@ versioning_systems:
   # EDOTs
   edot-collector:
     base: 9.0
-    current: 9.2.3
+    current: 9.2.4
   edot-ios:
     base: 1.0
     current: 1.4.0
@@ -82,7 +82,7 @@ versioning_systems:
     current: 1.4.0
   edot-dotnet:
     base: 1.0
-    current: 1.2.0
+    current: 1.2.2
   edot-java:
     base: 1.0
     current: 1.8.0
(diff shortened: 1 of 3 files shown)


src/api/Elastic.Documentation.Api.Core/Search/SearchUsecase.cs
```diff
@@ -6,81 +6,82 @@
 
 namespace Elastic.Documentation.Api.Core.Search;
 
-public partial class SearchUsecase(ISearchGateway searchGateway, ILogger<SearchUsecase> logger)
+// note still called SearchUseCase because we'll re-add Search() and ensure both share the same client.
+public partial class SearchUsecase(IFindPageGateway findPageGateway, ILogger<SearchUsecase> logger)
 {
-	public async Task<SearchApiResponse> Search(SearchApiRequest request, Cancel ctx = default)
+	public async Task<FindPageApiResponse> FindPageAsync(FindPageApiRequest request, Cancel ctx = default)
```
reakaleek: Are we sure about fully separating search and find pages?

If we plan to have a "View more results" button within the dropdown that leads to the search view, then I would expect the same results in the search view. And I think this would be the most consistent if the find pages feature is basically using the search endpoint, but maybe with some predefined query params.

Mpdreamz: They will be different, the auto complete on the full page search will be for terms or semantic query phrases.  

The search backend implementation will share most of the search (e.g relevance) but I think it's best to give each endpoints their own dto's.

reakaleek (approved):

PR #2454, opened 2026-01-13
Support mutation M.M-1

eedugon:
We need this functionality to keep track of the versions supported by default on Elastic Cloud Hosted, which are:

* The two latest minor versions of the latest major version
* The latest minor version of the previous major version

This PR focuses on the first item.

I'd like {{version.stack | M.M-1 | M.M }} to render to 9.1 if version.stack is for example 9.2.3.

If minor version is 0 we return the original value without deducing anything, to avoid a negative number.

PS - I've used cursor to implement this.

This functionality would be required by https://github.com/elastic/docs-content/pull/4607

### DIFF
--- a/docs/syntax/substitutions.md
+++ b/docs/syntax/substitutions.md
@@ -77,6 +77,7 @@ For variables declaring a semantic version or `Major.Minor` the following operat
 | `M.x`    | Display major component followed by '.x' |
 | `M.M`    | Display only the major and the minor     |
 | `M+1`    | The next major version                   |
+| `M-1`    | The previous major version (returns original value if major is 0) |
 | `M.M+1`  | The next minor version                   |
 | `M.M-1`  | The previous minor version (returns original value if minor is 0) |
 
@@ -108,6 +109,7 @@ sub:
 * M.M: {{version.stack | M.M }}
 * M: {{version.stack | M }}
 * M+1: {{version.stack | M+1 }}
+* M-1: {{version.stack | M-1 }}
 * M+1 | M.M: {{version.stack | M+1 | M.M }}
 * M.M+1: {{version.stack | M.M+1 }}
 * M.M-1: {{version.stack | M.M-1 }}
@@ -130,6 +132,7 @@ sub:
 * M.M: {{version.stack | M.M }}
 * M: {{version.stack | M }}
 * M+1: {{version.stack | M+1 }}
+* M-1: {{version.stack | M-1 }}
 * M+1 | M.M: {{version.stack | M+1 | M.M }}
 * M.M+1: {{version.stack | M.M+1 }}
 * M.M-1: {{version.stack | M.M-1 }}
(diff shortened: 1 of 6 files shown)


Mpdreamz (changes requested):

reakaleek:
> I agree it's better to anchor it to 0 and dont return an error. 

In the worst case this will show an incorrect version and we won't even be aware of it.

eedugon:
> I agree it's better to anchor it to 0 and dont return an error.

@Mpdreamz , do you mean to normalize it to 0 in a similar way than the +1 operations?

So, instead of returning the exact original value:

- `9.0.3` --> M.M-1 --> `9.0.0` --> Is that what you propose? I'm totally ok with that too, as it will be a bit more aligned with M+1 that always returns .0s after the affected item.

In such case I'd do M-1 in a similar way:

- `0.2.3` --> M-1 --> `0.0.0` (something that shouldn't be used unless your system has released 0.x and 1.x versions already).

Did you mean this approach or do you prefer to return the "original value" in these special cases?

reakaleek:
> > The two latest minor versions of the latest major version
> 
> ```
> Supported range is
> * {{version.stack | M.M-1 | M.M-1 | M.M }} to {{version.stack | M.M }}
> ```
> 
> This would render
> 
> version.stack	result
> 9.0	9.0 to 9.0
> 9.1	9.0 to 9.1
> 9.2	9.0 to 9.2
> 9.3	9.1 to 9.3
> Is that an error or acceptable? I am not sure its an error @reakaleek, wether its acceptable I'll defer to the writers @eedugon @elastic/docs-tech-leads
> 
> > The latest minor version of the previous major version
> 
> This relates to: #2111
> 
> @eedugon can you discuss this with @shainaraskas ?

What would happen if we were at `10.0`?

Mpdreamz:
version.stack | result
-- | --
10.0 | 10.0 to 10.0

eedugon:
@reakaleek , the behavior is documented

image

Do you mean in the code? THere's only one line with the payload and at the moment it's pretty simple:

```
SubstitutionMutation.DecreaseMinor => TryGetVersion(value, v => v.Minor == 0 ? value : $"{v.Major}.{v.Minor - 1}.0"),
```

We need to clarify if it's ok to return the original value when `-1` does not exist, or if we should normalize the original value with `.0s` also. For example for `9.0.1`, `M-M-1` could return `9.0.0` instead of the original value.

Whatever you prefer (cc: @Mpdreamz )

shainaraskas:
Sorry to come into this late - I am having trouble making heads or tails of what's happening here. however, if my understanding is correct, I think we should:

* backstop to 0 for the minor and patch segments
* error on 0.0 (major = 0 minor = 0) - if it never makes it into the codebase as 0.0, it shouldn't silently fail later.

this is going to be a _very infrequently_ used variable so I don't think we need to consider the edge cases super thoroughly, though.

Mpdreamz:
@elastic/docs-tech-leads what did we settle on here? 

My $0.02 is having sometimes ranges display as `10.0 and 10.0` is not a big issue.

Mpdreamz:
Closing this! Been open for a while and just took a quick poll with @elastic/docs-tech-leads . Lets see if the need emerges again!

PR #2464, opened 2026-01-14
equal and plus disambiguation in applies.md

florent-leborgne:
I think we can be slightly clearer with the difference between = and +


yetanothertw (approved):
👍

PR #2470, opened 2026-01-14
Fix HTMX navigation from "Find Pages" dropdown

reakaleek:
## Context 
Clicking search results caused htmx:oobErrorNoTarget errors on slow networks. The error happened because the search popover was unmounting before htmx could complete its request.

## Root cause
When the popover closed, the anchor element was removed from the DOM. htmx needs this element to exist during the entire request lifecycle.

## Solution
Keep the popover mounted in the DOM, but control visibility with CSS.
`isOpen={hasContent}` → Popover stays in DOM while there is search text
CSS (`visibility`, `opacity`, `pointer-events`) → Controls show/hide

On click: Hide popover instantly with CSS (element stays in DOM)
After htmx completes: Blur the input field

## Result
Instant visual feedback when user clicks a result
htmx works reliably on all network speeds
Clean user experience

### DIFF
--- a/src/Elastic.Documentation.Site/Assets/web-components/NavigationSearch/NavigationSearch.tsx
+++ b/src/Elastic.Documentation.Site/Assets/web-components/NavigationSearch/NavigationSearch.tsx
@@ -84,22 +84,12 @@ export const NavigationSearch = () => {
         inputRef.current?.select()
     })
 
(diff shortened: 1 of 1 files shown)


reakaleek:
Re-requesting reviews because I changed the code significantly since the last review.

I was able to reproduce the error locally with a throttled network speed in chrome.

See more details in the PR description.

PR #2500, opened 2026-01-19
Add product filtering to full-page search with enhanced active filter display

Mpdreamz:
- Add product and related_products fields to search index and API
- Create ProductFilter facet in sidebar with top 5 by count + alphabetical "Others"
- Display product badges on result cards (primary product highlighted, related outlined)
- Show active filters with facet type labels, icons, and remove buttons
- Add product aggregation on related_products.id with AND behavior for multi-select
- Include product display name lookup via productsConfig.ts

Also adds DocumentInference for version/deployment detection from applies_to metadata
with comprehensive test coverage.

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

https://github.com/user-attachments/assets/f2a5f697-f333-4400-8293-47fbfe39b792

### DIFF
--- a/src/Elastic.Markdown/Exporters/Elasticsearch/ElasticsearchMarkdownExporter.Export.cs
+++ b/src/Elastic.Markdown/Exporters/Elasticsearch/ElasticsearchMarkdownExporter.Export.cs
@@ -5,13 +5,10 @@
 using System.IO.Abstractions;
 using Elastic.ApiExplorer.Elasticsearch;
 using Elastic.Documentation.AppliesTo;
-using Elastic.Documentation.Configuration.Products;
-using Elastic.Documentation.Configuration;
 using Elastic.Documentation.Navigation;
 using Elastic.Documentation.Search;
 using Elastic.Ingest.Elasticsearch.Indices;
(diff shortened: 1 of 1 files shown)


reakaleek (approved):

shainaraskas:
@Mpdreamz 

> Version is complex, the version picker is just docs system for now, we can grow from there :)
> Doing the simplest thing first.

the issue is that "9.0" is not representative of v3. elastic stack 9.0 is one thing, but we have all sorts of versioned docs for other products. for example:

- how do I know I need to select 9.0 to get to the ECK 2.16 docs? 
- if I'm running stack 8.19 [old system] on ECK 3.0 [new system], what do I select?

at minimum, we need labels that tell the truth of this.

> What does 'versionless' mean? e.g search all versions at once? Thats explicitly not supported the use cases should be rare and we can leverage other mechanisms UX wise.

serverless elasticsearch is not == stack 9.0, but as a serverless user, there is no "version" I am using, so again I wouldn't know what to select (even though you and I know that the answer is "9.0"). same goes for elastic cloud hosted, which is versionless (albeit supports the versioned stack).

PR #2508, opened 2026-01-20
Use IFileSystem for ChangelogService and update tests

Mpdreamz:
Split ChangelogServiceTests into many isolated test classes grouped by the functionality they are testing

cc @lcawl


cotti (approved):
🙏

PR #2522, opened 2026-01-21
Fix ChangelogConfiguration.cs links

lcawl:
This PR fixes some broken links in https://docs-v3-preview.elastic.dev/elastic/docs-builder/tree/main/contribute/changelog to:

- https://github.com/elastic/docs-builder/blob/main/src/services/Elastic.Changelog/Configuration/ChangelogConfiguration.cs
- https://github.com/elastic/docs-builder/blob/main/config/changelog.example.yml

I also updated the changelog configuration example to put `add_blockers` before `render_blockers` since that was the original intention due to the order of actions.


Mpdreamz (approved):
LGTM!

PR #2523, opened 2026-01-21
Restructure changelog configuration to use pivot section with inverted label mappings

Mpdreamz:
## Summary

Restructures changelog configuration to use a new `pivot` section with inverted label mappings, providing a more intuitive configuration format where values map to their labels (rather than labels mapping to values).

### Key Changes

- **New `pivot` configuration structure** - Groups types, subtypes, and areas with their associated GitHub labels
- **Flexible TypeEntry format** - Supports three forms:
  - Simple string: `bug-fix: ">bug"` (labels as comma-separated string)
  - Empty/null: `feature:` (no labels)
  - Object: `{ labels: "...", subtypes: {...} }` (with type-specific subtypes)
- **Computed mappings** - `LabelToType` and `LabelToAreas` are now derived by inverting the pivot entries
- **Derived available lists** - `AvailableTypes`, `AvailableSubtypes`, `AvailableAreas` computed from pivot keys

### New Configuration Format

```yaml
pivot:
  types:
    breaking-change: ">breaking, >bc"    # labels as comma-separated string
    bug-fix: ">bug"
    feature:                             # empty = no labels
    deprecation:                         # complex object form
      labels: ">deprecation"
      subtypes:
        api: ">api-deprecation"
  subtypes:
    api:
    behavioral:
  areas:
    Search: ":Search/Search"
    Security: ":Security/Security"
```

### Breaking Change

This is a clean break from the old format. Existing `changelog.yml` files using `available_types`, `available_subtypes`, `label_to_type`, and `label_to_areas` will need to be migrated to the new `pivot` structure.

## Test plan

- [x] All 118 existing changelog tests pass
- [x] New tests added for pivot configuration parsing
- [x] New tests for LabelToType and LabelToAreas mapping computation
- [x] Build succeeds with no errors

cc @lcawl this would be breaking for existing examples, need to be migrated with AI.


cotti (approved):

config/changelog.example.yml
```diff
@@ -1,46 +1,65 @@
...
+    enhancement:
+    docs:
+    known-issue:
+    other:
+    regression:
+    security:
+    # Complex object form with type-specific subtypes
+    deprecation:
+      labels: ">deprecation"
+      subtypes:
```
lcawl: >      subtypes:

We don't currently allow people to make up their own subtypes and we don't have any subtypes for the deprecation type, so I think this should not be used in the example. Rather, I think we should show the breaking change subtypes.

config/changelog.example.yml
```diff
@@ -1,46 +1,65 @@
...
-  - behavioral
-  - configuration
-  - dependency
-  - subscription
-  - plugin
-  - security
-  - other
+  # Default subtype definitions (used when type doesn't define its own subtypes)
+  # Keys are subtype names, values are label strings (or null for no labels)
+  subtypes:
```
lcawl: >  subtypes:

We don't currently allow people to make up their own types or subtypes and breaking changes is the only type that currently has subtypes, so unless we want to decentralize that decision-making, I think we should hide or remove this "default subtype" declaration.

config/changelog.example.yml
```diff
@@ -1,46 +1,65 @@
...
+    security:
+    other:
+
+  # Area definitions with labels
+  # Keys are area display names, values are label strings
+  # Multiple labels can be comma-separated
+  areas:
+    # Example mappings - customize based on your label naming conventions
+    # Autoscaling: ":Distributed Coordination/Autoscaling"
+    # Search: ":Search/Search"
```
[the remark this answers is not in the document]
Mpdreamz: Yep! its an N-N mapping. 

Labels can be comma, separated to have multiple label triggers (an example exists in the config already). 

A label can also appear under multiple areas to set multiple with a single label

config/changelog.example.yml
Mpdreamz: Updated it to ensure types/subtypes is a fixed list, and subtypes may only appear under breaking changes. Updated the example config to reflect this.

PR #2532, opened 2026-01-22
Add Markdown support to the CSV directive

theletterf:
This adds Markdown rendering support to the CSV include directive.

Some notes:

- `NormalizeCsvCellMarkdown + IsPlainUrl`: These exist because the main `MarkdownParser.Pipeline` doesn't include `UseAutoLinks()`. If we added `.UseAutoLinks()` to the pipeline, bare URLs would automatically become links and these two methods could be deleted. However, that's a global change that might affect other content unexpectedly.
- RenderInlineMarkdown: This avoids wrapping single-paragraph content in `<p>` tags inside table cells. Without it, `**Bold**` would render as `<td><p><strong>Bold</strong></p></td>` instead of `<td><strong>Bold</strong></td>`.

The most impactful simplification would be adding UseAutoLinks() to the main pipeline, which would eliminate NormalizeCsvCellMarkdown and IsPlainUrl entirely. The inline rendering optimization is CSV/table-specific.

**Update**
Removing autolinks as an exception from the PR so I can open a separate PR just for autolinks.
—— 

LLM disclosure: I've used GPT Codex 5.2 and Claude Opus 4.5 in Cursor.

### DIFF
--- a/src/Elastic.Documentation.Configuration/Builder/ConfigurationFile.cs
+++ b/src/Elastic.Documentation.Configuration/Builder/ConfigurationFile.cs
@@ -33,7 +33,7 @@ public record ConfigurationFile
 
 	public Dictionary<string, LinkRedirect>? Redirects { get; }
 
-	public HashSet<Product> Products { get; } = [];
+	public HashSet<Product> Products { get; private set; } = [];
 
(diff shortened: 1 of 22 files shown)


shainaraskas:
>poll all docs tech leads

I wish we had it because mostly not having it introduces bugs for us (people putting in plain links without us noticing)

cc @florent-leborgne + @szabosteve

PR #2545, opened 2026-01-23
Change button text from 'Elastic Fundamentals' to 'Get started'

KOTungseth:
User signals for this change:
- From 25-9-25 to 21-1-26, the `Elastic fundamentals` was clicked  13,559 times.
- From 24-5-25 to 24-9-25, the `Get started` CTA was clicked 82,253 times.


theletterf (changes requested):
See comment.

theletterf (approved):
I think we can do this :)

PR #2586, opened 2026-01-29
Add Mermaid diagrams using Beautiful Mermaid

theletterf:
This adds Mermaid diagram rendering through fenced code blocks. The diagrams are rendered through Beautiful Mermaid in lazy-loading mode. This works in air-gapped mode, too, since the minified library is downloaded only when building the binary.

### DIFF
--- a/src/Elastic.Documentation.Site/Assets/mermaid.ts
+++ b/src/Elastic.Documentation.Site/Assets/mermaid.ts
@@ -1,13 +1,81 @@
-// Mermaid.js is loaded from local _static/ to avoid client-side CDN calls
+// Beautiful Mermaid is loaded from local _static/ to avoid client-side CDN calls
 // The file is copied from node_modules during build (see package.json copy:mermaid)
-declare const mermaid: {
-    initialize: (config: object) => void
-    render: (id: string, text: string) => Promise<{ svg: string }>
+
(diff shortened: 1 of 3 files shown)


src/Elastic.Documentation.Site/Assets/mermaid.ts
```diff
@@ -0,0 +1,90 @@
...
+let diagramCounter = 0
+
+/**
+ * Load Mermaid.js from CDN
+ */
+async function loadMermaid(): Promise<void> {
+    if (mermaidLoaded) return
+    if (mermaidLoading) return mermaidLoading
+
+    mermaidLoading = new Promise((resolve, reject) => {
```
reakaleek: I think we should install mermaid through NPM and load it async (for lazyloading) only when a mermaid directive exists

reakaleek:
Wondering if we should center mermaid diagrams by default.

```css
.mermaid-rendered {
  display: flex
  justify-content: center
}
```

src/Elastic.Documentation.Site/Assets/mermaid.ts
theletterf: You mean installing / baking it into the binary? I'll try that approach, I like the idea of loading only when needed.

PR #2611, opened 2026-02-03
Add EDOT CF Azure to nav

theletterf:
Dependent on PR that will add the docs to the Azure repo (PR number WiP).


Mpdreamz (approved):
works locally!

PR #2620, opened 2026-02-03
Expand exception handling in CLAUDE.md

reakaleek:


Mpdreamz (changes requested):
I really do not like this codeql rule or the one that forces filtering with where in the loop statements. 

Sometimes an early continue inside the loop is much saner. 

Generic catch exceptions are fine for the majority of code we write. 

Sadly we can not control codeql rules

PR #2687, opened 2026-02-11
Add `mcp` subcommand to docs-builder CLI

theletterf:
## Summary

- Adds the missing `mcp` subcommand to docs-builder so that `docs-builder mcp` starts the MCP server over stdio, matching the [existing documentation](https://github.com/elastic/docs-builder/blob/main/docs/mcp/index.md).
- Previously, running `docs-builder mcp` would crash because ConsoleAppFramework didn't recognize it as a valid command, writing colored exception output to stdout and breaking the MCP JSON-RPC protocol.
- The fix intercepts `mcp` early in `Program.cs` before ConsoleAppFramework initializes, booting a dedicated `Host` with all logging directed to stderr. MCP tool definitions (`LinkTools.cs`, `Responses.cs`) are shared from the existing `Elastic.Documentation.Mcp` project via linked files to avoid duplication.

### DIFF
--- a/src/tooling/docs-builder/Commands/McpCommand.cs
+++ b/src/tooling/docs-builder/Commands/McpCommand.cs
@@ -4,9 +4,9 @@
 
 using ConsoleAppFramework;
 using Elastic.Documentation.Assembler.Links;
+using Elastic.Documentation.Assembler.Mcp;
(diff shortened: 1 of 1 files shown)


Mpdreamz (approved):
Left a small fix still todo but other than that LGTM

PR #2717, opened 2026-02-17
Fix errors and only disable non `/docs` links for assembler builds

reakaleek:


reakaleek:
Force merging because it's urgent

PR #2736, opened 2026-02-18
Add Vector Sizing Calculator

john-wagster:
Add the React + EUI web component, directive examples, and integration docs under tools/vector_sizing_calculator to stage migration into docs-builder.


reakaleek (changes requested):
What files are in tools folder? Are they still needed?

Also the *.cs files should probably be in https://github.com/elastic/docs-builder/tree/main/src/Elastic.Markdown/Myst/Directives

pmpailis:

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
```diff
@@ -0,0 +1,617 @@
...
+
+const DIMENSION_COMBO_OPTIONS: EuiComboBoxOptionOption<string>[] =
+    DIMENSION_PRESETS.map((n) => ({
+        label: String(n),
+        value: String(n),
+    }))
+
+const TOOLTIPS = {
+    vectors:
+        'Count every vector you will store in the index, not just documents. One document can produce multiple vectors. For example, one embedding per product image plus one for the description.',
```
KOTungseth:
```suggestion
        'The total number of vectors you plan to store in this index, not the number of documents. A single document can produce several vectors, such as one embedding for each product image and one for the description.',
```

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
```diff
@@ -0,0 +1,617 @@
...
+    dimensions:
+        'The length of each vector, set by your embedding model (for example 768 or 1536).',
```
KOTungseth:
```suggestion
        'The number of values in each vector, set by your embedding model.',
```

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
```diff
@@ -0,0 +1,617 @@
...
+    elementType:
+        'The numeric format used to store each dimension. In most cases use float32 and quantization to reduce memory.',
```
KOTungseth:
```suggestion
        'The numeric format used to store each value in a vector. Use float32 in most cases, since Elasticsearch quantizes float vectors by default to reduce memory.',
```

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
```diff
@@ -0,0 +1,617 @@
...
+        label: String(n),
+        value: String(n),
+    }))
+
+const TOOLTIPS = {
+    vectors:
+        'Count every vector you will store in the index, not just documents. One document can produce multiple vectors. For example, one embedding per product image plus one for the description.',
+    dimensions:
+        'The length of each vector, set by your embedding model (for example 768 or 1536).',
+    elementType:
```
KOTungseth, line 65: Should this be `Element type` or `Numeric format`?

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
```diff
@@ -0,0 +1,617 @@
...
+        'The numeric format used to store each dimension. In most cases use float32 and quantization to reduce memory.',
+    indexStructure:
+        'HNSW builds a graph for fast approximate search. Flat is exact brute-force. DiskBBQ stores vectors on disk.',
```
KOTungseth:
```suggestion
        'How vectors are indexed, which drives memory use and search speed. HNSW keeps a graph in memory for fast approximate search. Flat compares every vector for exact results, best at small scale. DiskBBQ reads compressed clusters from disk and needs far less memory than HNSW.',
```

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
```diff
@@ -0,0 +1,617 @@
...
+    graphConnections:
+        'How many neighbors each node connects to in the HNSW graph. Higher values improve recall but increase memory and build time.',
```
KOTungseth:
```suggestion
        'How many neighbors each node connects to in the HNSW graph. Higher values improve recall, but increase memory and build time.',
```

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
```diff
@@ -0,0 +1,617 @@
...
+    quantization:
+        'Compresses vectors to reduce memory. Options depend on element type and index structure.',
```
KOTungseth:
```suggestion
        'Compresses vectors to reduce memory, at some cost to accuracy. Applies to the float and bfloat16 element types. Available options depend on your index structure.',
```

KOTungseth: "Options depend on element type and index structure" describes UI behavior, not the decision the user needs to make. It tells the user the list will change without telling them what changed it or what to pick. This change replaces it with the constraint that actually matters here, that quantization is a float-only feature.

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
KOTungseth: This setting has no effect outside HNSW-based structures. It applies only to the hnsw, int8_hnsw, int4_hnsw, and bbq_hnsw index types. If Index structure is set to Flat or DiskBBQ, this field should be disabled rather than accepting a value that changes nothing in the estimate. If the UI cannot disable it, add "Applies only when Index structure is HNSW." below the field.

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
KOTungseth: "Length" is ambiguous next to a field about vector counts, since both read as quantities of something. "Number of values in each vector" names the thing being counted.

"Commonly 768 or 1536" instead of "for example," since these are the two most frequent real values rather than arbitrary illustrations.

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
KOTungseth: "not just documents" implied documents plus something else. "Not the number of documents" states the actual contrast.

Turned the example into a complete clause instead of a fragment, and tied both embeddings to the same document so the "one document, many vectors" point is visible in the example.

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
KOTungseth: "Each dimension" assumes the reader has already absorbed the Dimensions tooltip. "Each value in a vector" is self-contained and matches the phrasing I suggested in the previous tooltips.

Reframed the memory point as an explanation of why float is still the right pick.

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
KOTungseth: The three options were not parallel. HNSW got a mechanism, Flat got a category label, DiskBBQ got a storage location, so users have no shared way for comparing them. Each now states what the structure does and two state when to pick it.

Dropped "brute-force," which is jargon doing no work once "compares every vector" is there.

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
```diff
@@ -0,0 +1,617 @@
...
+    replicas:
+        'Number of replica shards (not including the primary). Total index copies = 1 primary + replica shards.',
```
KOTungseth:
```suggestion
        'The number of replica copies, not counting the primary. Each replica holds a full copy of the vector data, so replicas multiply storage and memory. The default is 1.',
```

KOTungseth: The parenthetical and the formula sentence were doing the same job twice. Kept the clarification inline and dropped the formula.

Added the sizing consequence, which is the reason this field exists in a sizing calculator. 

Changed "shards" to "copies" in the count. A user that enters 1 is asking for one additional copy of every shard, not one additional shard.

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
```diff
@@ -0,0 +1,617 @@
...
+    vectorsPerCluster:
+        'For DiskBBQ, vectors are grouped into clusters. This value sets how many vectors each cluster holds and affects cluster count and disk use.',
+    offHeapRam:
+        'Share of the posting lists (cluster vectors) kept in off-heap RAM. Lower values save memory; higher values improve query throughput and latency (all centroids are always fully in RAM).',
```
KOTungseth:
```suggestion
        'The percentage of clustered vector data, called posting lists, held in off-heap RAM. Supported values are 0 to 10 percent. Lower values save memory. Higher values raise query throughput and lower latency. This estimate assumes all centroids stay in RAM, since performance drops sharply otherwise.',
```

KOTungseth: "Improve query throughput and latency" is the one substantive error. Improving throughput means raising it and improving latency means lowering it, so a single verb cannot govern both. A reader tuning this field could reasonably take "improve latency" as "increase latency." Split into two verbs.

Pulled the centroid fact out of parentheses. It is the most decision-relevant sentence in the tooltip, because it tells the user there is a memory floor this slider cannot go below. Parentheses signal an aside.

Reordered the definition so the recognizable concept leads and the internal term follows. A user who has just chosen DiskBBQ knows vectors are clustered. "Posting lists" is the unfamiliar term.

Replaced the semicolon with periods so each idea gets one sentence.

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
```diff
@@ -0,0 +1,617 @@
...
+        'HNSW builds a graph for fast approximate search. Flat is exact brute-force. DiskBBQ stores vectors on disk.',
+    graphConnections:
+        'How many neighbors each node connects to in the HNSW graph. Higher values improve recall but increase memory and build time.',
+    quantization:
+        'Compresses vectors to reduce memory. Options depend on element type and index structure.',
+    replicas:
+        'Number of replica shards (not including the primary). Total index copies = 1 primary + replica shards.',
+    vectorsPerCluster:
+        'For DiskBBQ, vectors are grouped into clusters. This value sets how many vectors each cluster holds and affects cluster count and disk use.',
+    offHeapRam:
```
KOTungseth, line 77: The "Posting lists cached in off-heap RAM (%)" label carries more jargon than the others. Something like "Vector data cached in RAM (%)" would improve identification of the field.

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
KOTungseth: Are centroids actually always fully resident? Is "off-heap" the right qualifier?

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
```diff
@@ -0,0 +1,617 @@
...
+    indexStructure:
+        'HNSW builds a graph for fast approximate search. Flat is exact brute-force. DiskBBQ stores vectors on disk.',
+    graphConnections:
+        'How many neighbors each node connects to in the HNSW graph. Higher values improve recall but increase memory and build time.',
+    quantization:
+        'Compresses vectors to reduce memory. Options depend on element type and index structure.',
+    replicas:
+        'Number of replica shards (not including the primary). Total index copies = 1 primary + replica shards.',
+    vectorsPerCluster:
+        'For DiskBBQ, vectors are grouped into clusters. This value sets how many vectors each cluster holds and affects cluster count and disk use.',
```
KOTungseth:
```suggestion
        'The target number of vectors per cluster. Smaller clusters mean more centroids held in RAM. Larger clusters mean more data read per query. The default is 384.',
```

KOTungseth: "Affects cluster count and disk use" named the effect without a direction, so user are unable to tell which way to move the number. Both directions are now consequences.

Dropped "For DiskBBQ, vectors are grouped into clusters" as a standalone sentence and folded into the definition. If Index structure already gates this field, the conditional framing is redundant.

Called it a target. "Sets how many vectors each cluster holds" overpromises. 

Added the default. There was a discuss post (https://discuss.elastic.co/t/explanation-about-diskbbq-parameters/384248) identified in this project that said a cluster size default of 384 and a visit percentage default of 0.

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ResultsPanel.tsx
```diff
@@ -0,0 +1,189 @@
...
+                            size="xs"
+                            className="vectorSizingCalc__disclaimer"
+                        >
+                            {SIZING_DISCLAIMER}{' '}
+                            <EuiLink
+                                href={KNN_MEMORY_DOC}
+                                target="_blank"
+                                external
+                            >
+                                More info
```
KOTungseth:
```suggestion
                                Learn more
```

KOTungseth: We typically use "Learn more" instead of "More info" at Elastic

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/panels.render.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+    indexType: 'hnsw',
+    quantization: 'bbq',
+    replicas: 1,
+    hnswM: 16,
+    efConstruction: 100,
+    vectorsPerCluster: 384,
+    offHeapRamPercent: 10,
+}
+
+it('smoke: ComponentBreakdownTable renders rows + total + legend', () => {
```
KOTungseth: The "Component breakdown (per replica)" does not belong in the calculator design. This will be in the documentation so the writers can easily change the text in future Elastic versions.

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/panels.render.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+    expect(container.textContent).toContain('Total on disk')
+})
+
+it('smoke: ExplainersPanel renders all accordion sections', () => {
+    const { getByText } = render(
+        <EuiProvider colorMode="light">
+            <ExplainersPanel />
+        </EuiProvider>
+    )
+    expect(getByText('How it is computed')).toBeInTheDocument()
```
KOTungseth: The "How it is computed" does not belong in the calculator design. This will be in the documentation so the writers can easily change the text in future Elastic versions.

KOTungseth (changes requested):
Recommended copy changes for the tooltips and labels. 

## Open questions for engineering
- Does the disk estimate actually change with Vectors per cluster? BBQ benchmarking shows nearly identical disk usage across configurations. 
- Are centroids always fully resident? If not, the memory floor in the posting-lists tooltip needs to change.
- Is 384 the cluster size default?
- Which version does the calculator target? Replica default and float quantization default vary between stateful and Serverless.

## Structural recommendation
Disable dependent fields rather than explaining dependencies. Four fields apply conditionally, including Graph connections (HNSW only), Quantization (float only), Posting lists cache and Vectors per cluster (DiskBBQ only). Having them always present could introduce user confusion.

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
pmpailis: ++ makes total sense, thanks (missed this from the initial draft)

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
pmpailis: I might be missing something but I think that the graph connections slider is currently hidden (the panel changes) depending on the index type selected ([link](https://docs-v3-preview.elastic.dev/elastic/docs-builder/pull/2736/syntax/vector-sizing-calculator)).

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
pmpailis: ++ on the rewording, one minor note is that quantization is also supported for bfloat16 not just float elements ([link](https://github.com/elastic/elasticsearch/blob/2af77dd362fb639dfe80aefb020dcd69a94b69f0/server/src/main/java/org/elasticsearch/index/mapper/vectors/DenseVectorFieldMapper.java#L1905)).

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
pmpailis: `element type` is also how it's referred to in the [documentation](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/dense-vector#default-quantization-types) (and also in code).

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
pmpailis: ++  with one minor comment on the `Smaller clusters mean more centroids held in RAM. Larger clusters mean more data read per query.`  -> shall we also explain the effect on accuracy / overall performance ?

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
pmpailis: ++ updated

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
pmpailis: ++ reworded. Thinking about this maybe it'd be better to just  remove  the `768 or 1536`  reference? wdyt?

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
pmpailis: Technically no, this is not always the case. Although centroids are really small in size, and because they're hot, they usually reside in memory, it's not always true.  This is mainly an assumption for the sizing estimator as otherwise performance takes a serious hit.. Alternatively we could introduce another knob, albeit the memory requirements are almost negligible in the overall system. Worth noting that we also have a `[0,10]` supported range for the vector data (i.e. doesn't go all the way to 100).

pmpailis: maybe we could say something like `all centroids are assumed to be in RAM` ?

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/panels.render.test.tsx
pmpailis: Fair enough; added it to be transparent make the computations easier to understand from a single location but can move this to the doc pages as suggested.

KOTungseth: We can still use the information to update what currently exists in the docs.

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
KOTungseth: I updated to `This estimate assumes all centroids stay in RAM, since performance drops sharply otherwise.`

What do you think?

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
KOTungseth: Updated!

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
KOTungseth: I'm always a fan of "less is more"

KOTungseth (approved):

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
pmpailis: ++ looks nice, thanks!

src/Elastic.Documentation.Site/Assets/web-components/VectorSizingCalculator/components/ConfigurationPanel.tsx
pmpailis: what if we rephrase it to something like: 
```
Smaller values create more centroids in RAM and usually improve recall at the cost of performance.
```

pmpailis: updated!

pmpailis:
Thanks for the review @KOTungseth ! @reakaleek given that you've requested some changes a while back, would you like to take another look or are you ok with proceeding?  

Also, more generic question, shall we keep the changes in `docs-preview-local.yml` or should we revert them before merging ?

tools/vector_sizing_calculator/sizing-section.md
reakaleek, line 1: Do you think there is a more appropriate place for this file?

reakaleek:
> Also, more generic question, shall we keep the changes in `docs-preview-local.yml` or should we revert them before merging ?

We fixed that workflow on main already. Maybe with the same changes. (I guess so, because there is no merge conflict)

So IMO, you can just keep it.

tools/vector_sizing_calculator/sizing-section.md
pmpailis: Not sure about the best place for the draft tbh; maybe under  docs/syntax? or alternatively we could include [in this PR](https://github.com/elastic/docs-content/pull/7138/) ?

pmpailis: moved under docs/syntax; lmk if you prefer an alternative!

reakaleek: LGTM

reakaleek (approved):

pmpailis:
Thanks again  for the reviews @KOTungseth and @reakaleek !  Merging!

PR #2833, opened 2026-03-02
Codex: add icon to page frontmatter with docset override

reakaleek:
## What

- Add `icon` to page frontmatter alongside `description`
- Frontmatter icon overrides docset icon on Codex landing page cards
- Deprecate `icon` in docset.yml with a build-time hint

## Why

- Title and description live in index.md frontmatter, but icon was in docset.yml. Keeping icon in the same document as title and description is more consistent.
- This prepares for a future navigation refactor or redesign where icon may be used in navigation.

### DIFF
--- a/docs/syntax/frontmatter.md
+++ b/docs/syntax/frontmatter.md
@@ -10,21 +10,23 @@ In the frontmatter block, you can define the following fields:
 ---
 navigation_title: This is the navigation title. <1>
 description: This is a description of the page. <2>
-applies_to: <3>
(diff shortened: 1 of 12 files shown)


Mpdreamz (changes requested):
I don't think `icon` should live in yaml frontmatter. 

Its not something that applies to all pages, even if we might have icons in navigation at some point those should live in `navigation.yml` config IMO. Even in that case folks should not set icons to random page.

Issue #760, opened 2025-03-17
[Request] Generate config settings docs


kilfoyle:
Our first YAML-sourced [config settings file](https://github.com/elastic/kibana/blob/main/docs/settings-gen/source/kibana-alert-action-settings.yml) is now fully ready to render into the docs output.

I've:
 - Converted all descriptions, admonitions, and examples to Markdown format.
 - Added `applies_to` properties with the versioning details (Note that these aren't exhaustive: I'm basically saying "supported on self-hosted and cloud" versus "supported on self-hosted only", since that's all I know for any given setting, and that's all that we currently have documented.).

This early [Automated settings reference](https://elastic.github.io/docs-builder/syntax/automated_settings/) prototype looks very nice. I think what we still need is:

 - Split the setting descriptions into multiple paragraphs. 
 - Call out the `datatype`, and any `default` or `options` on separate lines.
 - Make sure that code examples are rendering.

---

**xpack.actions.ssl.proxyVerificationMode** ![Image](https://github.com/user-attachments/assets/9ae89de7-0342-43d4-98a5-c2302abfc196)

Controls the verification for the proxy server certificate that Kibana receives when making an outbound SSL/TLS connection to the proxy server.

Use `full` to perform hostname verification, `certificate` to skip hostname verification, and `none` to skip verification.

[Equivalent Kibana setting](/kibana/docs/reference/configuration-reference/general-settings.md#elasticsearch-ssl-verificationMode)

Datatype: `enum`
Default: `full`
Options:
 - full
 - certificate
 - none
---

lcawl:
In case it affects this issue, I've done some cleanup and consolidation of the manually-maintained copy of this settings file in https://github.com/elastic/kibana/pull/214934, so we'd want to make sure any changes there are reflected in the yaml too before replacing the markdown file.

kilfoyle:
@lcawl Thanks for the heads up. I can redo the yaml version of the Alert and Action settings file. Can you please tag me on the issue or PR for "Part 2" of the cleanup work when it happens? I might as well wait until that's complete just to avoid porting the changes over twice.

lcawl:
[...]

PR #2909, opened 2026-03-16
Add schema-compliant automated settings rendering

theletterf:
## Summary
- implement the full settings schema model for automated settings, including nested settings, `applies_to`, metadata, and setting-level admonitions/examples
- render settings as definition lists in HTML and definition-style blocks in LLM markdown, with inline applies-to output and inherited applies-to resolution
- normalize legacy AsciiDoc `[source,*]` snippets in settings text to markdown fenced code blocks, and expand docs/examples/tests for the new schema behavior

Closes https://github.com/elastic/docs-builder/issues/760

Contributes to https://github.com/elastic/docs-content/issues/5355

## Test plan
- [x] `dotnet test tests/Elastic.Markdown.Tests/Elastic.Markdown.Tests.csproj --filter \"FullyQualifiedName~SettingsInclusion\"`
- [x] `dotnet test tests/authoring/authoring.fsproj --filter \"FullyQualifiedName~settings directive\"`
- [x] Manual validation on `http://localhost:3000/syntax/automated_settings` for nested setting tip/admonition rendering and no raw `:::` artifacts

### DIFF
--- a/src/Elastic.Markdown/Myst/Components/ApplicableToViewModel.cs
+++ b/src/Elastic.Markdown/Myst/Components/ApplicableToViewModel.cs
@@ -2,6 +2,7 @@
 // Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information
 
+using Elastic.Documentation;
 using Elastic.Documentation.AppliesTo;
 using Elastic.Documentation.Configuration.Versions;
 
@@ -113,7 +114,7 @@ public class ApplicableToViewModel
 	{
 		var rawItems = new List<RawApplicabilityItem>();
 
-		if (AppliesTo.Stack is not null)
+		if (AppliesTo.Stack is not null && !IsGenericGa(AppliesTo.Stack))
 			rawItems.AddRange(CollectFromCollection(AppliesTo.Stack, ApplicabilityMappings.Stack));
 
 		if (AppliesTo.Product is not null)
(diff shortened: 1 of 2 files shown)


theletterf:
Thanks @kilfoyle and @florent-leborgne!

@florent-leborgne: I've implemented the split you preferred, stack-style badges stay with the setting name, and deployment or serverless-style availability is on a separate **Supported on:** line in the body. Where Kibana only encodes stack-style `applies_to`, we synthesize a sensible **Self-managed** line so that row isn’t empty (would that be OK?)

@kilfoyle: I've tightened YAML handling for `applies_to`, normalized a lot of the AsciiDoc and link shapes coming out of settings-gen, expanded docset substitutions for your sample tokens, and fixed a bug that could crash serve when internal links were rewritten after navigation resolution. Same-page fragments now account for both setting `id`s and group headings (including slugified group titles when there is no explicit `id`).

You can test that here: https://docs-v3-preview.elastic.dev/elastic/docs-builder/pull/2909/testing/kibana-settings-yaml-samples

Notice that each YAML block sits under an `##` filename heading; group headings inside the directive step down a level from that parent (same idea as the stepper).

docs/syntax/settings-schema-sample.yml
```diff
@@ -0,0 +1,113 @@
...
+    id: general-settings
+    description: |
+      Core settings shared across environments.
+    example: |
+      ```yaml
+      xpack.sample.enabled: true
+      ```
+    settings:
+      - setting: xpack.sample.enabled
+        id: xpack-sample-enabled
```
Mpdreamz: is id mandatory? I forgot, we might want to make sure the example only uses id overrides once as an example

PR #3013, opened 2026-04-01
Add skip-labels to evaluate-pr's output

cotti:
This pull request enhances the PR evaluation process in the `ChangelogPrEvaluationService` by adding support for collecting and outputting labels that cause a PR to be skipped due to exclusion rules. The main focus is on improving traceability and diagnostics when PRs are skipped based on label configurations.

Key changes:

**Skip label handling and output:**

* Added a new method, `CollectExcludeLabels`, which gathers all unique exclude-mode labels from both global and per-product `CreateRules`. This method returns a comma-separated string of labels or `null` if none are configured.
* Modified the skip check logic in `EvaluatePr` to collect skip labels and pass them to the output when a PR is skipped due to all products being blocked by label rules.
* Updated the `SetOutputs` method to accept an optional `skipLabels` parameter and output it as `skip-labels` when present. [[1]](diffhunk://#diff-37df27cba13363ff150536d3e4dac531776b6035e6efacbbe87f115dc4796efeL171-R174) [[2]](diffhunk://#diff-37df27cba13363ff150536d3e4dac531776b6035e6efacbbe87f115dc4796efeR201-R238)
* Ensured that when a PR is evaluated without required labels, the collected skip labels are included in the outputs for better diagnostics.


Mpdreamz (approved):
Lgtm, maybe add a quick unit test?

reakaleek (approved):

PR #3026, opened 2026-04-02
[API explorer] Replace raw x-state badges with applies_to popover

szabosteve:
## Summary

Relates to https://github.com/elastic/docs-eng-team/issues/452

- Replace raw `x-state` string badges (`<span class="version-badge">`) with structured `<applies-to-popover>` web component badges on API explorer endpoint pages
- Parse `x-state` into lifecycle (GA, Preview, Beta, Deprecated, Removed) and version data using the same logic as `OpenApiDocumentExporter`
- Badges now render consistently with the rest of the documentation site

## Changes

- **New:** `AvailabilityBadgeHelper.cs` -- parses `x-state` extensions and produces badge attributes for the `<applies-to-popover>` web component
- **Edit:** `OperationView.cshtml` -- replaced `version-badge` span with `applies-to-popover` below the h1 title; removed leftover debug spans
- **Edit:** `_PropertyItem.cshtml` -- same badge treatment for property-level `x-state` annotations
- **Edit:** `_UnionOptions.cshtml` -- propagates `VersionsConfiguration` to child contexts
- **Edit:** `RenderContext.cs` -- added optional `VersionsConfiguration` property to `PropertyRenderContext` and `UnionVariantsContext`

## Notes

- Serverless badges are **not shown** because the OpenAPI specs contain no serverless availability data in `x-state`
- Popover detail panel is disabled (`show-popover="false"`) -- can be enabled as a follow-up
- No new project dependencies added; the helper is self-contained in `Elastic.ApiExplorer`


lcawl:
I confirmed that the badges appear in a local test:

image

I couldn't find any at the parameter or property-level, however I didn't test exhaustively

Mpdreamz (changes requested):
I think it may be better to project x-state into the applies to yaml and using:

```csharp
var applicableTo = YamlSerialization.Deserialize<ApplicableTo>(yaml, appliesToDirective.Build.ProductsConfiguration);
```

Could then continue to take on the load of parsing it properly.

Mpdreamz (approved):

PR #3107, opened 2026-04-14
Search: Add ai_autocomplete_questions field for autocomplete

reakaleek:
## What
Add a new `ai_autocomplete_questions` AI enrichment field with a prompt targeting short, simple questions for search bar autocomplete. Keep the existing `ai_questions` field unchanged.

## Why
The existing `ai_questions` generates detailed technical questions (6-15 words) useful for semantic search, but too complex for autocomplete suggestions. Users searching naturally type simpler queries like "What is agent builder?" rather than "How do I import external tools using Model Context Protocol?"

## How
- Add `AiAutocompleteQuestions` property to `DocumentationDocument` with a prompt targeting 3-10 word questions
- Add lexical mapping with `SearchAsYouType` completion and `Completion` suggest multi-fields
- Add semantic text mapping for the new field
- Add `suggest` completion multi-field to existing `ai_questions` mapping (per #3108)

## Test plan
- Verify build passes
- Review generated questions after next AI enrichment run to confirm they are simpler and more natural


Mpdreamz (changes requested):
Should we have both? I feel the current will become super useful when we do more data/ai generated curation in the future.

Mpdreamz (approved):

PR #3108, opened 2026-04-14
Search: Add completion suggest field to ai_questions mapping

reakaleek:
## What
Add an FST-based `completion` type multi-field (`ai_questions.suggest`) to the ai_questions mapping for use with the completion suggester API.

## Why
The existing `ai_questions.completion` multi-field is `search_as_you_type` (inverted index with shingles), which is incompatible with the completion suggester API. The completion suggester requires a field of type `completion` (FST).

## How
Added a new `suggest` multi-field with `Completion()` type alongside the existing `completion` SearchAsYouType multi-field.

## Test plan
- Build compiles
- Requires a crawler run to create new indices with the updated mapping

## Notes
Named `suggest` since the `completion` multi-field name is already taken by SearchAsYouType.

### DIFF
--- a/src/Elastic.Documentation/Search/DocumentationMappingConfig.cs
+++ b/src/Elastic.Documentation/Search/DocumentationMappingConfig.cs
@@ -73,7 +73,8 @@ public class LexicalConfig : IConfigureElasticsearch<DocumentationDocument>
 			.MultiField("completion", mf => mf.SearchAsYouType()
 				.Analyzer("synonyms_fixed_analyzer")
 				.SearchAnalyzer("synonyms_analyzer")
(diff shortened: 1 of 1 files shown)


Mpdreamz (approved):
The search as you type is great if we have single autocomplete with mixed results and scoring. I believe we will bucket the results so question suggestions and doc suggestions are isolated from eachother. FST is great for that. 

However other things become harder too, synonyms, partial (mid sentence words) starts, and scoring (weighting). 

Approving we can play with this but I suspect we'd end up doing a search on the search_as_you type field instead.

reakaleek:
Will be handled in https://github.com/elastic/docs-builder/pull/3107

PR #3114, opened 2026-04-15
Fix API explorer empty navigation for single-tag OpenAPI specs

Mpdreamz:
## Summary

The local API explorer showed a shell page (title and layout) but an empty sidebar and overview for OpenAPI documents where every operation shares a single tag (for example the Kibana HTTP dashboard spec).

## Root cause

1. **Grouping**: The previous LINQ comprehension did not reliably build classification → tag → API groups, so no navigation tree was produced for some specs.
2. **Single-tag overwrite**: When all operations use one tag, endpoints are attached directly to the root `LandingNavigationItem`. The code then assigned `rootNavigation.NavigationItems = topLevelNavigationItems` even when that list was still empty, which cleared the sidebar.

## Changes

- Rebuild operation grouping with explicit nested `GroupBy` calls.
- Only assign or merge from `topLevelNavigationItems` when appropriate so single-tag specs keep their endpoints; merge when both operation nav and schema sections exist.
- Add `DashboardOpenApiNavigationTests` using `docs/dashboard-openapi.json`.
- Register `dashboard` in `docs/_docset.yml` and align `PhysicalDocsetTests` with the third API entry.

## Testing

- `dotnet test` on `Elastic.ApiExplorer.Tests` and the PhysicalDocset deserialization test.

### DIFF
--- a/docs/_docset.yml
+++ b/docs/_docset.yml
@@ -49,6 +49,7 @@ suppress:
 api:
    elasticsearch: elasticsearch-openapi.json
    kibana: kibana-openapi.json
+   dashboard: dashboard-openapi.json
 
 toc:
   - file: index.md
--- a/docs/dashboard-openapi.json
+++ b/docs/dashboard-openapi.json
@@ -1,0 +1,15783 @@
+{
+  "components": {
+    "schemas": {
+      "byteFormat": {
+        "additionalProperties": false,
+        "properties": {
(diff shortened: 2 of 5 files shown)


lcawl (approved):
I think it's likely going to be rare that we ever hit this in production, but doesn't hurt to fix.

cotti (approved):

PR #3123, opened 2026-04-16
Sitemap: Add protocol limit validation

reakaleek:
## What
Validate sitemap.xml against the sitemap protocol limits (50,000 URLs and 50 MB file size), throwing on violation and warning early at 80% thresholds.

## Why
The sitemap protocol enforces hard limits. Exceeding them produces an invalid sitemap that search engines will reject. Early warnings at 40k entries / 40 MB give the team time to implement sitemap index files before hitting the wall.

## How
- `SitemapBuilder.Generate()` throws `InvalidOperationException` when entries exceed 50k or file size exceeds 50 MB
- `Generate()` now returns a `SitemapResult(EntryCount, FileSizeBytes)` record
- Both call sites (`AssemblerSitemapService`, `AssemblerBuildService`) emit `EmitGlobalWarning` at 40k entries and 40 MB file size

## Test plan
- `dotnet test --filter SitemapTests` — 9 tests pass including 3 new ones covering the entry count limit and result metadata

### DIFF
--- a/src/services/Elastic.Documentation.Assembler/Building/AssemblerBuildService.cs
+++ b/src/services/Elastic.Documentation.Assembler/Building/AssemblerBuildService.cs
@@ -138,7 +138,20 @@ public class AssemblerBuildService(
 				.Distinct();
 			var now = DateTimeOffset.UtcNow;
 			var entries = urls.ToDictionary(u => u, _ => now);
-			SitemapBuilder.Generate(entries, assembleContext.WriteFileSystem, assembleContext.OutputWithPathPrefixDirectory);
+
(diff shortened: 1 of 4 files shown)


Mpdreamz (approved):
Good for now, we should be doing nested sitemaps automatically when we ever hit this error again (when api docs come back).

reakaleek:
> Good for now, we should be doing nested sitemaps automatically when we ever hit this error again (when api docs come back).

I didn't actually see the error.. we are currently at ~8k entries so there is still a lot of buffer.

But I thought that this might be useful in the future.. especially when we add api docs.

PR #3128, opened 2026-04-16
ContentDateEnrichment: Use real IngestChannel in tests, skip redundant semantic _update_by_query

reakaleek:
## What
- Replace hand-rolled scripted upserts in integration tests with the real `IngestChannel<DocumentationDocument>` from `Elastic.Ingest.Elasticsearch`
- Skip the redundant `_update_by_query` on the semantic index

## Why
The tests were manually constructing bulk requests that didn't match what `HashedBulkUpdate` actually does in production. Running the real channel revealed that:
- On a fresh index, the channel uses `index` bulk actions (pipeline fires automatically)
- On index reuse, the channel uses `update` actions (HashedBulkUpdate scripted upserts — pipeline skipped)
- The semantic index gets dates resolved during the `_reindex` step (`final_pipeline` fires), making the subsequent `_update_by_query` redundant (always 0 matches)

## How
- Tests now use `DocumentationMappingContext` type context with full analysis config, matching the production exporter setup
- Second-run tests create a fresh channel with `indexNameOverride` to exercise the reuse/update path
- Removed the `_update_by_query` call for the semantic read alias in `StopAsync`

## Test plan
- All 5 integration tests pass (requires Docker for Testcontainers)
- `dotnet test tests-integration/Elastic.ContentDateEnrichment.IntegrationTests/`

## Notes
Diagnostic output from the tests confirmed `HashedBulkUpdate` stores `content_last_updated` as NULL (field missing), not `0001-01-01`. The existing `must_not range` filter handles this correctly.


Mpdreamz (approved):

reakaleek:

PR #3163, opened 2026-04-21
Changelog: Bundle upload support and Private-Public Bundling SQS Lambda

cotti:
This pull request introduces the new "changelog-scrubber" Lambda function, which sanitizes changelog and bundle files in S3 by removing private repository references before making them public. The changes include the Lambda's implementation, build and deployment automation, and supporting project files. The release workflow is updated to build, package, and deploy this Lambda alongside the existing link index updater Lambda.

Key changes:

**Changelog Scrubber Lambda Implementation**

* Added the `changelog-scrubber` Lambda function in `Program.cs` to process SQS events, scrub changelog and bundle YAML files using `LinkAllowlistSanitizer`, and copy sanitized versions to a public S3 bucket. Handles S3 object creation/removal events and supports pass-through for specific JSON files.
* Created `SerializerContext.cs` for optimized JSON serialization of Lambda event types.
* Added project file `docs-lambda-changelog-scrubber.csproj` with dependencies, embedded resource configuration, and AOT publishing settings.
* Included a Dockerfile (`lambda.DockerFile`) for building the Lambda binary on Amazon Linux 2023 with .NET 10 AOT.
* Added AWS Lambda deployment defaults in `aws-lambda-tools-defaults.json`.
* Provided a detailed `README.md` explaining build, event handling, and scrubbing logic.

**CI/CD and Release Workflow Updates**

* Added a new GitHub Actions workflow (`build-changelog-scrubber-lambda.yml`) to build and archive the Lambda binary.
* Updated `release.yml` to build, package, and deploy the changelog-scrubber Lambda, including artifact handling and release asset upload. Also refactored jobs for clarity and robustness. [[1]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L72-R91) [[2]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L96-R103) [[3]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L116-R166) [[4]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R212-R214) [[5]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R223) [[6]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R232-R234)

**Link Allowlist Scrubbing Enhancements**

* Improved `LinkAllowlistSanitizer` to support scrubbing of individual changelog entries and free-text fields, and added regexes for GitHub URL and short-form reference detection. [[1]](diffhunk://#diff-9360a6ca4e69d1b73f1089d083543c9f7fbcc120c0c83b4372bdaababe18e037R5) [[2]](diffhunk://#diff-9360a6ca4e69d1b73f1089d083543c9f7fbcc120c0c83b4372bdaababe18e037R14-R23)

These changes collectively add an automated, secure way to sanitize and publish changelog and bundle files, ensuring no private repository references are leaked to the public.


Mpdreamz (changes requested):
## Request changes

Thanks for wiring up the private → public pipeline and the deserialize → mutate → serialize scrubbing path. With how we plan to use these buckets, I think we need a few adjustments before this is safe to ship.

### Bucket roles (why scrubbing must be strict)

- **Private S3** backs indexing into Elasticsearch (internal/trusted context).
- **Public S3** is what **docs-builder** will use to render release notes **on anyone’s machine** (untrusted context).

So the **public** objects must not carry internal repo names, PR numbers, or URLs that slipped through a new field or a renamed property. The bar for the public artifact is “no disclose,” not only “no clickable link.”

### Sentinel-style redaction vs real removal

Using `# PRIVATE: <original>` on structured refs can still leave **the full original reference in plain text** in YAML that lands in the public bucket. That conflicts with the goal above: if the private copy remains authoritative in private S3, the public copy should **remove** disallowed references (or replace them with a **non-identifying** placeholder, e.g. a fixed string with no owner/repo/URL), not preserve them after a prefix.

Please align bundle scrubbing with that requirement: **no residual GitHub URLs or `owner/repo#N` patterns** in public bundle YAML unless they are explicitly allowlisted.

### Post-serialize validation

Deserialize → mutate → serialize is the right core loop, but it won’t catch:

- **New** YAML fields that contain links in a future schema revision.
- **Renamed** fields where scrubbing wasn’t extended.

Please add a **post-serialize validation pass** on the **final string** destined for the public bucket (or equivalently, a second parse + walk), e.g.:

- Assert the serialized text contains **no** GitHub PR/issue URLs and **no** short-form `owner/repo#` references **outside** an explicit allowlist (reuse or share the same patterns as `ScrubText` / link detection).
- Fail the Lambda batch item (or fail CI tests) if validation fails so we don’t quietly publish a leaky object.

That gives us defense in depth when the model changes.

### Summary

**Request:**

1. Public bundle/changelog YAML should **remove or non-identifying-redact** disallowed references so internal details aren’t recoverable from the file.
2. **Post-serialize validation** on emitted content to guard against **new/renamed fields** leaking links.

Happy to revisit once those are addressed.

cotti:
Changes for residual private references:

- Added `StripBundleSentinels` -- strips `# PRIVATE:` entries from all prs/issues across a bundle's entries. The Lambda now calls this after `TryApplyBundle` before serializing bundle output.

Changes for post-serialize validation

- Added `ValidateNoPrivateReferences(string yaml, IReadOnlyList<string> allowRepos)`, which scans the final serialized YAML string using the same `GitHubUrlRegex` and `ShortFormRefRegex` patterns, plus checks for residual `# PRIVATE:` sentinels. Throws InvalidOperationException if any non-allowlisted references are found.

The Lambda now calls this on every output (both changed and unchanged paths) before writing to the public bucket. If validation fails, the exception bubbles up and the SQS message goes to DLQ -- no unsafe content gets published.

Added some test scenarios to cover scrubbing:

`StripBundleSentinels`:

- Removes sentinels while keeping allowed refs in a mixed list
- All-sentinel lists become empty
- Null prs/issues are preserved as null
- Multiple entries are all stripped

`ValidateNoPrivateReferences`:

- Clean YAML with only allowed refs passes
- Private GitHub URL triggers exception
- Private short-form owner/repo#N triggers exception
- Residual # PRIVATE: sentinel triggers exception
- All-allowed content passes
- Empty YAML passes
- Mixed allowed + private throws (citing the private one)

`TryApplyChangelogEntry` edge cases:

- Mixed allowed + private PRs: keeps allowed, drops private, verifies count
- Private refs across ALL fields (prs, issues, description, impact, action): verifies complete scrubbing

Mpdreamz (changes requested):
## Request changes

### What we want from the two buckets

- **Private S3** is the trusted store: full YAML, real PR/issue links, whatever indexing and internal workflows need. It does **not** need redaction hints.
- **Public S3** is what **docs-builder** will use to render release notes on **any** machine. That copy should contain **only** allowlisted GitHub references. Anything else should be **absent**, not replaced with a marker that something used to be there.

So the public pipeline should not depend on an intermediate `# PRIVATE:` representation at all.

### Why drop the sentinel round-trip for public output

Right now the flow can read as: **rewrite disallowed refs to `# PRIVATE: …`, then strip sentinels, serialize, validate** (plus validation that still looks for residual `# PRIVATE:`).

That works, but it adds moving parts and two different concepts of “clean”:

1. **Mental model** — Operators and future maintainers should only need one rule for public: **disallowed links are removed**. They should not have to reason about a string format that is explicitly **not** allowed in the final artifact anyway.
2. **Complexity** — If nothing outside private S3 needs “a hint something was redacted,” the sanitizer’s public path can **omit** disallowed list entries (or never add them) **directly**, instead of **tag then strip**. Same outcome, fewer steps, fewer places for ordering bugs (e.g. strip forgotten on a code path).
3. **Validation** — `ValidateNoPrivateReferences` should primarily enforce **“no non-allowlisted link-shaped content”** in the serialized string. A separate **“no `# PRIVATE:`”** check is reasonable as a temporary guard while sentinel output still exists somewhere; once public code never emits that token, that check becomes redundant. Prefer **one** contract: public YAML must not contain private links **or** redaction sentinels, achieved by **never writing** either.

### Requested changes

1. **Public / Lambda scrubbing path** — For bundle (and any shared entry logic used only for public), **do not** use `# PRIVATE:` as an intermediary. Resolve each PR/issue reference against the allowlist; **drop** disallowed entries (or equivalent) before serialization. Keep **deserialize → mutate (remove) → serialize → validate string** without a sentinel phase for this path.
2. **Refactor / split if needed** — If other callers still need sentinel behavior for **private-side** tooling, isolate that behind an API that is **not** used when writing to the public bucket, or branch clearly so “public emission” never constructs `# PRIVATE:`.
3. **Strip / validate** — Remove `StripBundleSentinels`-style workarounds once they are unnecessary, and tighten `ValidateNoPrivateReferences` to match the final contract (link scans + any residual rules you still need during migration).

### Summary

**Private** artifacts can keep private links. **Public** artifacts should be **link-scrubbed by removal**, not by a two-phase sentinel encoding. Please adjust this PR so the public scrubbing path reflects that simpler model and the rationale above.

Happy to re-review once that’s in place.

cotti:
### Public / Lambda scrubbing path:

The public path now uses `FilterReferenceList`, which resolves each reference against the allowlist and either keeps it or drops it. The flow is:

`deserialize -> FilterReferenceList drops disallowed refs + ScrubText removes private links from text fields -> serialize -> ValidateNoPrivateReferences on the final string`. 

For bundles, `ScrubBundleForPublic` iterates entries and applies `TryApplyChangelogEntry` (which uses `FilterReferenceList`) to each, plus scrubs the bundle-level description. The `# PRIVATE:` constant is not referenced on any public code path.

Refactor / split if needed

The split is now clean at the method level:

*Private-side*: `TryApplyBundle -> ApplyToReferenceList -> ProcessPlainReference (produces sentinels)`. Used only by the `docs-builder changelog bundle` CLI.

*Public-side*: `ScrubBundleForPublic / TryApplyChangelogEntry -> FilterReferenceList (drops disallowed refs, never creates sentinels)`. Used by the Lambda.

The two paths share only the parsing utility (`TryGetGitHubRepo`) and the allowlist (`BuildAllowSet`).

Remove `StripBundleSentinels`-style workarounds once unnecessary, and tighten `ValidateNoPrivateReferences`

`StripBundleSentinels` was removed, alongside `DropSentinels`.

`ValidateNoPrivateReferences` enforces one contract: the serialized YAML must not contain non-allowlisted GitHub URLs, owner/repo#N short forms, or residual `# PRIVATE:` strings. The sentinel check remains as a transitional safety net (it will never fire since the public path never produces sentinels, but it catches programming errors if someone accidentally routes private-side output to the public path). The primary enforcement is the link-pattern scans.

Mpdreamz (approved):
## LGTM

The private → public scrub + post-serialize validation is in good shape for the intended defense-in-depth story. Shapes up.

### Follow-up (not blocking)

**Align scrubbing/validation with every PR/issue format we document or accept** — e.g. catalog what `docs-builder` (and changelog/bundle schema) allow in `prs` / `issues` and in prose fields (bare numbers, `owner/repo#N`, full URLs, alternate hosts). `ValidateNoPrivateReferences` and `ScrubText` only match a narrow set of **GitHub PR/issue URL** and **short-form** patterns today. If the product is effectively **freeform strings** in those fields, we should **tighten the contract** (schema/docs) and/or **expand** scrub + validation in lockstep so we do not give the impression of “all links removed” when some shapes slip through.

Please track a follow-up to: (1) list supported reference formats, (2) assert scrub/validation each cover the same set, and (3) if inputs stay freeform, add validation or heuristics appropriate to that.

### Gaps in current pattern coverage (defense-in-depth / known blind spots)

Below are **examples the current `GitHubUrlRegex` + `ShortFormRefRegex` (and thus post-validate) do *not* flag** as disallowed by owner/repo, even though they may still point at private work. `TryApply` paths that use `ChangelogTextUtilities.TryGetGitHubRepo` can accept additional shapes; validate/scrub are narrower.

| Example | Notes |
|--------|--------|
| `https://www.github.com/elastic/secret/pull/42` | `www.github.com` — not matched (pattern expects `github.com` immediately after `//`) |
| `https://github.com/elastic/secret` | Repo home only, no `pull`/`issues` path |
| `https://github.com/elastic/secret/wiki/Page` | Wiki / non–PR–issue path |
| `https://github.com/elastic/secret/compare/8.0...8.1` | Compare (no `/pull/`, `/issues/`) |
| `https://github.com/elastic/secret/blob/main/README.md` | Blob/tree paths |
| `https://github.com/elastic/secret/commit/abc123def` | Commit link |
| `git@github.com:elastic/secret.git` | SSH remotes in text |
| `https://gh.internal.example.com/...` | Other GitHub hostnames (Enterprise) |

(Allowlisted `https://github.com/org/repo/pull|issues/n` and allowlisted `owner/repo#n` **do** line up and pass as expected.)

This table is to inform the follow-up: either narrow accepted input, or broadened detection where we need parity.

Thanks for the work on this — happy to help refine the follow-up if useful.

PR #3303, opened 2026-05-12
Add docs as product

lcawl:
This PR updates products.yml so we can start down the path of creating a release notes page specifically for docs.
It uses the `product-reference:false` clause as described in https://elastic.github.io/docs-builder/configure/site/products/#structure


KOTungseth (approved):
It's finally happening!!

PR #3313, opened 2026-05-13
Add content_type to DocumentationDocument for site search parity

Mpdreamz:
## Why

Site-wide search combines documentation with other surfaces that classify documents in `_source` differently than our plain `type` field alone. We need a persisted keyword that can diverge from `type` when JSON carries an explicit classification, while staying aligned with website-ai-search so a future shared `DocumentationDocument` can round-trip the same shape. The intentional duplication supports **wider site search filtering** across those indices and clients.

## What

- Add `content_type` on `DocumentationDocument` with the same JSON-first normalization pattern used on the site side (stored value omitted when it matches `type` for stable round-trips).
- Document in XML why `type` and `content_type` coexist (polymorphic `$type` / JSON-ignored CLR type on the other repo; persisted filter via `content_type`).
- Extend serialization tests for emit, override-from-JSON, and round-trip.

### DIFF
--- a/src/Elastic.Documentation/Search/DocumentationDocument.cs
+++ b/src/Elastic.Documentation/Search/DocumentationDocument.cs
@@ -35,6 +35,34 @@ public record DocumentationDocument
 	[JsonPropertyName("type")]
 	public required string Type { get; set; } = "doc";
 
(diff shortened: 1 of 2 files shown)


reakaleek (approved):

theletterf:
Where would the ContentType value come from? I understand this is to leave the `type:` content-type frontmatter for other things?

PR #3316, opened 2026-05-13
Tracking: fix redirect validation skipping deleted files at repo root

### LINKED ISSUE #1920: Deleting a page and replacing with an xref evades the "deleted page" check
### Describe the bug

these two PRs didn't trigger an error for deleted pages:

https://github.com/elastic/docs-content/pull/2989
https://github.com/elastic/docs-content/pull/2811

both of them deleted a page, but added an equivalent xref in the sidebar. CI passed cleanly. when the PRs were merged, the pages 404'd

note: the crosslinks within the repo were already cleaned up. does the deleted page logic still trip if there are no incoming links? I want to test this but want to log this bug asap

### Expected behavior

you can't run or hide from the "deleted page with no redirect" error

### Steps to reproduce

_No response_

### Tooling

- [x] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

reakaleek:
## What

- Fix `IntegrationGitRepositoryTracker` so the redirect-validation step (`docs-builder diff validate`) no longer silently drops every changed file when the docset lives at the repository root.
- Add unit tests for the tracker and end-to-end tests for `LocalChangeTrackingService.ValidateRedirects` covering both docset-at-root and docset-in-subfolder layouts.

## Why

- On CI, the `Validate redirect rules` step in `docs-actions/.github/workflows/docs-build.yml` is meant to fail a PR check when a markdown file is deleted or renamed without a corresponding entry in `redirects.yml`.
- For repos whose docs live directly at the repository root (for example [`elastic/docs-content`](https://github.com/elastic/docs-content)), the relative path passed to the tracker is `.`. The previous code turned that into a `./` prefix and used `StartsWith` to filter the `ADDED_FILES` / `MODIFIED_FILES` / `DELETED_FILES` / `RENAMED_FILES` environment variables. PR file paths from the GitHub API have no `./` prefix, so the filter dropped every entry and the validator reported zero changes.
- This let PRs that delete markdown without setting up redirects pass the check. Example: [`elastic/docs-content#6479`](https://github.com/elastic/docs-content/pull/6479) deletes `troubleshoot/deployments/serverless.md` with no `redirects.yml` entry and the check passed green.

## Notes

- The fix normalizes both `""` and `"."` to an empty `LookupPath`, which makes `StartsWith("")` match every path. Real subfolder docsets (for example `docs/`) keep the trailing-slash anchor to avoid false positives like `docs-extra/foo.md` matching `docs`.
- Confirmed the regression locally: reverting the fix turns four of the new tests red (both unit-level tracker tests and the end-to-end docset-at-root tests); with the fix, all 12 pass.
- The new tests mutate process-wide env vars, so they share an xUnit collection with `DisableParallelization = true` to avoid races.


src/authoring/Elastic.Documentation.Refactor/Tracking/IntegrationGitRepositoryTracker.cs
```diff
@@ -9,7 +9,18 @@ namespace Elastic.Documentation.Refactor.Tracking;
...
+	// In that case there is no prefix to match against, so we use an empty string which makes
+	// StartsWith("") true for every path. Otherwise we anchor with a trailing slash to avoid
+	// false-positive prefix matches (e.g. "docs/" should not match "docs-extra/foo.md").
+	private string LookupPath { get; } = NormalizeLookupPath(lookupPath);
+
+	private static string NormalizeLookupPath(string lookupPath)
+	{
+		var trimmed = lookupPath.Trim(['/', '\\']);
+		return trimmed is "" or "." ? "" : $"{trimmed}/";
+	}
```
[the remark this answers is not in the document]
reakaleek: I checked the finding against the current code and it's not worth fixing here. Skipping with reason:

Skipping the null guard on NormalizeLookupPath.

Pre-existing, not introduced by this PR. The original code $"{lookupPath.Trim(['/', '\\'])}/" had the exact same null-deref shape on the same parameter. My refactor only added the is "" or "." branch — it didn't change null behavior.

The parameter is non-nullable in a nullable-enabled project. Elastic.Documentation.Refactor.csproj sets <Nullable>enable</Nullable>, and the primary-constructor signature IntegrationGitRepositoryTracker(string lookupPath) types the parameter as non-null. Static analysis already flags any caller that passes null without suppression.

The single internal caller is safe. LocalChangesService.cs:51 passes the result of Path.GetRelativePath(...), which never returns null.

NormalizeLookupPath is private static. The project's own style guide is explicit on this:

Validate at boundaries — public API entry points. Trust internal code — don't over-validate private methods.

Pushing a null check into the private helper duplicates what the nullable annotation on the public constructor already enforces.

The constructor is a primary constructor. Adding ArgumentNullException.ThrowIfNull at the boundary would require converting it to a regular constructor, which is a stylistic regression away from the codebase's preferred form, and still wouldn't add real safety beyond what nullable analysis gives us.

If a future caller really needs runtime hardening (e.g. a deserialization path that bypasses nullable), I'd add ArgumentException.ThrowIfNullOrWhiteSpace(lookupPath) on the public constructor in a separate change, not on a private helper.

No code changes, nothing to validate. Leaving the PR as-is.

src/authoring/Elastic.Documentation.Refactor/Tracking/IntegrationGitRepositoryTracker.cs
```diff
@@ -9,7 +9,18 @@ namespace Elastic.Documentation.Refactor.Tracking;
 
 public class IntegrationGitRepositoryTracker(string lookupPath) : IRepositoryTracker
 {
-	private string LookupPath { get; } = $"{lookupPath.Trim(['/', '\\'])}/";
+	// When the docset lives at the repository root, lookupPath is "." (or empty after trimming).
+	// In that case there is no prefix to match against, so we use an empty string which makes
+	// StartsWith("") true for every path. Otherwise we anchor with a trailing slash to avoid
```
Mpdreamz: We should really always use https://github.com/elastic/docs-builder/blob/ce8705a5e54e4f4a05af238da899ed4b7d767232/src/Elastic.Documentation/Extensions/IFileInfoExtensions.cs#L101 `IsSubPathOf` for these types of checks.

Mpdreamz (commented):
Add a note about using StartsWith for paths in general should use our extensions method to check this safely in general. 

PR LGTM, we can do a pass over this another time

cotti (approved):

src/authoring/Elastic.Documentation.Refactor/Tracking/IntegrationGitRepositoryTracker.cs
reakaleek: Fixed in c009034c989b0c752fec787e475c074bfae61a9c — IntegrationGitRepositoryTracker now takes the git root and documentation source directories and filters CI paths with IsSubPathOf instead of StartsWith.

Mpdreamz (approved):

PR #3318, opened 2026-05-13
feat(settings): add :deployment: filter option to {settings} directive

florent-leborgne:
## Why

Documentation consumers often only care about settings available for a specific deployment model (e.g. Elastic Cloud Hosted). Without a filter, they have to manually skip settings that do not apply to their environment, which adds friction and noise.

This is done with the idea of automating this page https://www.elastic.co/docs/reference/kibana/cloud/elastic-cloud-kibana-settings like other settings pages - but because this specific page is rather a legacy, ECH-only, view of kibana settings, I want to find a way to also relying on the individual yaml settings files (and reference them all in that page but to just show those on ECH). This allows this page to keep existing while centralizing its maintenance.

This page has value for ECH users as it avoids that they have to go through each individual settings page, so it would be sad to get rid of it while there is a simple path to keep it while avoiding manual maintenance.

Today some eng teams are confused and are updating this page instead of yaml files because it's the only .md file they find with their settings in it. By having it use a similar logic to other kibana settings files, we're making this a lot more obvious.

## What

Adds a `:deployment:` option to the `{settings}` directive. When set to `ech`, `ece`, `eck`, or `self`, the renderer hides every setting whose `applies_to.deployment` block either omits the requested deployment or marks it as removed/unavailable. Settings with no `applies_to` metadata at all remain visible (they are universally applicable).

The filtering is applied consistently in three places: HTML rendering (via `SettingsViewModel`), table-of-contents generation, and anchor collection. A warning is emitted when an unknown deployment value is supplied.

Documentation for the new option is added to `docs/syntax/automated_settings.md`, and live filtered examples using real Kibana YAML fixtures are added to the Kibana settings test page.


theletterf (approved):
Nice!

PR #3319, opened 2026-05-13
Don't autolink URLs used as markdown link text

### LINKED ISSUE #3317: Using a URL as text on a markdown link (external) results in two "external" icons
### Describe the bug

Image

https://www.elastic.co/docs/troubleshoot/observability/apm-agent-java/apm-java-agent

```
If you do so, **attach your debug logs** so that we can analyze the problem. Upload the **complete** logs to a service like [https://gist.github.com](https://gist.github.com).
```

### Expected behavior

only one external icon

### Steps to reproduce

put this anywhere

```
If you do so, **attach your debug logs** so that we can analyze the problem. Upload the **complete** logs to a service like [https://gist.github.com](https://gist.github.com).
```

### Tooling

- [x] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

shainaraskas:
## Why

When a URL is used as both the visible text and the target of a markdown link — e.g. [https://gist.github.com](https://gist.github.com) — the page renders two external-link icons. The custom bare-URL autolinker runs before the standard link parser, so it matches the URL inside the [...] text and produces a nested <a target="_blank">. The CSS rule that paints the external icon then fires on both anchors. Closes #3317.

Chose to do this through a code fix rather than fixing instances because sometimes the URL link text is the most accurate + making it a bare link is busywork + the problem is likely to recur.  the most expedient fix is in the code

## What

HtmxLinkInlineRenderer now walks the parent chain of each LinkInline it's about to write; if any ancestor is also a LinkInline, it skips the wrapping <a> and just renders the children. The outer link keeps its target="_blank" + external icon; the inner autolinked URL becomes plain text inside it. One anchor, one icon, valid HTML.

Adds two regression tests covering the exact issue case and a URL embedded mid-text inside link text.

| bad | good |
| --- | --- |
| image | image |


theletterf (approved):
Thanks!

PR #3321, opened 2026-05-14
ci: declare contents:read on smoke-test workflow

arpitjain099:
Adds a workflow-level `permissions: contents: read` block. The job here only checks out the repository and runs its tests / validation; no GitHub API call beyond the initial checkout is needed.

CVE-2025-30066 (the March 2025 `tj-actions/changed-files` supply-chain compromise) is the canonical motivation: a tampered third-party action exfiltrated `GITHUB_TOKEN` from workflow logs and the leaked token retained whatever scope was issued at the workflow level. Per-workflow caps bound that runtime authority irrespective of repo or org default, give drift protection if the default ever widens, and register with OpenSSF Scorecard's Token-Permissions check (which only credits explicit per-workflow declarations).

YAML validated locally with `yaml.safe_load`.


Mpdreamz:
@arpitjain099 the change looks good to me, thanks! Mind signing our CLA that way I may pull this in.

arpitjain099:
> @arpitjain099 the change looks good to me, thanks! Mind signing our CLA that way I may pull this in.

@Mpdreamz I signed the CLA earlier, can you please check again.
https://github.com/elastic/docs-builder/pull/3321#issuecomment-4452493571

arpitjain099:
@Mpdreamz CLA is signed now (EasyCLA check passing). Should be ready to merge whenever you have a moment.

cotti (approved):

PR #3326, opened 2026-05-15
Spruce up core nav

marciw:
WIP


theletterf:
Integrating these into the demo branches, thanks!

PR #3327, opened 2026-05-16
Fix docs-builder version output

theletterf:
## Why

`docs-builder --version` was reporting the CLR assembly version (`1.0.0.0`) instead of the semantic release version shown by older releases.

## What

Handle the fast `--version` path before Argh's intrinsic handler and print the assembly informational version, falling back only when that metadata is unavailable.

### DIFF
--- a/src/tooling/docs-builder/Program.cs
+++ b/src/tooling/docs-builder/Program.cs
@@ -2,6 +2,7 @@
 // Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information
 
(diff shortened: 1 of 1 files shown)


Mpdreamz (changes requested):
This needs fixing in upstream nullean/argh

Mpdreamz:
https://github.com/nullean/argh/pull/54

theletterf:
Closing this local workaround in favor of fixing the version source in Argh itself.

Upstream PR: https://github.com/nullean/argh/pull/55

PR #3328, opened 2026-05-16
Fix comparer collection initializers

theletterf:
## Why

Current `main` fails to compile because the compiler treats the comparer `with(...)` collection expressions as calls to a missing method.

## What

Replace those expressions with explicit `Dictionary` and `HashSet` constructors while preserving the intended case-insensitive comparers. Verified with `dotnet build src/tooling/docs-builder/docs-builder.csproj -c Release --verbosity minimal`.


theletterf:
Silly me, this is because I wasn't on the latest .NET.

PR #3330, opened 2026-05-18
Bump Nullean.Argh packages from 0.16.0 to 0.16.2

theletterf:
## Why

Nullean.Argh 0.16.2 was released on May 17, 2026 — keeping the dependency current picks up any bug fixes in the patch releases.

## What

Bumps `Nullean.Argh`, `Nullean.Argh.Hosting`, and `Nullean.Argh.Interfaces` from `0.16.0` to `0.16.2` in `Directory.Packages.props`.

PR #3331, opened 2026-05-18
Fix --version and __schema to report the semantic release version

theletterf:
## Why

`docs-builder --version` was printing the CLR assembly file version (`1.0.0.0`) instead of the MinVer-assigned semantic version. The same value leaked into the `version` field of the `__schema` output (used to generate `docs/cli-schema.json`). On a tagged release the version should be clean — e.g. `1.9.0` — not the CLR default.

This is a follow-up to the Nullean.Argh 0.16.2 bump (#3330), which exposed the mismatch because Argh 0.16.2 switched from `AssemblyFileVersionAttribute` to `AssemblyInformationalVersionAttribute` internally. The same fix applies to Argh 0.16.0 as well.

## What

Both `--version` and `__schema` are now intercepted in `Program.cs` before Argh's intrinsic handler runs:

- `--version` prints `AssemblyInformationalVersionAttribute` directly and exits.
- `__schema` redirects `Console.Out` to a `StringWriter`, registers a `ProcessExit` handler that patches the `version` JSON field with the informational version, then lets Argh write and exit as normal.

`docs/cli-schema.json` is regenerated to reflect the change (`1.0.0.0` → `1.0.0` for local non-MinVer builds; tagged releases will show the clean semver).

> **Note:** merge after #3330, which adds `jq`-based version-field exclusion to the CI schema diff so the check stays green across canary commits.

### DIFF
--- a/docs/cli-schema.json
+++ b/docs/cli-schema.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "name": "docs-builder",
-  "version": "1.0.0.0",
+  "version": "1.0.0",
   "description": null,
   "reservedMetaCommands": [
     "__complete",
--- a/src/tooling/docs-builder/Program.cs
+++ b/src/tooling/docs-builder/Program.cs
@@ -2,6 +2,7 @@
 // Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information
 
+using System.Reflection;
 using Documentation.Builder;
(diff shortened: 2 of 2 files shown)


theletterf:
@Mpdreamz After we merge #3330, we still need this fix.

Mpdreamz:
Fixing this in Argh instead again :)

Mpdreamz:
https://github.com/nullean/argh/pull/56

Mpdreamz:
No longer needed :)

PR #3333, opened 2026-05-18
[1/5] VSA: Merge Elastic.Documentation.Api.* into single project

Mpdreamz:
## Why

The three-project split (Api.Core / Api.Infrastructure / Api.App) implemented a hexagonal layering that provided no external reuse benefit — all consumers are internal to this repo. The same split caused Mcp.Remote to accumulate a dependency on the full Api stack just to call one OTel helper and use two search interfaces, making it harder to reason about and maintain independently.

## What

- `Elastic.Documentation.Api.Core`, `Api.Infrastructure`, and `Api.App` are collapsed into a single `Elastic.Documentation.Api` project (SDK.Web, AOT-publishable, same `AssemblyName=docs-builder-api`).
- `EuidSpanProcessor`, `EuidLogProcessor`, and a new `AddEuidEnrichment` helper move to `Elastic.Documentation.ServiceDefaults` so any web app in the repo can enrich spans/logs with the user euid without pulling in the full Api project.
- The three search/changes gateway interfaces (`IFullSearchGateway`, `INavigationSearchGateway`, `IChangesGateway`) and their DTOs move to `services/Elastic.Documentation.Search` alongside their implementations — the Search service no longer references the Api project at all.
- Mcp.Remote drops both Api.Core and Api.Infrastructure project references entirely; its deps are now `ServiceDefaults` + `services/Search` + `services/Assembler` only.

## How

`AddDocsApiOpenTelemetry` in the merged project becomes a thin wrapper: it calls the new `AddEuidEnrichment` (which owns the Elastic OTel SDK init, service.version wiring, euid cookie enrichment, and processors) and then adds the five API-specific activity sources on top. Mcp.Remote calls `AddEuidEnrichment` directly instead.

---

Part of a move to vertical slice architecture to be able to cleanly lift search contract:
1. #3333 — Merge Elastic.Documentation.Api.* into single project
2. #3335 — Collapse Gateway+Usecase into *Service per feature
3. #3345 — Slim Elastic.Documentation, extract heavy deps to Tooling
4. #3347 — Slim Configuration to YAML schema; organize root namespaces
5. #3350 — Extract Elastic.Documentation.Search.Contract as zero-dep library


reakaleek (commented):
Wanted to share a different read on the motivation before we collapse the projects.

  The split was never about external reuse. None of us expected another repo to consume `Api.Core`. It was there to enforce the
  Clean Architecture dependency rule (Ports & Adapters): source dependencies point in one direction only, `App` -> `Infrastructure`
  -> `Core`, and the use cases never depend on low level details like Elasticsearch, AWS, HTTP or OTel exporters.

  A few reasons that matters:

  1. Dependency Inversion at the boundary. Core defines ports (`IAskAiGateway`, `IFullSearchGateway`, `IDistributedCache`) and the
  use cases depend on those interfaces. Infrastructure provides the adapters. Core owns the contract, Infrastructure conforms to it.
   The moment Core can see Infrastructure types, the arrow flips and the use case starts depending on the vendor.

  2. Stable depends on stable. The use cases (ask AI, search navigation, record feedback) are the most stable thing we have. The
  Elasticsearch client, DynamoDB SDK, GCP token provider and OTel exporter are the least stable, they version, get swapped, get
  reconfigured. The current layering means churn in the volatile layer can't ripple back into the stable one. Merge them and any
  infra change is free to touch a use case.

  3. Framework independence. Core doesn't reference `Microsoft.AspNetCore.App`, the ES client, AWS SDK or any OTel package, only
  logging and config abstractions. That's what lets us reason about and unit test the use cases without a web host or a live ES.
  Once everything is `Microsoft.NET.Sdk.Web` in one assembly, that property is gone.

  4. The compiler is the enforcement mechanism. Project boundaries are checked by the build, folders are checked by reviewers on a
  good day. "Core must not reference Infrastructure" is currently impossible to violate. After the merge it becomes a convention
  that decays the first time someone is in a hurry, and the symptom (a use case `new`-ing an `ElasticsearchClient`) won't show up
  until it has already shipped.

  On Mcp.Remote, agreed the current shape is awkward, but I'd read it as evidence for the split rather than against it. The only
  thing pulling Infrastructure into Mcp.Remote is the OTel helper. The search interfaces it uses are already satisfied by the tiny
  Core assembly, because the ports live in Core. Your PR already does the right thing for that piece by lifting `AddEuidEnrichment`
  into `ServiceDefaults` and moving the gateway interfaces into `services/Elastic.Documentation.Search`. Both moves are good on
  their own, and after them Mcp.Remote drops its Infrastructure reference without needing the Api projects to merge. We could ship
  those two extractions and leave Core/Infrastructure/App intact: Mcp.Remote ends up exactly as clean as the PR description targets,
   the Api keeps its compile time guardrail, and the dependency rule still holds.

  Happy to split the OTel/search extractions out as their own PR if that's useful, those wins don't require deleting the boundary.

Mpdreamz:
Appreciate the deep read — and I think we're closer than it looks, just optimizing on different axes.

The shape I'm working toward is **vertical slice architecture**, not the merge-and-flatten you might be reading PR1 as. Concretely:

- **`App` (the merged Api project) IS the infrastructure layer.** Composition root, web host, adapters, all the stuff that bridges HTTP into the rest of the system. That's its job.
- **Implementations live in service assemblies under `src/services/`.** Already the case for search (`services/Elastic.Documentation.Search` owns interfaces *and* impls). Each service is a vertical slice — one cohesive feature, all its parts together.
- **Service projects pragmatically co-locate the interface with the impl.** When external reuse appears, we lift just the interface into a Contracts assembly. We don't pre-split for hypothetical sharers.
- **Cross-cutting infrastructure that's actually reused ends up in `Elastic.Documentation.ServiceDefaults`.** Telemetry, health checks, OTel plumbing — exactly the kind of thing your PR1 feedback pushed me to extract from `Api.Infrastructure`, and rightly so.

On testability specifically:

> that's what lets us reason about and unit test the use cases without a web host or a live ES

I don't think this depends on the project split. Tests depend on services — most of which do contain implementations, that's the practical reality — but the unit-test seam is the interface. We can mock `IFullSearchService` without a web host or a live ES regardless of which assembly its impl ships in. Framework-independent testability is a property of the abstraction (the interface), not of an enforced assembly boundary around it.

The thing your model gives up that mine doesn't is the **compiler firewall**: "use case cannot accidentally `new ElasticsearchClient()`" goes from build error to reviewer-enforced. That's a real cost, but in a single-executable Web app with one `MappingsExtensions.cs` funnel, the firewall hasn't been catching anything that PR review wouldn't. And the price was three types per feature (`IGateway` + `Gateway` + `Usecase`) with one impl in sight everywhere — the Usecase layer was a second interface dressed as a class.

I think both our shapes honor the load-bearing tenets of Clean Architecture — dependency inversion at the abstraction, framework-independent contracts, testable units, technology-free domain. We just optimize for different things. Your model optimizes for **intra-app vendor isolation** with the compiler as the enforcement mechanism. Mine optimizes for **cross-repo sharing** with the abstraction (interface or domain record) as the enforcement mechanism, and `Contracts` / `services/*` as the deployment boundary where reuse actually happens.

The concrete goal driving this is sharing our domain and service interfaces with `elastic/website-search-data` in a way that **does not bleed implementations across**. `Contracts` ships `DocumentationDocument`, `IndexedProduct`, slimmed `ApplicableTo` POCOs — no YAML, no diagnostics, no Elasticsearch client, no AspNetCore. The Search service interface can travel the same way when that repo needs it. The implementations stay home in `services/Elastic.Documentation.Search` and the App. That's the boundary that earns its keep, and it's the one the Core/Infrastructure/App split wasn't placing.

Mpdreamz:
One quick factual clarification on the Mcp.Remote framing:

> The search interfaces it uses are already satisfied by the tiny Core assembly, because the ports live in Core.

`IFullSearchGateway` *lived* in `Api.Core`, but Mcp.Remote was only ever reaching through it to get at the search service. The actual implementation has been in `services/Elastic.Documentation.Search` the whole time, and Mcp.Remote already referenced that project to instantiate it. The Api.Core dependency wasn't satisfying a real abstraction need — it was an artifact of the port being placed in Core under the horizontal-layering model when it logically belonged with its implementation.

PR1 step 2 just relocated the interface to live next to its impl (`services/Search`), so Mcp.Remote consumes one project for one vertical slice — the search service — instead of one project for the interface and another for the impl. That's the vertical slice argument restated: Mcp.Remote was already depending on the search *service*, not on a Core abstraction layer. Reading the Mcp.Remote shape as evidence for the split assumes the port was load-bearing for the consumer; in practice the consumer never cared.

Mpdreamz:
One more on this specifically:

> The search interfaces it uses are already satisfied by the tiny Core assembly, because the ports live in Core.

This isn't quite accurate. Mcp.Remote never depended on the `Gateway` or `Usecase` abstractions in Api.Core — its consumption pattern was always service-shaped. It reached for the search *service*, and that service has always lived in `services/Elastic.Documentation.Search`, which Mcp.Remote had a project reference to anyway (it has to, to instantiate anything). The `IFullSearchGateway` port lived in Api.Core only because the horizontal-layering model placed ports there; the actual implementation has always been one project away in services/Search.

So the pre-PR1 reality: Mcp.Remote referenced both Api.Core (for the interface) and services/Search (for the impl that satisfies it). One vertical concern, two projects. PR1 step 2 just moved the interface to live next to its impl — the same place Mcp.Remote was already reaching for code.

And critically: Mcp.Remote's consumption stays identical across PR1 *and* PR2. It depended on a service before, it depends on a service now (`IFullSearchService`). The Gateway and Usecase layers it never touched got collapsed; the service it actually uses remains. That's the vertical slice paying out — the consumer's coupling tracks the concept, not the layering.

Reading Mcp.Remote as evidence for the split assumes the Core port was load-bearing for that consumer. In practice it was an extra hop with no benefit — the impl wasn't swappable, the interface wasn't satisfied by anything other than what services/Search provides, and Mcp.Remote was never going to construct its own impl.

Mpdreamz:
@coderabbitai  review

Mpdreamz:
@reakaleek and I discussed offline. Ultimately it boils down to Simple Clean Architecture vs Vertical Slice (Clean) Architecture. 

Both solve the tenants of clean architecture, the change now to VSA was mostly precipicated because we already follow it for most or our `src/services` and that `elastic/website-search-data` only needs interfaces as contracts, not a paradigm through gateways and usecases which in this particular sharing story is not strictly needed.

reakaleek (approved):
Agreed in live call

PR #3342, opened 2026-05-18
Fix ShouldContainHtml expected comparison

### LINKED ISSUE #3337: `ShouldContainHtml` always passes due to typo in `PrettyHtmlExtensions.cs`
In [`PrettyHtmlExtensions.cs`](https://github.com/elastic/docs-builder/blob/main/tests/Elastic.Markdown.Tests/PrettyHtmlExtensions.cs#L84), `expectedCompare` is assigned from `actual` instead of `expected`:

```csharp
var actualCompare = actual.Replace("\t", string.Empty);
var expectedCompare = actual.Replace("\t", string.Empty); // should be `expected`
```

This means the contains check on line 88 is `actual.Contains(actual)`, which always passes. Any test using `ShouldContainHtml` silently passes regardless of output.

[Many test assertions](https://github.com/search?q=repo%3Aelastic%2Fdocs-builder+ShouldContainHtml&type=code) across the test suite use this helper.

Found while investigating [#3324](https://github.com/elastic/docs-builder/pull/3324), where the `MagicCallOutWithFormatting` test passes despite the callout parser rejecting its input.

omribz156:
## Summary

Fixes `ShouldContainHtml` so the comparison uses the pretty-printed expected HTML instead of comparing the actual HTML to itself.

Adds a regression test that verifies the helper throws when the expected HTML fragment is missing from the actual output.

Closes #3337.

## Verification

- `git diff --check`
- `dotnet test ./tests/Elastic.Markdown.Tests/Elastic.Markdown.Tests.csproj --filter FullyQualifiedName~PrettyHtmlExtensionsTests --no-restore` could not run locally because this checkout pins .NET SDK `10.0.100` and no .NET SDK is installed on this machine.

This was implemented with Codex assistance, with the patch kept focused and manually reviewed.

### DIFF
--- a/tests/Elastic.Markdown.Tests/PrettyHtmlExtensions.cs
+++ b/tests/Elastic.Markdown.Tests/PrettyHtmlExtensions.cs
@@ -81,7 +81,7 @@ public static class PrettyHtmlExtensions
 		actual = actual.Trim('\n').PrettyHtml(sanitize);
 
 		var actualCompare = actual.Replace("\t", string.Empty);
-		var expectedCompare = actual.Replace("\t", string.Empty);
+		var expectedCompare = expected.Replace("\t", string.Empty);
 
 		// we compare over unindented HTML, but if that fails, we rely on the pretty HTML Contain().
 		// to throw for improved error messages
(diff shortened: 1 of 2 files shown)


Mpdreamz:
Thanks for bringing this to our attention! 

I followed up with https://github.com/elastic/docs-builder/pull/3346 which also addresses the 10 tests that were silently passing because of this.

PR #3349, opened 2026-05-19
Extract Elastic.Documentation.Search.Contract as zero-dep library

Mpdreamz:
## Why

`website-search-data` needs to depend on `DocumentationDocument` and the Elasticsearch mapping config to interoperate with docs-builder indices, but pulling in the full `Elastic.Documentation` assembly drags in YAML deserialization, AppliesTo parsing, link registries, and other heavyweight concerns. Without this extraction there is no safe, minimal target to depend on.

## What

A new `Elastic.Documentation.Search.Contract` project contains `DocumentationDocument`, `ParentDocument`, `IndexedProduct`, `DocumentationMappingConfig`/`LexicalConfig`/`SemanticConfig`/`DocumentationAnalysisFactory`, and a new wire-format `AppliesToEntry` record. The project has zero `ProjectReference`s and a single NuGet dependency (`Elastic.Mapping`), so external consumers can reference it without inheriting anything from docs-builder itself.

`Elastic.Documentation` gains a `ProjectReference` to the contract, so every existing internal consumer (Markdown exporter, ApiExplorer, search service, Assembler, Mcp.Remote) gets the types transitively with no new project references on their end.

## How

`DocumentationDocument.Applies` was previously typed as `ApplicableTo?` — a rich domain type from the AppliesTo subsystem — which would have pulled that subsystem into the contract. The property is now typed as `IReadOnlyCollection<AppliesToEntry>?`, where `AppliesToEntry` is a flat wire-format record matching the JSON shape that `ApplicableToJsonConverter` already produced. The conversion is centralized in `ApplicableTo.ToAppliesTo()`, ensuring the lifecycle/sub-type mapping lives in exactly one place. The indexed JSON shape in Elasticsearch is unchanged.


Mpdreamz:
Reopening against correct base branch (refactor/organize-doc-namespaces)

PR #3354, opened 2026-05-19
Keep nav-v2 focused on the active section

theletterf:
Keep nav-v2 focused on the active section by collapsing folders that are outside the current page's branch. Without this, same-section HTMX navigation can leave stale folders expanded and make the sidebar look like multiple sections are active at once.

Demo: https://docs-v3-preview.elastic.dev/elastic/docs-builder/docs/3354/deploy-manage/distributed-architecture/clusters-nodes-shards/node-roles

## Why

The nav-v2 prototype is meant to keep the current section focused, but once you open another branch it can stay expanded after navigating within the same section. That leaves the sidebar with multiple open branches and makes it harder to tell where the current page belongs.

## What

This change recomputes folder expansion from the active page path and closes every folder that is not on that branch. It also uses the rendered folder hierarchy for manual accordion behavior, so opening one folder closes its siblings at the same level, and adds a focused frontend regression test for both behaviors.

### DIFF
--- a/src/Elastic.Documentation.Site/Assets/pages-nav-v2.test.ts
+++ b/src/Elastic.Documentation.Site/Assets/pages-nav-v2.test.ts
@@ -10,72 +10,74 @@ jest.mock('tippy.js', () => ({
 
 function renderNav() {
     document.body.innerHTML = `
-        <nav class="docs-sidebar-nav-v2" data-nav-v2>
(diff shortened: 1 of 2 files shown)


marciw (commented):
That's looking great!

Only thing I'm wondering about: if I click on something at the bottom of the nav, should the nav reposition / scroll up so that I can see the full selected section? will dm you a screen recording

PR #3356, opened 2026-05-19
Add MCP and API smoke tests to the integration suite on CI

Mpdreamz:
## Why

Until now, the integration suite only validated that the docs assembler could clone, build, and serve content — it had no coverage of the two HTTP services (`api` and `remote-mcp`) that actually power search and the MCP tooling in production. We had no automated signal that a change broke the API response shape, the MCP tool list, or the wiring between Aspire and those services.

This PR closes that gap by making the Aspire fixture wait for the API and MCP resources to become healthy before running tests, then adding smoke tests that exercise real endpoints against a real Elasticsearch index.

## What changed and why

**New smoke tests** (`Smoke/ApiSmokeTests`, `Smoke/McpSmokeTests`) — prove the Aspire-launched services are reachable and returning sensible responses. Search and changes assertions are skipped gracefully when the index is empty so the tests stay green on a cold environment.

**`Elastic.Assembler.IntegrationTests` → `Elastic.Documentation.IntegrationTests`** — the old name implied the tests only covered the assembler. They now cover the whole stack.

**`Elastic.Documentation.Api.IntegrationTests` → `tests/Elastic.Documentation.Api.Tests`** — those tests use `WebApplicationFactory` with mocked services; they're unit tests that happen to test an ASP.NET app, not integration tests. Moving them to `tests/` means they run in `./build.sh unit-test` (fast, every PR) instead of the slow Aspire suite.

**AppHost: skip Docker when not needed** — previously Aspire registered an Elasticsearch container unconditionally and just marked it `WithExplicitStart()`. Aspire v13 still health-checks Docker for all container resources, even unstarted ones, which crashed the AppHost on CI and any machine without Docker running. Now the container is only registered when `--start-elasticsearch` is passed; using an external ES URL skips it entirely.

**Port conflict fix** — both `Elastic.Documentation.Api` and `Elastic.Documentation.Mcp.Remote` hardcoded port 8080. Aspire injects `ASPNETCORE_HTTP_PORTS` when it manages endpoints, so both services now skip the hardcoded port when that variable is set.

**CI wiring** — the `integration` job fetches dedicated readonly Elasticsearch credentials from AWS SSM (via GitHub OIDC, no long-lived secrets) and injects them as dotnet user secrets so the Aspire AppHost picks them up without any code changes.

## Test plan

- [ ] `dotnet build` — 0 warnings
- [ ] `./build.sh unit-test` — picks up `Elastic.Documentation.Api.Tests`
- [ ] `./build.sh integrate` — boots Aspire, smoke tests pass against prod ES data
- [ ] Running `dotnet run --project aspire` without Docker does not crash


.github/workflows/ci.yml
```diff
@@ -195,5 +199,20 @@ jobs:
...
+        with:
+          role-to-assume: arn:aws:iam::197730964718:role/elastic-docs-v3-integration-tests
+          aws-region: us-east-1
+
+      - name: Fetch ES credentials
+        run: |
+          ES_URL=$(aws ssm get-parameter --name /elastic-docs-v3/prod/docs-elasticsearch-github-actions-readonly-url --with-decryption --query Parameter.Value --output text)
+          ES_KEY=$(aws ssm get-parameter --name /elastic-docs-v3/prod/docs-elasticsearch-github-actions-readonly-api-key --with-decryption --query Parameter.Value --output text)
+          echo "::add-mask::$ES_URL"
+          echo "::add-mask::$ES_KEY"
```
reakaleek, line 213: 🚀

reakaleek (approved):

PR #3378, opened 2026-05-22
Sync EUI icon library with Jan–May 2026 renames and additions

florent-leborgne:
## Why

EUI renamed virtually all icons from camelCase to snake_case in January 2026 (elastic/eui#9279) and continued adding new icons through May 2026. Without this sync, docs-builder was serving outdated or missing icon glyphs and would break the build whenever a doc page used a renamed icon name.

## What

- **580 canonical icons** now match EUI main exactly — every root icon and all 56 token icons are in sync
- **97 icon aliases + 56 token aliases** added to `EuiSvgIcons` so that any existing doc page using an old camelCase name (e.g. `{icon}\`checkCircle\``, `{icon}\`errorFilled\``, `{icon}\`wordWrap\``) transparently resolves to the current EUI glyph without a build error — old SVG files deleted, aliases handle resolution
- **~200 new root SVGs** added (new icons like `chart_*`, `drag_*`, `chevron_*`, `ellipsis`, `server`, `significant_events`, etc.)
- **All 56 token SVGs** renamed from camelCase (`tokenAlias.svg`) to snake_case (`token_alias.svg`) to match EUI
- **`docs/syntax/icons.md`** regenerated from the actual files on disk (580 entries, alphabetical)

## How

The alias mechanism lives in `EuiSvgIcons.TryGetIcon` / `GetIcon` / `TryGetToken`. On lookup, the name is checked against a static `IconAliases` / `TokenAliases` dictionary first; if matched, the canonical name is substituted before hitting the embedded-resource dictionary. This means old doc pages keep working at zero maintenance cost, and future EUI renames only require adding an alias entry rather than a bulk docs update.


theletterf:
@florent-leborgne Out of curiosity, how many icon roles are we using throughout the docs presently? We could consider converting them to the new format using AI and then remove the aliases later on.

theletterf (approved):

florent-leborgne:
> We could consider converting them to the new format using AI and then remove the aliases later on.

Yes that's probably something we should do regularly, but I think we still need the logic to handle transitions (release change in docs builder, then clean up, etc.)

PR #3407, opened 2026-05-27
Hide Supported on line for settings whose stack is fully planned

### LINKED ISSUE #3408: Hide "Supported on" line on {settings} entries whose stack is fully planned
## Problem

In the `{settings}` directive, when a setting's `applies_to.stack` only points at versions greater than the current released stack, the inline stack badge correctly renders as **Planned**, but the **Supported on** line can still list `ECH`, `Self-managed`, etc. as if the setting were available today. This is misleading: ECH, ECE, ECK, and Self-managed all run the Elastic Stack, so a setting that has not shipped in any released stack version cannot actually be used on those surfaces.

## Example

The [Internationalization settings in Kibana](https://www.elastic.co/docs/reference/kibana/configuration-reference/internationalization-settings) page shows `i18n.defaultLocale` with:

- inline badge: **Stack | Planned**
- supported on: **ECH, Self-managed**

A user tried to set this on ECH and it did not work — the setting is [planned for Kibana 9.5.0](https://github.com/elastic/kibana/pull/260835). Discussion: [Slack thread](https://elastic.slack.com/archives/C0JF80CJZ/p1779885675811119).

## Proposed behaviour

Hide the **Supported on** line for any individual setting whose `applies_to.stack` is non-null and where every entry has a `Version.Min` greater than the current stack version. Settings with mixed past + future stack entries (e.g. `stack: ga 9.0, deprecated 9.5`) are still usable today, so the line stays. Settings with no `stack:` at all are unaffected.

## Known edge case

A setting with `stack: ga 9.5` AND `serverless: ga` would also have its Supported on line hidden, even though it exists on serverless today. Per the Slack thread, the settings corpus is published from Kibana `main` precisely because serverless usually runs ahead of stack, so this combination is not theoretical. For v1 we hide; if pages surface where this loses useful info we can narrow the rule (e.g. only hide when the would-be Supported-on badges are all deployment-stack-bound: ECH, ECE, ECK, Self-managed).

## Related

PR: #3407

florent-leborgne:
## Why

In the `{settings}` directive, when a setting's `applies_to.stack` only points at versions greater than the current released stack, the inline stack badge renders as "Planned" but the "Supported on" line could still claim ECH and Self-managed availability today. Readers see e.g. `i18n.defaultLocale` marked **Stack | Planned** while also listing **Supported on: ECH, Self-managed**, which is misleading: ECH, ECE, ECK, and Self-managed all run the Elastic Stack and cannot expose a setting that has not shipped yet.

Closes #3408.

## What

Suppresses the "Supported on" line for any individual setting whose `applies_to.stack` is non-null and where every entry has a `Version.Min` greater than the current stack version. A setting with mixed past+future stack entries (e.g. `stack: ga 9.0, deprecated 9.5`) is still usable today, so the line stays. Settings with no `stack:` at all are unaffected.

Implementation: a new `SettingsViewModel.IsStackFullyPlanned(...)` helper compares each `Stack` entry's `Version.Min` against `VersionsConfig.GetVersioningSystem(VersioningSystemId.Stack).Current`, and `SettingsView.cshtml` extends its `showSupportedOn` predicate with that check.

Three new tests cover the positive case (fully-planned), the released case, and the mixed case. One existing fixture that used `stack: ga 9.2` only to demonstrate badge rendering was switched to `stack: ga 7.0` so its positive `Should().Contain("settings-supported-on")` assertion still holds under the new rule.

### Known edge case

A setting with `stack: ga 9.5` AND `serverless: ga` will also have its "Supported on" line hidden, even though it does exist on serverless today. Per the [Slack thread](https://elastic.slack.com/archives/C0JF80CJZ/p1779885675811119), the settings corpus is published from Kibana `main` precisely because serverless usually runs ahead of stack, so this combination is not theoretical. If this comes up in practice we can narrow the rule to only hide when the would-be Supported-on badges are all deployment-stack-bound.

### DIFF
--- a/tests/Elastic.Markdown.Tests/SettingsInclusion/IncludeTests.cs
+++ b/tests/Elastic.Markdown.Tests/SettingsInclusion/IncludeTests.cs
@@ -86,7 +86,7 @@ groups:
         datatype: object
         default: "[]"
         applies_to:
-          stack: ga 9.2
+          stack: ga 7.0
         options:
           - option: strict
             description: Strict mode.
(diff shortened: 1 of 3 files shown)


tests/Elastic.Markdown.Tests/SettingsInclusion/IncludeTests.cs
```diff
@@ -388,3 +388,125 @@ public void RendersAppliesToRoleAsBadgeNotLiteralText()
...
+	{
+		// language=yaml
+		fileSystem.AddFile("docs/_settings/stack-fully-planned.yml", """
+groups:
+  - group: Example
+    settings:
+      - setting: i18n.defaultLocale
+        description: Locale setting planned for a future stack version.
+        applies_to:
+          stack: ga 9.5
```
florent-leborgne:
```suggestion
          stack: ga 99.99
```
maybe better this way?

tests/Elastic.Markdown.Tests/SettingsInclusion/IncludeTests.cs
```diff
@@ -388,3 +388,125 @@ public void RendersAppliesToRoleAsBadgeNotLiteralText()
...
+	{
+		// language=yaml
+		fileSystem.AddFile("docs/_settings/stack-mixed-versions.yml", """
+groups:
+  - group: Example
+    settings:
+      - setting: xpack.sample.mixed
+        description: GA since 7.0, scheduled for deprecation in 9.0.
+        applies_to:
+          stack: ga 7.0, deprecated 9.0
```
florent-leborgne:
```suggestion
          stack: ga 7.0, deprecated 99.0
```

tests/Elastic.Markdown.Tests/SettingsInclusion/IncludeTests.cs
```diff
@@ -388,3 +388,125 @@ public void RendersAppliesToRoleAsBadgeNotLiteralText()
...
+{
+	protected override void AddToFileSystem(MockFileSystem fileSystem)
+	{
+		// language=yaml
+		fileSystem.AddFile("docs/_settings/stack-mixed-versions.yml", """
+groups:
+  - group: Example
+    settings:
+      - setting: xpack.sample.mixed
+        description: GA since 7.0, scheduled for deprecation in 9.0.
```
florent-leborgne:
```suggestion
        description: GA since an existing version, scheduled for deprecation in a future version.
```

theletterf (approved):

PR #3409, opened 2026-05-27
Delete stale CloudFront redirect KVS entries

cotti:
## Why

The redirect KVS sync has had a one-line bug since #1503 (July 2 2025) that prevented any redirect from ever being deleted from the prod/staging/preview/edge CloudFront KeyValueStores. Operand order on `Except` was flipped so the delete batch resolved to "keys we are about to PUT" (always a no-op) instead of "keys already in the KVS that no longer appear in the new redirects file".

This surfaced as a writer-visible bug today: elastic/docs-content#5818 added a bad redirect for `/.../azure-native-isv-service`, and two follow-up content PRs (#6667, #6716) could not undo it via `redirects.yml`. Every redirect ever pushed since the regression has accumulated; the prod KVS had 806 keys at the time of investigation.

The bad Azure ISV entry was manually purged from the prod KVS to unblock the writer, but the underlying bug needs fixing so future removals from `redirects.yml` actually propagate to CloudFront.

## What

- Move the put/delete diff calculation into `RedirectKvsDiff`, a pure static helper with no AWS dependencies, and swap the `Except` operands so stale KVS keys are flagged for deletion.
- Add a wipe guard (`WouldWipeAllExisting`): refuse to push an empty `redirects.json` against a populated KVS. Before this fix the bug masked this risk; after the fix an empty sourced file would wipe every redirect, which is almost certainly a broken build rather than intent.
- Cover the regression with `RedirectKvsDiffTests` — including `ComputeBatchUpdates_AzureIsvRegression_ScenarioEndToEnd`, which pins the exact "stale key in KVS, dropped from sourced file, must end up in `toDelete`" semantics that broke here.
- Log the computed batch sizes at info level so future deploys make the diff visible in CI output.

After this lands, the next prod deploy will also clean up other stale entries that have accumulated since the regression. A read-only audit of those is being shared separately.

### DIFF
--- a/src/services/Elastic.Documentation.Assembler/Deploying/Redirects/AwsCloudFrontKeyValueStoreProxy.cs
+++ b/src/services/Elastic.Documentation.Assembler/Deploying/Redirects/AwsCloudFrontKeyValueStoreProxy.cs
@@ -37,13 +37,17 @@ public class AwsCloudFrontKeyValueStoreProxy(IDiagnosticsCollector collector, IL
 		if (!listingSuccessful)
 			return;
 
-		var toPut = sourcedRedirects
-			.Select(kvp => new PutKeyRequestListItem { Key = kvp.Key, Value = kvp.Value })
-			.ToArray();
-		var toDelete = sourcedRedirects.Keys
-			.Except(existingRedirects)
-			.Select(k => new DeleteKeyRequestListItem { Key = k })
-			.ToArray();
+		if (RedirectKvsDiff.WouldWipeAllExisting(sourcedRedirects, existingRedirects))
+		{
+			Collector.EmitError("", $"Refusing to update redirects: sourced redirects are empty but the KVS contains {existingRedirects.Count} entries. " +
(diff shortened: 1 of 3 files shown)


reakaleek:
I feel like we should get a list of redirects that will be deleted, and check if these URLs are in use. Can you do an analysis of that?

Mpdreamz (approved):
LGTM. The operand fix is clearly correct, the wipe guard adds a solid safety layer, and the test suite thoroughly pins the regression.

One minor follow-up worth tracking: `ComputeBatchUpdates` currently re-puts every sourced entry on every deploy, even entries whose value hasn't changed. This burns unnecessary CloudFront KVS API quota on large redirect sets. A future optimisation could skip the no-op puts by filtering `sourcedRedirects` down to entries where the key is absent from `existingRedirects` or the value differs. The `ComputeBatchUpdates_KeyInBoth_IsPutAndNotDeleted` test already pins this behavior so any change there would be caught.

PR #3421, opened 2026-06-01
Add zizmor GitHub Actions security scanning

Mpdreamz:
## Why

GitHub Actions workflows were not statically audited for common misconfigurations (unpinned third-party actions, credential persistence, overly broad permissions, and related supply-chain risks). Without automated checks, regressions are easy to miss in review.

## What

Adds a zizmor CI workflow and `.github/zizmor.yml` policy: `actions/*` and `elastic/*` may use ref pins; all other actions must be hash-pinned. Applies remediations across existing workflows (third-party SHA pins, scoped permissions, Dependabot cooldown, checkout credential hygiene, and targeted ignores for intentional `pull_request_target` usage) so `zizmor .` passes locally and in CI.


reakaleek (approved):

.github/dependabot.yml
```diff
@@ -9,6 +9,8 @@ updates:
       time: '08:00'
     labels:
       - chore
+    cooldown:
+      default-days: 7
```
reakaleek: let's even do 14..

I thought I already did this.

PR #3436, opened 2026-06-01
changelog: render CDN-hosted bundles in the {changelog} directive

cotti:
## Why

Scrubbed changelog bundles live in the public S3 / CloudFront bucket, but the `{changelog}` directive could previously only render bundles checked into the same repo. That meant a docset couldn't show another product's changelog without vendoring its bundle files, and there was no manifest to discover bundles without listing S3.

## What

- **Producer:** when a bundle is uploaded, publish/update a per-product `registry.json` manifest in the public bucket (carried through by the scrubber's pass-through copy). Manifest writes use S3 conditional PUTs (`If-Match`/`If-None-Match`) with bounded retries so parallel uploads merge safely instead of clobbering each other.
- **Consumer:** a new `:cdn: <product>` option on the `{changelog}` directive fetches that product's `registry.json` plus the referenced bundles from CloudFront and renders them inline. A new `:version:` option filters to a single version (matched against bundle target or filename); in CDN mode it only downloads the matching bundle.
- **Naming:** the per-product manifest is `registry.json`. "registry-index" is reserved for a possible future bucket-wide meta-registry, so the producer/consumer types and keys were renamed accordingly.

## How

CDN fetching happens inside the synchronous Markdig finalize pass, so `CdnChangelogFetcher` does a synchronous HTTP fetch backed by a static in-memory cache (keyed by base URL + product + version) and the bundles it returns are self-contained. The base URL defaults to the production CloudFront domain and can be overridden via `DOCS_BUILDER_CHANGELOG_CDN` for testing.


Mpdreamz:
Quick ponder:

```
:cdn: <product>
```

Can that be optional? For most cases we can infer the product from the repository that holds the doc?

cotti:
> Quick ponder:
> 
> ```
> :cdn: <product>
> ```
> 
> Can that be optional? For most cases we can infer the product from the repository that holds the doc?

Ah, yes. Added inferring for the product as default.

src/Elastic.Markdown/Myst/Directives/Changelog/ChangelogBlock.cs
```diff
@@ -445,9 +493,41 @@ private void LoadAndCacheBundles()
...
+		}
+
+		var fetcher = new CdnChangelogFetcher(NullLoggerFactory.Instance, Build.ReadFileSystem);
+		var loadedBundles = fetcher.Fetch(baseUri, product, VersionFilter, msg => this.EmitError(msg), msg => this.EmitWarning(msg), Cancel.None);
+
+		ApplyLoadedBundles(new BundleLoader(Build.ReadFileSystem), loadedBundles);
+		Found = LoadedBundles.Count > 0;
+	}
+
+	private void ApplyLoadedBundles(BundleLoader loader, IReadOnlyList<LoadedBundle> loadedBundles)
```
technige: **Suggestion:** `loader` is declared but never used inside `ApplyLoadedBundles` — looks like it was left over from the refactoring. Would it be okay to drop the parameter? The CDN call site on line 520 constructs a fresh `BundleLoader` just to satisfy the signature right now.

src/Elastic.Markdown/Myst/Directives/Changelog/ChangelogBlock.cs
```diff
@@ -197,6 +211,29 @@ public override void FinalizeAndValidate(ParserContext context)
...
+				: InferCdnProductFromRepository();
+
+			if (string.IsNullOrWhiteSpace(product))
+			{
+				this.EmitError(
+					"The :cdn: product could not be inferred from the repository; specify it explicitly, e.g. ':cdn: elasticsearch'.");
+				return;
+			}
+
+			CdnProduct = product;
```
technige, line 237: **Suggestion:** `CdnProduct` is assigned here, before `LoadCdnBundles` has had a chance to validate the product name. That means an invalid product like `"invalid$product"` ends up stored on the block even when an error is emitted (the test `CapturesCdnProductOption` documents this). Would it be cleaner to move `IsValidCdnProduct` check up here and only assign `CdnProduct` after it passes?

src/Elastic.Documentation.Configuration/ReleaseNotes/CdnChangelogFetcher.cs
```diff
@@ -0,0 +1,212 @@
...
+/// </para>
+/// </remarks>
+public sealed class CdnChangelogFetcher(ILoggerFactory logFactory, IFileSystem fileSystem, HttpMessageHandler? handler = null)
+{
+	private const int SupportedSchemaVersion = 1;
+
+	private static readonly ConcurrentDictionary<string, IReadOnlyList<LoadedBundle>> Cache = new(StringComparer.Ordinal);
+
+	private readonly ILogger _logger = logFactory.CreateLogger<CdnChangelogFetcher>();
+	private readonly HttpClient _httpClient = handler is null ? new HttpClient() : new HttpClient(handler, disposeHandler: false);
```
technige: **Suggestion:** `HttpClient` is created per `CdnChangelogFetcher` instance but neither the fetcher nor `ChangelogBlock` implements `IDisposable`, so each instance leaks a socket handle. In practice the static cache means most fetcher instances never actually use `_httpClient`, but it's still a resource leak. Would it be worth either making `CdnChangelogFetcher : IDisposable` (and disposing after `LoadCdnBundles` returns), or sharing a static/injected `HttpClient` the way `CrossLinkFetcher` probably does?

src/Elastic.Markdown/Myst/Directives/Changelog/ChangelogBlock.cs
```diff
@@ -445,9 +493,41 @@ private void LoadAndCacheBundles()
...
+
+		if (ResolveCdnBaseUri() is not { } baseUri)
+		{
+			this.EmitError(
+				$"No valid changelog CDN base URL is configured. Set the {CdnBaseUrlEnvironmentVariable} environment variable to an absolute http(s) URL.");
+			return;
+		}
+
+		var fetcher = new CdnChangelogFetcher(NullLoggerFactory.Instance, Build.ReadFileSystem);
+		var loadedBundles = fetcher.Fetch(baseUri, product, VersionFilter, msg => this.EmitError(msg), msg => this.EmitWarning(msg), Cancel.None);
```
technige: **Nit:** `Cancel.None` here means a build cancellation can't interrupt an in-flight CDN fetch. I understand this is a constraint of the synchronous Markdig pipeline — would a short comment be useful for the next reader so the choice doesn't look accidental?

src/services/Elastic.Changelog/Uploading/RegistryBuilder.cs
```diff
@@ -0,0 +1,315 @@
...
+	private static List<RegistryBundle> Merge(
+		IReadOnlyList<RegistryBundle> existing,
+		IReadOnlyList<RegistryBundle> incoming)
+	{
+		var byFile = existing.ToDictionary(b => b.File, b => b, StringComparer.Ordinal);
+		foreach (var entry in incoming)
+			byFile[entry.File] = entry;
+
+		return byFile.Values
+			.OrderByDescending(b => b.Target ?? string.Empty, StringComparer.Ordinal)
```
technige: **Question:** `StringComparer.Ordinal` sorts by string bytes, so `"9.10.0"` sorts before `"9.9.0"` (lexicographic). The docs say the manifest is "sorted newest first" — is ordinal close enough for the expected range of version strings, or would using the same `VersionOrDate.Parse` comparator the consumer uses be worth it here? (The consumer re-sorts anyway, so this only affects how `registry.json` looks on disk.)

tests/Elastic.Changelog.Tests/Uploading/RegistryBuilderTests.cs
```diff
@@ -0,0 +1,381 @@
...
+		var result = await _builder.RefreshAsync(_collector, targets, TestContext.Current.CancellationToken);
+
+		result.Updated.Should().Be(1);
+		A.CallTo(() => _s3Client.GetObjectAsync(A<GetObjectRequest>._, A<CancellationToken>._))
+			.MustHaveHappenedTwiceExactly();
+		A.CallTo(() => _s3Client.PutObjectAsync(A<PutObjectRequest>._, A<CancellationToken>._))
+			.MustHaveHappenedTwiceExactly();
+
+		// The successful (second) PUT used the re-read ETag and merged both the concurrent and local entries.
+		_puts.Should().ContainSingle();
```
technige, line 360: **Nit:** `_puts.Should().ContainSingle()` here is checking the *captured* list, which only holds the PUT that didn't throw — the first (failed) call throws before `_puts.Add` runs. That's correct, but it's easy to misread as "only one PUT was attempted". A short comment like `// first PUT threw, so only the successful retry is captured` might save the next reader a double-take.

tests/Elastic.Markdown.Tests/Directives/ChangelogBasicTests.cs
```diff
@@ -192,6 +292,168 @@ public void RendersContent()
...
+				- title: Faster vector search on the CDN
+				  type: enhancement
+				  products:
+				  - product: cdn-render-test
+				    target: 9.4.0
+				  prs:
+				  - "999"
+				""")
+		], _ => { });
+		CdnChangelogFetcher.PrimeCacheForTesting(new Uri(ChangelogBlock.DefaultCdnBaseUrl), Product, null, bundles);
```
technige: **Nit:** `PrimeCacheForTesting` takes `new Uri(ChangelogBlock.DefaultCdnBaseUrl)`, which ties the test to the production constant being `internal`. That's a reasonable tradeoff given the synchronous-pipeline constraint, but would a brief comment here explaining *why* the production default URL has to match help future test authors avoid confusion?

tests/Elastic.Markdown.Tests/Directives/ChangelogBasicTests.cs
```diff
@@ -192,6 +292,168 @@ public void RendersContent()
...
+public class ChangelogCdnInvalidProductTests(ITestOutputHelper output) : DirectiveTest<ChangelogBlock>(output,
+	// language=markdown
+	"""
+	:::{changelog}
+	:cdn: invalid$product
+	:::
+	""")
+{
+	[Fact]
+	public void CapturesCdnProductOption() => Block!.CdnProduct.Should().Be("invalid$product");
```
technige: **Nit:** This asserts `CdnProduct == "invalid$product"` even though the product fails validation. It accurately documents current behaviour, but it made me wonder whether `CdnProduct` should remain unset when the product is invalid (i.e., only assigned after `IsValidCdnProduct` passes). Leaving here as a question in case that's worth revisiting alongside the suggestion above.

technige (commented):
Nice feature overall — the optimistic-concurrency design and the security guards are solid. A few things I'd suggest looking at before merging.

lcawl:
I'm happy to test this out if a `registry.json` is added to the S3 bucket (e.g. for the cloud-serverless bundles). If that won't occur until after this PR is merged, just let me know and I'll happily start using it then (so I don't have to keep generating markdown output for serverless release docs).

docs/syntax/changelog.md
```diff
@@ -161,6 +163,51 @@ Explicit path to a `changelog.yml` or `changelog.yaml` configuration file, relat
...
+
+Sources bundles for a single **product** from the public changelog CDN instead of a local folder, so a docset can render release notes without vendoring bundle YAML.
+
+```markdown
+:::{changelog}
+:cdn: elasticsearch
+:::
+```
+
+The value names the product (must match `[a-zA-Z0-9_-]+`) and maps to `{product}/registry.json` plus the bundles it lists on the CDN. The value is **optional**: leave it blank to infer the product from the repository that holds the doc (the common case where the repository name is the product id, for example the `elasticsearch` repo renders the `elasticsearch` product).
```
lcawl: > The value names the product (must match `[a-zA-Z0-9_-]+`) 

My assumption is that it must also match a valid value from https://github.com/elastic/docs-builder/blob/main/config/products.yml Is that correct? If so, IMO that should be mentioned here too.

docs/syntax/changelog.md
```diff
@@ -161,6 +163,51 @@ Explicit path to a `changelog.yml` or `changelog.yaml` configuration file, relat
...
+
+```markdown
+:::{changelog}
+:cdn:
+:::
+```
+
+Inference only works for repositories whose name matches the product id. Repositories that publish multiple products (for example `cloud`, which publishes `cloud-hosted`, `cloud-serverless`, and `cloud-enterprise`) must name the product explicitly; if a product cannot be inferred the block emits an error. When `:cdn:` is set, the local-folder argument is ignored. All other options (`:type:`, `:link-visibility:`, `:description-visibility:`, `:dropdowns:`, `:subsections:`) and `hide-features` apply identically to CDN-sourced bundles.
```
lcawl: IMO we should use public repos in the example
```suggestion
Inference only works for repositories whose name matches the product ID. For example, the `cloud-serverless` product release notes are published from the `docs-content` repo and therefore must specify the product explicitly. If a product cannot be inferred the block emits an error. When `:cdn:` is set, the local-folder argument is ignored. All other options (`:type:`, `:link-visibility:`, `:description-visibility:`, `:dropdowns:`, `:subsections:`) and `hide-features` apply identically to CDN-sourced bundles.
```

Mpdreamz (changes requested):
## Design question — should CDN sourcing be declared in `docset.yml`?

The `:cdn:` option is per-directive inline configuration, meaning CDN dependencies are discovered at render time rather than at startup. This is the same concern raised on #3470 about `changelog bundle`'s implicit CDN entry sourcing.

Compare how crosslinks work: repos are declared upfront in `docset.yml` under `cross_links`. docs-builder knows the external dependencies at startup, can fail fast if the index is unreachable, and fetches all link indexes concurrently before any page renders. The `{changelog}` directive gets none of that.

A docset-level declaration would look similar:

```yaml
# docset.yml
release_notes:
  - product: elasticsearch
  - product: kibana
```

With that:
- All registries and bundles are fetched once at startup, concurrently, before any page renders — rather than serially on first directive encounter.
- A CDN failure is a startup error with a clear config pointer, not a silent empty render mid-build.
- The `{changelog}` directive just references an already-loaded in-memory set, with no `HttpClient` calls at parse time.
- `:cdn:` could still exist as a convenience override, but `docset.yml` would be the primary declaration path.

The process-wide `ConcurrentDictionary` cache means a product is only fetched once per build, so the per-directive overhead is bounded. But the fail-fast and concurrency gaps remain. Is this the direction you want, or is there a reason to keep it directive-inline?

---

## Code issues

**`NullLoggerFactory.Instance` for CDN fetches** (`ChangelogBlock.cs` `LoadCdnBundles`): the fetcher is constructed with `NullLoggerFactory.Instance`, discarding all `ILogger` output (info-level fetch timing, debug-level cache hits). The block has access to build context — pass the real logger factory rather than null. CDN round-trip logs are exactly what you want when debugging a silent empty render.

**`repository == "unavailable"` sentinel check** (`InferCdnProductFromRepository`): this couples product inference to a specific sentinel string from the git information service. If that sentinel changes, inference silently fails with no diagnostic. A `null`/empty check or a typed signal from the git service would be safer.

**Multi-line XML doc comments on private methods** (`LoadCdnBundles`, `FilterByVersion`, `ResolveCdnBaseUri`, etc. in `ChangelogBlock.cs`): project style is one short line max on private methods, or none. Several new private methods have multi-sentence `<summary>` blocks.

---

## Minor

- **`ChangelogVersionMatch.Matches` — `null`/empty → match-all**: the non-nullable `string requested` parameter silently returns `true` when empty/whitespace. A short comment explaining the match-all semantic would help callers.
- **Test gap**: no test for `:cdn:` + `:version:` together in CDN mode (both are tested independently but not combined).

cotti:
> ## Design question — should CDN sourcing be declared in `docset.yml`?
> 
> The `:cdn:` option is per-directive inline configuration, meaning CDN dependencies are discovered at render time rather than at startup. This is the same concern raised on #3470 about `changelog bundle`'s implicit CDN entry sourcing.
> 
> Compare how crosslinks work: repos are declared upfront in `docset.yml` under `cross_links`. docs-builder knows the external dependencies at startup, can fail fast if the index is unreachable, and fetches all link indexes concurrently before any page renders. The `{changelog}` directive gets none of that.
> 
> A docset-level declaration would look similar:
> 
> ```yaml
> # docset.yml
> release_notes:
>   - product: elasticsearch
>   - product: kibana
> ```
> 
> With that:
> 
>     * All registries and bundles are fetched once at startup, concurrently, before any page renders — rather than serially on first directive encounter.
> 
>     * A CDN failure is a startup error with a clear config pointer, not a silent empty render mid-build.
> 
>     * The `{changelog}` directive just references an already-loaded in-memory set, with no `HttpClient` calls at parse time.
> 
>     * `:cdn:` could still exist as a convenience override, but `docset.yml` would be the primary declaration path.
> 
> 
> The process-wide `ConcurrentDictionary` cache means a product is only fetched once per build, so the per-directive overhead is bounded. But the fail-fast and concurrency gaps remain. Is this the direction you want, or is there a reason to keep it directive-inline?

Makes sense. I think adding it to docset also plays well with the intent of release notes being a product-centric entity in contrast with changelogs being a repo-level entity. The reasoning behind the initial approach was in allowing slighter quicker pacing, but with the benefit of hindsight we have consistently won more with tighter specs and RN already has a lot of input variety.

I'll make the changes needed around, and notify the current products involved.

cotti:
### Updated: declarative, opt-in CDN sourcing via `docset.yml` (addresses @mpdreamz's suggestion)

Reworked CDN changelog sourcing from dynamic, per-directive fetching into a **declarative, startup-time prefetch** model driven by `release_notes` in `docset.yml`, mirroring the existing `cross_links` mechanism.

**Schema**
- New product-keyed `release_notes:` list in `docset.yml`. Each id is validated against `products.yml` (must exist and carry the `release-notes` feature; underscores normalized to hyphens).

**New components** (`Elastic.Documentation.Configuration/ReleaseNotes`)
- `ChangelogCdn` — single source of truth for the CDN base URL (`DOCS_BUILDER_CHANGELOG_CDN`, default distribution).
- `IReleaseNotesResolver` (+ `NoopReleaseNotesResolver`), `FetchedReleaseNotes` (immutable prefetch result), and `ReleaseNotesFetcher` (async, concurrent per-product fetch, **strict fail-fast** on registry errors including 404).
- `CdnChangelogFetcher` refactored into a stateless async fetch engine (removed the static cache / shared `HttpClient`).

**Wiring**
- `IReleaseNotesResolver` threaded through `DocumentationSet` → `ParserContext`, prefetched at every entry point (isolated build, `serve`, codex, assembler), with a `Noop` default everywhere else.

**Directive**
- `{changelog}` `:cdn:` is now a **selector** over the prefetched set — no network at parse time.
- Valueless `:cdn:` infers the product from the repo, mapped to its canonical id via `products.yml` (e.g. `elastic-otel-java` → `edot-java`); multi-product repos must name the product explicitly.
- Removed the directive-local `ResolveCdnBaseUri` / env constants. An undeclared product now errors with guidance to add it under `release_notes:`.

**Docs / tests** — updated `syntax/changelog.md` and the registry dev doc; added directive CDN tests, async fetcher tests, and `release_notes` validation tests.

src/Elastic.Markdown/Myst/Directives/Changelog/ChangelogBlock.cs
```diff
@@ -463,6 +533,44 @@ private void LoadAndCacheBundles()
 		}
 	}
 
+	/// <summary>
+	/// Applies the optional <c>:version:</c> filter to loaded bundles. A bundle matches when its
+	/// target version or its file stem equals the requested value. If nothing matches a warning is
+	/// emitted and the block renders empty (rather than silently falling back to all versions).
+	/// </summary>
+	private IReadOnlyList<LoadedBundle> FilterByVersion(IReadOnlyList<LoadedBundle> bundles)
```
Mpdreamz, line 537: Project style is one short comment line max on private methods (or none). This 4-line summary on a private method should be shortened to a single line or removed.

src/Elastic.Markdown/Myst/Directives/Changelog/ChangelogBlock.cs
```diff
@@ -463,6 +533,44 @@ private void LoadAndCacheBundles()
...
+	private static bool IsValidCdnProduct(string product) =>
+		product.Length > 0 && product.All(c => char.IsAsciiLetterOrDigit(c) || c is '_' or '-');
+
+	/// <summary>
+	/// Infers the CDN product for a valueless <c>:cdn:</c> from the current repository. CDN registries are
+	/// keyed by the canonical product id, not the GitHub repository name (e.g. the <c>elastic-otel-java</c>
+	/// repo publishes the <c>edot-java</c> product), so the repository is mapped via products.yml. Returns
+	/// null when git information is unavailable.
+	/// </summary>
+	private string? InferCdnProductFromRepository()
```
Mpdreamz, line 556: Same style issue — 3-line summary on a private method. One line or none.

src/Elastic.Markdown/Myst/Directives/Changelog/ChangelogBlock.cs
```diff
@@ -463,6 +533,44 @@ private void LoadAndCacheBundles()
...
+	{
+		var repository = Build.Git.RepositoryName;
+		if (string.IsNullOrWhiteSpace(repository) || repository == "unavailable")
```
Mpdreamz: The `"unavailable"` sentinel is still here. `IsNullOrWhiteSpace` now guards the empty case, but this still couples product inference to an internal string from the git info service. If that sentinel ever changes, inference silently falls back to the repository name with no diagnostic. Consider exposing a typed property (e.g. `Build.Git.IsAvailable`) rather than matching the raw string.

src/Elastic.Documentation/ReleaseNotes/ChangelogVersionMatch.cs
```diff
@@ -0,0 +1,35 @@
...
+/// the directive's post-load filter and the CDN fetcher's download-time filter agree.
+/// </summary>
+public static class ChangelogVersionMatch
+{
+	/// <param name="requested">The user-supplied <c>:version:</c> value (already trimmed).</param>
+	/// <param name="target">The bundle's declared target (may be null/empty).</param>
+	/// <param name="file">The bundle file name or path (may be null/empty).</param>
+	public static bool Matches(string requested, string? target, string? file)
+	{
+		if (string.IsNullOrWhiteSpace(requested))
```
Mpdreamz, line 21: The match-all-on-empty semantic is non-obvious to callers. Add a short inline comment: `// empty/whitespace → match all versions`.

tests/Elastic.Markdown.Tests/Directives/ChangelogBasicTests.cs
```diff
@@ -192,6 +293,193 @@ public void RendersContent()
...
+	}
+}
+
+/// <summary>
+/// Verifies CDN-sourced bundles render into the page body (not just the page TOC). The directive is a
+/// selector over release notes prefetched at startup, so the test injects a resolver holding the bundle
+/// instead of hitting the network. Regression guard: the HTML renderer previously gated on the
+/// (CDN-null) local bundles folder path and silently emitted an empty body.
+/// </summary>
+public class ChangelogCdnRenderTests(ITestOutputHelper output) : DirectiveTest<ChangelogBlock>(output,
```
Mpdreamz, line 329: Still missing a test that combines `:cdn:` + `:version:` together in CDN mode. Both options are tested independently but the interaction — filtering prefetched CDN bundles by version — has no coverage.

Mpdreamz (changes requested):
Thanks for the architectural rework — the startup prefetch via `docset.yml` addresses all the main concerns from the previous review. A few small items still need attention.

src/Elastic.Markdown/Myst/Directives/Changelog/ChangelogBlock.cs
cotti: Trimmed to a single-line summary in f0a79cb8.

src/Elastic.Markdown/Myst/Directives/Changelog/ChangelogBlock.cs
cotti: Trimmed to a single-line summary in f0a79cb8. Also swept the other private methods this PR adds (`IsSafeBundleFileName`, `ResolveInlineEntries`, and the `RegistryBuilder` helpers) down to a single line or none.

src/Elastic.Markdown/Myst/Directives/Changelog/ChangelogBlock.cs
cotti: Done in f0a79cb8: added `GitCheckoutInformation.IsAvailable` (the sentinel now lives only in its owning type) and `InferCdnProductFromRepository` checks `Build.Git.IsAvailable` instead of matching the raw `"unavailable"` string.

src/Elastic.Documentation/ReleaseNotes/ChangelogVersionMatch.cs
cotti: Added the inline comment `// empty/whitespace → match all versions` above the guard in f0a79cb8.

tests/Elastic.Markdown.Tests/Directives/ChangelogBasicTests.cs
cotti: Added `ChangelogCdnVersionFilterTests` in f0a79cb8 — injects two prefetched CDN bundles (9.4.0 + 9.3.0) with `:cdn:` + `:version: 9.4.0` and asserts only the matching bundle renders.

Mpdreamz (approved):

PR #3456, opened 2026-06-02
Search: Fix modal rejecting valid responses with content type "docs"

reakaleek:
## Why

The upstream search contract (`Elastic.Internal.Search.Contract`) changed the field name from `type` to `content_type` and updated the value from `"doc"` to `"docs"` on indexed documents. The frontend Zod schemas still declared `type: z.enum(['doc', 'api'])`, so `SearchResponse.parse()` threw a `ZodError` on every result item — rejecting the entire valid 200 response and showing "Error loading search results" in the codex search modal.

## What

- Renamed `type` → `content_type` and updated `z.enum(['doc', 'api'])` to `z.enum(['docs'])` in both `useModalSearchQuery` and `useNavigationSearchQuery`.
- Removed the now-dead `result.type === 'api'` branch from both result list components; codex is docs-only so `typePrefix` is always `'Docs'`.
- Added an `else if (query.error)` branch to the error-logging effect in both hooks, so `ZodError`/parse failures are no longer silent — they now emit `logError` to OTLP and `console.error` to devtools.
- Updated the test fixture to use `content_type: 'docs'`.

### DIFF
--- a/src/Elastic.Documentation.Site/Assets/web-components/ModalSearch/ModalSearchResultsList.tsx
+++ b/src/Elastic.Documentation.Site/Assets/web-components/ModalSearch/ModalSearchResultsList.tsx
@@ -178,9 +178,9 @@ const ModalSearchResultRow = ({
             return result.parents.map((p) => p.title)
         }
 
-        const typePrefix = result.type === 'api' ? 'API' : 'Docs'
+        const typePrefix = 'Docs'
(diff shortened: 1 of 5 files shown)


reakaleek:
@copilot this fix is wrong. The schema changed. It's now content_type with the value "docs".

Just change the key and fix the value in the zod schema

reakaleek:
Merging because this is broken on live elastic codex.

PR #3457, opened 2026-06-02
Codex: add sync plan/apply commands backed by shared deploying lib

reakaleek:
## Why

Codex produces a static site to `.artifacts/codex/docs` but had no way to incrementally sync that output to S3 — only the assembler had that capability. The sync strategies were tightly coupled to `AssembleContext`, making them unusable from codex without duplication.

## What

The incremental S3 sync machinery (plan strategy, apply strategy, plan model, validator, and orchestrator) is extracted from `Elastic.Documentation.Assembler` into a new neutral `Elastic.Documentation.Deploying` assembly. A small `IDocsSyncContext` interface captures the five properties the strategies actually use — `ReadFileSystem`, `WriteFileSystem`, `OutputDirectory`, `Collector`, and `EnvironmentName` — which both `AssembleContext` and `CodexContext` now implement.

Two new commands are wired under `codex sync`:

- `codex sync plan <codex.yml> --s3-bucket-name <bucket> [--environment <env>] --out <file>` — diffs local output against S3 and writes a plan
- `codex sync apply <codex.yml> --s3-bucket-name <bucket> [--environment <env>] --plan <file>` — executes the plan

The assembler's `assembler deploy plan/apply` surface is unchanged; `DeployCommands` now builds its `AssembleContext` locally before calling the generalized service.

## How

`IDocsSyncContext` is the only new abstraction introduced. The strategies never needed more than `ReadFileSystem`, `WriteFileSystem`, `OutputDirectory`, `Collector`, and one environment-name log string — so the interface is deliberately narrow. Both context types already exposed all five; no data was restructured.

## Test plan

- `dotnet build docs-builder.slnx` — clean build
- `docs-builder assembler deploy plan --help` — flags unchanged from before
- `docs-builder codex sync plan --help` / `docs-builder codex sync apply --help` — new commands present with correct flags


docs/cli-schema.json
```diff
@@ -4754,7 +4754,239 @@
...
+                    },
+                    {
+                      "kind": "fileExtensions",
+                      "values": [
+                        "json",
+                        "plan"
+                      ]
+                    }
+                  ]
+                },
```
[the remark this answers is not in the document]
reakaleek: Valid. Made `plan` a non-nullable required parameter (matching how `assembler deploy apply` handles `planFile`), which fixes the schema automatically. Regenerated `docs/cli-schema.json`. Fixed in `01d1348d2`.

src/Elastic.Codex/CodexContext.cs
```diff
@@ -31,6 +32,9 @@ public class CodexContext
 		? "codex"
 		: $"codex-{Configuration.Environment}";
 
+	/// <inheritdoc cref="IDocsSyncContext.EnvironmentName"/>
+	public string EnvironmentName => Configuration.Environment ?? "codex";
```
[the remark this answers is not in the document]
reakaleek: Valid. Replaced `??` with `string.IsNullOrEmpty`, now consistent with `IndexNamespace`. Also took the opportunity to add an `environmentOverride` constructor parameter so the resolved environment from the CLI flag can actually be applied (see comment below). Fixed in `01d1348d2`.

src/tooling/docs-builder/Commands/Codex/CodexSyncCommand.cs
```diff
@@ -0,0 +1,122 @@
...
+	/// <summary>Compute a diff of what would change when deploying to S3 and write it to a plan file.</summary>
+	/// <remarks>
+	/// Two-step deployment: <c>plan</c> computes the diff and writes a plan file; <c>apply</c> executes it.
+	/// Review the plan before applying to avoid accidental mass deletions.
+	/// </remarks>
+	/// <param name="config">Path to the <c>codex.yml</c> configuration file.</param>
+	/// <param name="s3BucketName">S3 bucket to deploy to.</param>
+	/// <param name="environment">Named deployment target. Defaults to the value in <c>codex.yml</c> or the <c>ENVIRONMENT</c> env var.</param>
+	/// <param name="out">Path to write the plan file. Defaults to <c>stdout</c>.</param>
+	/// <param name="deleteThreshold">Abort if the plan would delete more than this percentage of objects (0–100).</param>
```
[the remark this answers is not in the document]
reakaleek: Not a valid new issue. This behavior is identical to `assembler deploy plan`, which also passes `state.@out?.FullName ?? ""` and skips serialization when `--out` is omitted. The doc comment 'Defaults to stdout' is a pre-existing inaccuracy in the assembler design that predates this PR — out of scope here.

src/tooling/docs-builder/Commands/Codex/CodexSyncCommand.cs
```diff
@@ -0,0 +1,122 @@
...
+		var codexConfig = CodexConfiguration.Load(configFile);
+		var resolvedEnvironment = environment
+			?? codexConfig.Environment
+			?? Environment.GetEnvironmentVariable("ENVIRONMENT")
+			?? "internal";
+
+		var context = new CodexContext(codexConfig, configFile, collector, fs, fs, null, null);
+		var service = new IncrementalDeployService(logFactory, githubActionsService);
+		serviceInvoker.AddCommand(service, (context, s3BucketName, @out, deleteThreshold, resolvedEnvironment),
+			static async (s, collector, state, ctx) => await s.Plan(collector, state.context, state.s3BucketName, state.@out?.FullName ?? "", state.deleteThreshold, ctx)
```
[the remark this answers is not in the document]
reakaleek: Valid. Added an `environmentOverride` optional parameter to `CodexContext` which `EnvironmentName` uses when set. Both `Plan` and `Apply` now pass `resolvedEnvironment` into the context constructor, and the dead tuple member is removed. Fixed in `01d1348d2`.

reakaleek:
**Outside-diff comment (IncrementalDeployService.cs:79-83):** Valid. `ServiceInvoker` emits a global error when a command returns `false` without having emitted any errors itself, turning a no-op apply (already-synced bucket) into exit code 1. Changed `return false` to `return true` for the empty-plan path. Fixed in `01d1348d2`.

src/services/Elastic.Documentation.Deploying/IncrementalDeployService.cs
```diff
@@ -88,7 +79,7 @@ public async Task<bool> Apply(IDiagnosticsCollector collector, string environmen
 		if (plan.TotalSyncRequests == 0)
 		{
 			_logger.LogInformation("Plan has no files to sync, skipping incremental synchronization");
-			return false;
+			return true;
 		}
 		var validator = new DocsSyncPlanValidator(logFactory);
 		var validationResult = validator.Validate(plan);
```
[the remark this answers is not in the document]
reakaleek: Is it really needed to do validation if the sync is a noop?

src/tooling/docs-builder/Commands/Codex/CodexSyncCommand.cs
```diff
@@ -0,0 +1,87 @@
...
+		return await serviceInvoker.InvokeAsync(ct);
+	}
+
+	private (CodexContext context, IncrementalDeployService service) LoadContext(FileInfo config)
+	{
+		var fs = FileSystemFactory.RealRead;
+		var configFile = fs.FileInfo.New(config.FullName);
+		var codexConfig = CodexConfiguration.Load(configFile);
+		return (new CodexContext(codexConfig, configFile, collector, fs, fs, null, null),
+			new IncrementalDeployService(logFactory, githubActionsService));
```
Mpdreamz, line 85: Bug: `fs` (`FileSystemFactory.RealRead`) is passed as **both** `readFileSystem` and `writeFileSystem`. `RealRead` doesn't have `AllowedSpecialFolders = Temp`, but `AwsS3SyncApplyStrategy.Upload()` creates a temp staging directory via `context.WriteFileSystem.Path.GetTempPath()`. This will fail at runtime with a scope violation when apply tries to stage files.

The round-trip test doesn't catch it because `MockFileSystem` is wrapped with `ScopeCurrentWorkingDirectoryForWrite` which does include temp.

The second argument should be `FileSystemFactory.RealWrite` (same as `DeployCommands` does for the assembler path):

```suggestion
		return (new CodexContext(codexConfig, configFile, collector, fs, FileSystemFactory.RealWrite, null, null),
```

src/services/Elastic.Documentation.Deploying/IncrementalDeployService.cs
```diff
@@ -61,21 +56,14 @@ public async Task<bool> Plan(IDiagnosticsCollector collector, string environment
...
 			MinSizeBeforePartUpload = S3EtagCalculator.PartSize
 		});
-		if (!readFileSystem.File.Exists(planFile))
-		{
-			collector.EmitError(planFile, "Plan file does not exist.");
-			return false;
-		}
-		var planJson = await readFileSystem.File.ReadAllTextAsync(planFile, ctx);
+		var planJson = await context.ReadFileSystem.File.ReadAllTextAsync(planFile, ctx);
 		var plan = SyncPlan.Deserialize(planJson);
```
Mpdreamz, line 66: The old code had a `File.Exists` guard here (with `collector.EmitError` + `return false`) before reading the plan file. The `[Existing]` CLI attribute protects the current entry points, but the service method itself lost its defensive check. If called programmatically in the future, a missing plan file would throw an unhandled `FileNotFoundException` instead of returning a clean diagnostic error.

Worth restoring the guard above this line for defense-in-depth:

```cs
if (!context.ReadFileSystem.File.Exists(planFile))
{
    collector.EmitError(planFile, "Plan file does not exist.");
    return false;
}
```

src/services/Elastic.Documentation.Deploying/IncrementalDeployService.cs
```diff
@@ -88,7 +76,7 @@ public async Task<bool> Apply(IDiagnosticsCollector collector, string environmen
 		if (plan.TotalSyncRequests == 0)
 		{
 			_logger.LogInformation("Plan has no files to sync, skipping incremental synchronization");
-			return false;
+			return true;
 		}
```
Mpdreamz, line 73: Correct fix — the old `return false` here meant an empty plan (nothing to sync) caused the apply CI step to fail with a non-zero exit code, which is wrong since a no-op sync is a success.

Worth noting: the workflow still calls apply on empty plans because `plan-valid` gates on validation, not on whether there are changes. That's fine since apply handles it gracefully, just something to be aware of.

Mpdreamz (commented):
Functional review — traced the assembler deploy path end-to-end against the production workflow in `.shared.assemble-build-and-deploy.yml`. That path is safe: same data, just different wiring. Found two issues in the new codex path and one note on the empty-plan behavior change.

Mpdreamz (changes requested):
see comments

src/tooling/docs-builder/Commands/Codex/CodexSyncCommand.cs
reakaleek: Valid. `RealRead` (no `AllowedSpecialFolders.Temp`) was passed as both read and write filesystem; `AwsS3SyncApplyStrategy` stages uploads under `WriteFileSystem.Path.GetTempPath()`, so any non-empty `apply` would hit a scope violation at runtime. Changed to `FileSystemFactory.RealWrite` (matches how `DeployCommands` wires the assembler path). Fixed in https://github.com/elastic/docs-builder/pull/3457/commits/f190e25a3e6fbe17f7d68a4d6191c7e9676f67b9.

src/services/Elastic.Documentation.Deploying/IncrementalDeployService.cs
reakaleek: Valid. Restored the `File.Exists` guard before `ReadAllTextAsync`. The `[Existing]` attribute protects the CLI entry point, but `Apply()` is `public` — a programmatic caller with a missing path would get an unhandled `FileNotFoundException` instead of a clean diagnostic error. Fixed in https://github.com/elastic/docs-builder/pull/3457/commits/64dd72e32aa7e2e6e24556b63f9e7e752985d992.

src/services/Elastic.Documentation.Deploying/IncrementalDeployService.cs
reakaleek: Agreed — no change needed. The no-op `return true` is the correct behaviour (an already-in-sync bucket is a success, not an error), and `apply` handling empty plans gracefully means the CI workflow's unconditional `apply` step continues to work without special-casing.

Mpdreamz:
I guess the only thing I am unsure about is:

https://github.com/elastic/docs-internal-workflows/blob/main/.github/workflows/.shared.assemble-build-and-deploy.yml#L158 

`steps.deploy-plan.outputs.plan-valid == 'true'` will now be true for files with 0 changes.

reakaleek:
> I guess the only thing I am unsure about is:
> 
> https://github.com/elastic/docs-internal-workflows/blob/main/.github/workflows/.shared.assemble-build-and-deploy.yml#L158
> 
> `steps.deploy-plan.outputs.plan-valid == 'true'` will now be true for files with 0 changes.

0 files should be a valid plan though.. it never happens because of the llm.zip

Regardless, an apply to 0 files should be a noop.

Mpdreamz (approved):

PR #3458, opened 2026-06-02
Removing deprecated search label

jilldoty-elastic:
Tiny change to remove reference to deprecated :Search/Search label.


docs/data/release-notes/configure-ref.md
```diff
@@ -236,7 +236,7 @@ Also controls how the `changelog add` command maps GitHub labels to various chan
 
 | Setting           | Description                                |
 | ----------------- | ------------------------------------------ |
-| `pivot.areas`     | Lists the valid area values. Optionally maps area names to GitHub labels (for example, `"Search": ":Search/Search"`). |
+| `pivot.areas`     | Lists the valid area values. Optionally maps area names to GitHub labels (for example, `"Search Relevance": ":Search Relevance/Search"`). |
```
shainaraskas, line 255: based on your [other change](https://github.com/elastic/elasticsearch/pull/150539), I think this is wrong

```suggestion
| `pivot.areas`     | Lists the valid area values. Optionally maps area names to GitHub labels (for example, `"Search": ":Search Relevance/Search"`). |
```

jilldoty-elastic: We want to remove the catch-all "Search" label - you have to be more specific now.  "Search" should not map to "Search Relevance" - I just wanted to leave an example in this documentation.

shainaraskas: ok. I want @lcawl or @cotti to weigh in here because I don't know how these are used.

fwiw, your change in the other PR _does_ map search directly to relevance. if that's not what you want, you probably need to make some further changes.

jilldoty-elastic: Fair, the other maps the general Search area to "Search Foundations" - it's the closest to the original purpose.  I was concerned that removing it all together would break something.

shainaraskas: ah, reading too fast. will leave this for the others regardless :)

reakaleek:
cc @lcawl @cotti

Mpdreamz (approved):
LGTM, this is only docs example with not impact. Trust @jilldoty-elastic knows best which labels are relevant 😸

PR #3468, opened 2026-06-04
docs: clarify blank-line rule in admonition multi-line examples

### LINKED ISSUE #3131: Bug: Admonition documentation does not clearly support multi-line/formatting use, causing confusion
### Describe the bug

There is a bug in the documentation regarding admonitions: the current documentation claims admonitions support multiple lines and formatting, but the rendered output does not match this claim. For example, the sample usage in the documentation is misleading, possibly due to missing or extra newlines. This causes confusion for users relying on the documentation to properly format admonitions.

### Expected behavior

Admonitions in the documentation should support multiple lines and formatting as stated. The sample example should clearly demonstrate the correct format, including any necessary newlines, so that the rendered output is as expected.

### Steps to reproduce

1. Go to the documentation on admonitions.
2. Copy the multiline admonition example as shown.
3. Render or build using docs-builder.
4. Observe that the output may not preserve the expected formatting, especially if there isn't the required extra newline.

### Tooling

true,false,false

[View original Slack conversation](https://elastic.slack.com/archives/C07APH4RCDT/p1776347701503579?thread_ts=1776347701.503579&cid=C07APH4RCDT)

cotti:
## Why

Closes #3131. The admonitions page claims callouts "can span multiple lines and support inline formatting," but the **Plain** example reproduced the exact misleading pattern that confused authors: its content reads "It can *span* multiple lines…" yet sits on a single line with no blank line, so it renders as one paragraph. The page also never stated the standard Markdown rule that adjacent lines collapse into one paragraph.

#3129 already corrected the Note example; this finishes the remaining gap.

## What

- Added a blank line inside the Plain example (both the Output and Markdown tabs) so it actually renders as two paragraphs, mirroring the Note fix.
- Added one sentence near the top of "Basic admonitions" explaining that paragraphs inside an admonition must be separated by a blank line.


docs/syntax/admonitions.md
```diff
@@ -4,7 +4,8 @@ Admonitions allow you to highlight important information with varying levels of
 
 ## Basic admonitions
 
-Admonitions can span multiple lines and support inline formatting.
+Admonitions can span multiple lines and support inline formatting. To split the content into separate paragraphs, separate them with a blank line — adjacent lines collapse into a single paragraph.
```
shainaraskas: this follows the method of all paragraphs so we don't need it.

```suggestion
Admonitions can span multiple lines and support inline formatting. 
```

shainaraskas (approved):

PR #3469, opened 2026-06-05
Fix crash when toc references a missing file

### LINKED ISSUE #3245: NullReferenceException in navigation when TOC references a missing file
### Describe the bug

Running `docs-builder serve` on a docset whose `docset.yml` `toc` references a Markdown file that does not exist causes a global unhandled **`System.NullReferenceException`** while building navigation lookups, instead of a clear validation error.

Stack trace (abbreviated):

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Elastic.Documentation.Navigation.NavigationItemExtensions.BuildNavigationLookupsRecursive(INavigationItem, ConditionalWeakTable`2, Dictionary`2)
   at Elastic.Documentation.Navigation.NavigationItemExtensions.BuildNavigationLookups(INavigationItem, ConditionalWeakTable`2)
   at Elastic.Markdown.IO.DocumentationSet..ctor(BuildContext, ILoggerFactory, ICrossLinkResolver)
   at Documentation.Builder.Http.ReloadableGeneratorState..ctor(ILoggerFactory, IDirectoryInfo, IDirectoryInfo, BuildContext, Boolean)
   at Documentation.Builder.Http.DocumentationWebHost..ctor(ILoggerFactory, String, Int32, ScopedFileSystem, ScopedFileSystem, IConfigurationContext, Boolean)
   at Documentation.Builder.Commands.ServeCommand.<Serve>d__4.MoveNext()
```

### Expected behavior

`docs-builder serve` (or a normal build) should report a **diagnostic** that the TOC entry points to a missing file (path + `docset.yml` location), and exit without an unhandled exception.

### Steps to reproduce

Tested with docs-builder **1.7.0** (install script from [README](https://github.com/elastic/docs-builder)).

1. Create an empty directory and a `docs` folder inside it.
2. Add `docs/docset.yml`:

   ```yaml
   icon: magnifyWithPlus
   registry: internal
   toc:
     - file: index.md
   ```

3. **Do not** create `docs/index.md` (the TOC must reference a file that is absent).
4. From the **repository root** (parent of `docs/`), run:

   ```bash
   docs-builder serve
   ```

5. Observe the process crash with `NullReferenceException` as above.

Minimal tree:

```text
repro-root/
  docs/
    docset.yml   # toc -> index.md only
```

### Tooling

- [x] docs-builder
- [ ] migration tooling
- [ ] I'm not sure

cotti:
## Why

Running `docs-builder build`/`serve` on a docset whose `toc` references a Markdown file that doesn't exist on disk crashed with an unhandled `NullReferenceException` while building navigation lookups, instead of reporting a clear validation error. The validation error was actually already computed — the process just died on an unguarded null dereference before it could be shown.

Closes #3245.

## What

When a TOC entry resolves to a file the factory can't create, that nav item is dropped. If a node ends up with zero items, its `Index` is set to a `null!` sentinel. Because `INodeNavigationItem<out TIndex, …>` is covariant, the root node matches the `INodeNavigationItem<IDocumentationFile, …>` branch in both `BuildNavigationLookups` and `DocumentationSet.VisitNavigation`, where `Index.Model` was dereferenced unconditionally → the reported NRE.

This change:
- Guards the null `Index` sentinel at both dereference sites, so the build no longer crashes and the already-emitted validation error surfaces, exiting cleanly.
- Clarifies the diagnostic to say whether the referenced file is missing on disk or exists but isn't a valid Markdown document, anchored to the `docset.yml` location.
- Adds a `Navigation.Tests` case that drives `BuildNavigationLookups` with a missing-file TOC and asserts it does not throw and emits a clear error (the existing validation tests never exercised this path).

Verified end-to-end: a docset with a missing `file:` entry now prints `Table of contents references '…' but the file does not exist on disk.` and exits with an error rather than crashing.

### DIFF
--- a/src/Elastic.Documentation.Navigation/Isolated/Node/DocumentationSetNavigation.cs
+++ b/src/Elastic.Documentation.Navigation/Isolated/Node/DocumentationSetNavigation.cs
@@ -196,14 +196,18 @@ public class DocumentationSetNavigation<TModel>
 	private TModel? CreateDocumentationFile(
 		IFileInfo fileInfo,
 		IFileSystem fileSystem,
-		IDocumentationSetContext context,
-		string fullPath
+		IDocumentationSetContext context
 	)
 	{
 		var relativePath = Path.GetRelativePath(context.DocumentationSourceDirectory.FullName, fileInfo.FullName);
 		var documentationFile = _factory.TryCreateDocumentationFile(fileInfo, fileSystem);
 		if (documentationFile == null)
-			context.EmitError(context.ConfigurationPath, $"File navigation '{relativePath}' could not be created. {fullPath}");
+		{
+			var reason = fileInfo.Exists
+				? "the file exists but is not a valid Markdown document"
+				: "the file does not exist on disk";
+			context.EmitError(context.ConfigurationPath, $"Table of contents references '{relativePath}' but {reason}.");
+		}
 
 		return documentationFile;
 	}
@@ -248,7 +252,7 @@ public class DocumentationSetNavigation<TModel>
 			DetectionRuleRef ruleRef => ruleRef.FileInfo,
 			_ => ResolveFileInfo(context, fullPath)
 		};
-		var documentationFile = CreateDocumentationFile(fileInfo, context.ReadFileSystem, context, fullPath);
+		var documentationFile = CreateDocumentationFile(fileInfo, context.ReadFileSystem, context);
 		if (documentationFile == null)
 			return null;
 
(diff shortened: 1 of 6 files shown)


Mpdreamz (approved):
The fix is correct and well-scoped. Both dereference sites are properly guarded, the improved diagnostic is clearer, and the parameter removal is clean.

One note: the new test covers the `BuildNavigationLookups` path but not the `VisitNavigation` fix in `DocumentationSet.cs`. Both sites have the same structural fix so the risk is low, but a test that constructs a `DocumentationSet` and drives `VisitNavigation` would give complete coverage of the reported crash. Not a blocker — happy to merge as-is.

PR #3470, opened 2026-06-05
changelog: source bundle entries from the CDN with a per-product registry

cotti:
> [!NOTE]
> Stacked on `changelog_directive_s3`. Base this against that branch and merge it first; the diff shown here is only this change.

## Summary

Source the individual changelog entries that make up a bundle from the public CDN by default (scoped to the bundle's product), instead of the local `docs/changelog/` folder. This makes bundles reflect what was actually published to S3 — i.e. with private references scrubbed — and decouples bundle creation from a checkout that happens to hold every entry.

The bundle filter is **content-based**: whether an entry belongs in a bundle depends on the `products` / `prs` / `issues` fields *inside* the YAML, not on the file name. CloudFront has no `ListObjects`, so we publish a small per-product index on upload and the client enumerates → fetches → filters.

## What changed

- **Per-product entry registry**: `upload` now writes `{product}/changelog/registry.json` alongside the existing bundle registry (`RegistryBuilder` / `RegistryKey`); the scrubber key allow-list recognizes it.
- **CDN entry sourcing**: new `CdnChangelogEntryFetcher` downloads the entry index and each entry; CDN base resolution is centralized in `ChangelogCdn` (shared with the changelog directive's `cdn:` mode).
- **Opt-out + safe fallback**: `bundle.use_local_changelogs: true` forces the local folder, and we fall back to local automatically when no concrete product can scope the per-product CDN fetch.
- **`--plan` cdn_url**: `changelog bundle --plan` now emits `cdn_url` (`{base}/{product}/bundle/{file}`) so CI can poll for the scrubbed bundle. `docs/cli-schema.json` regenerated.
- **No silent gaps**: a registry-listed entry that hasn't propagated to the CDN yet is retried with short backoff + cache-busting (defeats a CloudFront-cached 404); a persistent miss **fails the bundle** rather than shipping an incomplete release.

## Verification

- `dotnet format --verify-no-changes`: clean
- `dotnet publish -c Release` (docs-builder): 0 trim/AOT warnings
- `Elastic.Documentation.Configuration.Tests` (409/409) and `Elastic.Changelog.Tests` (732/732): pass
- `cli-schema.json` matches `-- __schema` output

## Test plan

- [ ] `upload` publishes `{product}/changelog/registry.json` and individual entries
- [ ] `bundle` (default) fetches scrubbed entries from the CDN and applies the content filter
- [ ] `bundle` with `use_local_changelogs: true` uses the local folder
- [ ] `bundle --plan` surfaces a correct `cdn_url`
- [ ] A not-yet-propagated entry recovers via retry; a persistently missing entry fails the run

### DIFF
--- a/docs/cli-schema.json
+++ b/docs/cli-schema.json
@@ -2784,7 +2784,7 @@
               "name": "plan",
               "type": "boolean",
               "required": false,
-              "summary": "Emit GitHub Actions step outputs (needs_network, needs_github_token, output_path) describing network requirements and the resolved output path, then exit without generating the bundle. Intended for CI actions.",
+              "summary": "Emit GitHub Actions step outputs (needs_network, needs_github_token, output_path, and cdn_url when a product is resolvable) describing network requirements, the resolved output path, and the public CDN URL of the scrubbed bundle, then exit without generating the bundle. Intended for CI actions.",
               "defaultValue": "false"
             },
             {
--- a/docs/contribute/configure-changelogs-ref.md
+++ b/docs/contribute/configure-changelogs-ref.md
@@ -53,6 +53,7 @@ These settings are relevant to one or all of the `changelog bundle`, `changelog 
 | `bundle.release_dates`    | When `true`, bundles include a `release-date` field (default: true). |
 | `bundle.repo`             | Default GitHub repository name (for example, `elasticsearch`). Used by the `{changelog}` directive to generate correct PR and issue links. Only needed when the product ID doesn't match the GitHub repository name. |
 | `bundle.resolve`          | When `true`, changelog contents are copied into bundle (default: `true`). |
+| `bundle.use_local_changelogs` | When `true`, always source entries from the local folder and never from the CDN (default: `false`). Refer to [Entry sourcing](#bundle-entry-sourcing). |
 
 :::
 
@@ -62,6 +63,23 @@ When `bundle.link_allow_repos` is omitted, no link filtering occurs.
 - For private repos, set it to `[]` or add related public repos to the list.
 - For public repos, add your `owner/repo` to the list at a minimum.
 :::
+
+### Entry sourcing [bundle-entry-sourcing]
+
+`changelog bundle` reads the individual changelog entries it aggregates either from the local folder or from the public CDN. CDN sourcing is **opt-in per product** (declared-gate): an entry is pulled from the CDN only when its product is declared under `release_notes` in the docset's `docset.yml`.
+
+```yaml
+# docset.yml
+release_notes:
+  - product: elasticsearch
+```
+
+Sourcing is decided per run:
+
+- **Local folder (default).** Used when no product in scope is declared under `release_notes`, when `bundle.use_local_changelogs: true`, when `--directory` is passed, or when the filter resolves no concrete product (for example, `--all` or PR/issue-only filters). The folder must contain the changelog files.
+- **CDN.** Used only when every product in scope is declared under `release_notes` and none of the local-sourcing conditions above apply. The product must also exist in [products.yml](https://github.com/elastic/docs-builder/blob/main/config/products.yml) with the `release-notes` feature enabled.
+
+The product ID under `release_notes` matches the product format described in [](/cli/changelog/bundle.md#product-format). This is the same declaration the `{changelog}` directive's `:cdn:` mode consumes, so a repository that opts into CDN-sourced bundling and CDN-rendered release notes declares each product once.
 
 ### Bundle descriptions [bundle-descriptions]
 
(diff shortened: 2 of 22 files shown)


Mpdreamz (changes requested):
A few code-quality notes, then a broader design question worth discussing before this merges.

## Code issues

**`HttpClient` is never disposed in `CdnChangelogEntryFetcher`.**
The class allocates an `HttpClient` in a field initializer but doesn't implement `IDisposable`, so the connection pool leaks in any reuse scenario (tests, future service mode). Should implement `IDisposable` and dispose the client, or accept an `IHttpClientFactory`.

**`Fetch()` blocks the calling thread for all HTTP I/O.**
`FetchRegistry` and `FetchText` use the synchronous `HttpClient.Send()` overload. Everything else in the service layer is async; this blocks a thread-pool thread for the full CDN round-trip per product.

**Multi-paragraph XML doc comments on private methods.**
`FetchCdnEntries`, `ResolveCdnBundleUrl`, and `ResolvePrimaryProduct` all have verbose `<summary>` blocks. Project style is one short line max (or none) on private methods.

**`PlanBundleAsync` duplicates the CDN-producibility check.**
The plan path and the bundle path compute "is a product resolvable?" with different expressions at different pipeline stages (pre- vs post-`ApplyConfigDefaults`). Currently correct, but there is no comment explaining the difference, so they are likely to diverge as the code evolves.

---

## Design question — should this be declarative in `docset.yml`?

The current approach fetches changelog entries imperatively inside `changelog bundle`, driven by a `use_local_changelogs` escape hatch in `changelog.yml`. That works, but it means:

- The fetch happens deep in the bundle command where concurrency is ad-hoc (synchronous loop over entries per product).
- The build cannot validate or prefetch CDN requirements up-front — it discovers the dependency at execution time.
- The changelog directive has nowhere to point the user on a fetch error; it can only say "fetch failed" rather than "declare this in your config".

Compare how crosslinks work: repos are declared in `docset.yml` under `cross_links`, docs-builder knows about the external dependency at startup, can fail fast if the index is unreachable, and can fetch all link indexes concurrently before any page is rendered.

A similar pattern for changelog sourcing might look like:

```yaml
# docset.yml
release_notes:
  - repository: elastic/elasticsearch
  - repository: elastic/kibana
```

With that declaration:
- docs-builder knows at startup which CDN products to pull, can fetch all registries and entries concurrently, and can fail fast before any directive tries to render.
- The `changelog` directive just references an already-loaded in-memory set — no per-directive HTTP.
- If a directive references a product not declared in `docset.yml`, it emits a clear error pointing the user at the config key rather than a raw CDN failure at render time.
- `changelog bundle` reads the same declaration to know which products to source from the CDN, keeping the config in one place.

The `use_local_changelogs` flag would still make sense as an escape hatch, but the primary sourcing decision would be declared rather than implied by the bundle command's product filter.

Is this the direction you want to go, or is there a reason to keep it embedded in `changelog.yml` / the bundle command?

cotti:
### Rebased onto the updated #3436 and reconciled with the opt-in model; addressed the `HttpClient` feedback

**Rebase**
- Rebased onto #3436's declarative model; dropped the now-duplicate `ChangelogCdn` and the superseded directive edits (they live in #3436).

**`HttpClient` (review feedback)**
- `CdnChangelogEntryFetcher` is now `IDisposable` and fully async: a shared `HttpClient` in production, an owned client disposed with the instance only when a test injects a handler, and async retry backoff (`Task.Delay`) instead of blocking sleeps. Propagated through `FetchCdnEntriesAsync`.

**Declared-gate sourcing**
- `changelog bundle` sources entries from the CDN **only when every in-scope product is declared under `release_notes` in `docset.yml`** — the same opt-in declaration the directive consumes. Otherwise (or with `bundle.use_local_changelogs`, `--directory`, or no concrete product in scope) it falls back to local folder sourcing.
- The same gate drives the `--plan` `needs_network` output, so the planning step and the Docker bundle run stay in agreement.
- Unified product resolution across the run + plan paths; trimmed verbose private-method XML docs.

**Docs / tests** — added a `bundle.use_local_changelogs` entry + an "Entry sourcing" section to `configure-changelogs-ref.md`, a producer-side declared-gate section to the registry dev doc, and declared-gate test coverage (`OptionMode_UndeclaredProduct_FallsBackToLocal`, `Plan_ProfileMode_UndeclaredProduct_ReturnsNoNetwork`).

**Client / docs-actions hand-off**
- Opt-in is now required per product: `elastic-otel-java` declares `edot-java`; `cloud` declares its rendered products (e.g. `cloud-hosted`, `cloud-serverless`, `cloud-enterprise`).
- `docs-actions` needs **no workflow change** — the registry refresh is already a side-effect of `changelog upload`, and `--plan` reflects the gate. Just publish each product's registry **before** declaring it (strict fail-fast).

Mpdreamz (approved):

PR #3472, opened 2026-06-05
Report all changelog errors at once and surface fetch failures in add

cotti:
## Why

Two pain points were reported while bundling a serverless release:

1. `changelog bundle` aborts on the **first** invalid entry, so a release with many broken changelog files (e.g. missing `title`) forces a fix-and-rerun loop, one error at a time.
2. `changelog add` silently degrades when a PR/issue can't be fetched from GitHub: the entry **bypasses `rules.create` label filtering** and is written with `title`/`type` commented out. With a missing/unauthorized `GITHUB_TOKEN` this produced ~225 unfiltered, title-less changelogs instead of 68 — and the failure was buried among hundreds of per-PR warnings, only showing up later as a `bundle` error.

## What

- `BundleBuilder.BuildResolvedEntries` (the `bundle` path) and `BundleValidationService.ValidateBundleEntriesAsync` (the `render` path) now validate every entry and report all problems in a single pass before failing, instead of returning on the first error.
- `changelog add` now emits a single aggregate summary at the end of bulk creation when PRs/issues couldn't be fetched (e.g. `3 of 225 pull request(s) could not be fetched from GitHub...`), with remediation guidance. Default behavior is unchanged: warning, exit 0, best-effort files still written.
- New `--strict-fetch` flag escalates fetch failures to an error (non-zero exit) so CI fails loudly on a token/rate-limit problem rather than producing unfiltered changelogs. Files are still written for inspection.
- Regenerated `docs/cli-schema.json` and documented the flag + fetch-failure behavior in `docs/cli/changelog/cmd-add.md`.

Tests added for both behaviors; full `Elastic.Changelog.Tests` suite passes and the AOT publish is clean.

### DIFF
--- a/docs/cli-schema.json
+++ b/docs/cli-schema.json
@@ -2556,6 +2556,14 @@
             },
             {
               "role": "flag",
+              "name": "strict-fetch",
+              "type": "boolean",
+              "required": false,
+              "summary": "Optional: Treat a failure to fetch any PR or issue from GitHub (with --prs, --issues, or --report) as an error that exits non-zero, instead of a warning. Use in CI so a missing or unauthorized GITHUB_TOKEN fails the run rather than silently producing unfiltered changelogs with missing titles. Files are still written so they can be inspected.",
+              "defaultValue": "false"
+            },
+            {
+              "role": "flag",
               "name": "subtype",
               "type": "string",
               "required": false,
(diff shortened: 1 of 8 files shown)


Mpdreamz (approved):
Two minor notes:

1. **Multi-paragraph docstring on `ReportBulkFetchFailures`** — project style requires one short line max (or no doc comment at all for private helpers). Drop or collapse it.

2. **Missing `--issues + StrictFetch` test** — `CreateChangelogsForMultipleIssuesAsync` got the same changes as the PR path but there's no corresponding test. Low risk since the code is symmetric, but worth adding for parity.

Otherwise the fix is correct and well-scoped — the `--report` path routes through `CreateChangelogsForMultiplePrsAsync` so the main motivating scenario is covered.

PR #3493, opened 2026-06-12
feat(mcp): add deeper OpenTelemetry integration for MCP tool calls

Mpdreamz:
## Why

MCP tool calls were invisible in Elastic Observability — every call collapsed into a single `POST /docs/_mcp/` transaction because all tools share one HTTP route. There was also no metrics data for tool calls at all.

## What

Each tool call now appears as its own named transaction (e.g. `tools/call search_docs`) in APM, and tool-level metrics are exported.

**Tracing:** `McpSpanRenameProcessor` renames the ASP.NET Core server span in `OnEnd` — after the route-based name is set — to `tools/call {tool}`. The standard OTel MCP semconv attributes `mcp.method.name` and `mcp.tool.name` are written to the server span by `McpToolTelemetry.StartActivity` so the processor can read them.

**Metrics:** A `Meter` (`Elastic.Documentation.Api.McpTools`) was added with two instruments — `mcp.tool.calls` (counter) and `mcp.tool.duration` (histogram) — both dimensioned by tool name, method, server profile, and outcome. They're recorded in `LogCompletion`, which all three tool classes already call in their `finally` blocks, so no per-tool changes were needed.

**OTel registration cleanup:** `Extensions.cs` and `EuidEnrichmentExtensions.cs` were replaced by a unified `AddDocumentationOpenTelemetry` extension that all three services (docs-api, docs-mcp, docs-builder CLI) now use consistently.

**Aspire 13.4.3 bump:** Fixes a new `required` member (`ExecuteCommandContext.Arguments`) in the integration test bootstrap.


theletterf:
So necessary, thanks!

theletterf (approved):

PR #3501, opened 2026-06-15
chore(mcp): document stateless transport posture and remove SseKeepAliveMiddleware

Mpdreamz:
## Why

The `SseKeepAliveMiddleware` was added as a workaround for a Cursor SSE-fallback bug, and the
stateless-mode comment blamed that same bug. Both are now wrong: in SDK 1.4+, stateless mode and
legacy SSE are mutually exclusive (enabling both throws at startup), and the Cursor bug is an
unfixed client-side regression that server config cannot address.

The keepalive middleware allocated an `SseKeepAliveStream` + `SemaphoreSlim` on **every** MCP
request even though no `text/event-stream` response is ever produced in stateless mode — the timer
never started, but the allocations always happened.

## What

- Rewrites the stateless-mode comment to explain the actual architectural justification: pure
  request/response tools with no server push, running behind a load balancer with no session
  affinity. Adds a one-line pointer to the `mcp-remote` stdio bridge for SSE-only clients.
- Removes `SseKeepAliveMiddleware` and its `SseKeepAliveStream` inner class (~285 lines).
- Updates the `ConfigureHttpJsonOptions` comment to reflect its real purpose (not legacy SSE).

## How

In ModelContextProtocol.AspNetCore 1.4.0, `WithHttpTransport(o => o.Stateless = true)` + `MapMcp("")`
mounts exactly one route: `POST /` (Streamable HTTP request/response). Legacy SSE (`/sse` +
`/message`) is opt-in via `EnableLegacySse` (default `false`) and throws `InvalidOperationException`
at startup if combined with `Stateless = true`. The SDK emits no SSE heartbeat, but in stateless
mode there is no SSE stream to keep alive.


reakaleek (approved):

aspire/Properties/launchSettings.json
```diff
@@ -7,7 +7,7 @@
       "launchBrowser": true,
       "applicationUrl": "https://localhost:17166;http://localhost:15066",
       "environmentVariables": {
-        "ASPNETCORE_ENVIRONMENT": "Development",
+        "ASPNETCORE_ENVIRONMENT": "dev",
```
Mpdreamz:
```suggestion
        "ASPNETCORE_ENVIRONMENT": "Development",
```

PR #3510, opened 2026-06-15
OtlpProxy: buffer request body to survive stale-connection retry

reakaleek:
## Why

`AdotOtlpService` was wrapping `context.Request.Body` directly in `StreamContent`. `SocketsHttpHandler` has a built-in transparent connection retry (`SendWithVersionDetectionAndRetryAsync`) that fires when it picks a stale pooled keep-alive connection. For a non-seekable stream, `StreamContent.AllowRetry` returns `false`, so `SocketsHttpHandler` skips the transparent retry and throws `HttpRequestException` wrapping an `IOException`. The existing `catch` handles this, but it was returning 502 — which the browser OTLP exporter interprets as "please retry", defeating fire-and-forget semantics. The noisy error logs were also misleading.

## What

Three small changes:

- **`PooledConnectionLifetime = 30s`** on the `SocketsHttpHandler` — proactively recycles connections before the sidecar can close them, making stale connections extremely rare.
- **Map `HttpRequestException { InnerException: IOException }` → 204** — a silent drop. The browser exporter sees 204 (success) and moves on rather than retrying.
- **`otlp.proxy.stale_connection.dropped` counter** — emitted on every silent drop so the rate is observable.

### DIFF
--- a/src/api/Elastic.Documentation.Api/Telemetry/AdotOtlpService.cs
+++ b/src/api/Elastic.Documentation.Api/Telemetry/AdotOtlpService.cs
@@ -34,8 +34,17 @@ public class AdotOtlpService(
 			var targetUrl = $"{options.Endpoint.TrimEnd('/')}/v1/{signalType.ToStringFast(true)}";
 			logger.LogDebug("Forwarding OTLP {SignalType} to ADOT collector at {TargetUrl}", signalType, targetUrl);
 
-			using var request = new HttpRequestMessage(HttpMethod.Post, targetUrl);
-			request.Content = new StreamContent(requestBody);
+			// Buffer the body so SocketsHttpHandler's transparent connection retry can replay it.
+			// The raw request body is forward-only; a stale pooled connection to the localhost
+			// collector triggers a silent resend that otherwise throws "stream was already consumed".
+			using var buffer = new MemoryStream();
+			await requestBody.CopyToAsync(buffer, ctx);
+			buffer.Position = 0;
+
+			using var request = new HttpRequestMessage(HttpMethod.Post, targetUrl)
+			{
+				Content = new StreamContent(buffer) // seekable — SocketsHttpHandler can replay on retry
+			};
 			_ = request.Content.Headers.TryAddWithoutValidation("Content-Type", contentType);
 
 			using var response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseContentRead, ctx);
--- a/src/api/Elastic.Documentation.Api/Telemetry/AdotOtlpService.cs
+++ b/src/api/Elastic.Documentation.Api/Telemetry/AdotOtlpService.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information
 
 using System.Diagnostics;
+using System.Diagnostics.Metrics;
 using System.Net.Sockets;
 using Microsoft.Extensions.Logging;
 
@@ -18,6 +19,10 @@ public class AdotOtlpService(
 {
 	public const string HttpClientName = "OtlpProxy";
 	private static readonly ActivitySource ActivitySource = new(TelemetryConstants.OtlpProxySourceName);
+	private static readonly Meter Meter = new(TelemetryConstants.OtlpProxySourceName);
+	internal static readonly Counter<int> StaleConnectionDrops =
+		Meter.CreateCounter<int>("otlp.proxy.stale_connection.dropped",
+			description: "OTLP batches silently dropped due to a stale pooled connection to the ADOT collector");
 	private readonly HttpClient _httpClient = httpClientFactory.CreateClient(HttpClientName);
 
 	/// <inheritdoc />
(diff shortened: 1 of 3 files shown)


tests/Elastic.Documentation.Api.Tests/OtlpProxyTests.cs
```diff
@@ -264,6 +264,91 @@ public async Task OtlpProxy_CollectorUnavailable_FailsFastWithoutRetries()
...
+
+		// Act — should not throw; the buffered body is replayable
+		using var response = await client.PostAsync("/docs/_api/v1/o/t", content, TestContext.Current.CancellationToken);
+
+		// First attempt threw HttpRequestException → mapped to 502 by the service.
+		// The retry here is at the application level (service catches the exception);
+		// the point is no InvalidOperationException escapes.
+		response.StatusCode.Should().BeOneOf(HttpStatusCode.NoContent, HttpStatusCode.BadGateway);
+		callCount.Should().BeGreaterThan(0);
+	}
```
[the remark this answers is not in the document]
reakaleek: https://github.com/elastic/docs-builder/pull/3510/commits/680d7768e31ff184096d2e4e6fbaec57607cb1ed

src/api/Elastic.Documentation.Api/Telemetry/AdotOtlpService.cs
```diff
@@ -34,8 +34,17 @@ public async Task<OtlpForwardResult> ForwardOtlp(
 			var targetUrl = $"{options.Endpoint.TrimEnd('/')}/v1/{signalType.ToStringFast(true)}";
 			logger.LogDebug("Forwarding OTLP {SignalType} to ADOT collector at {TargetUrl}", signalType, targetUrl);
 
-			using var request = new HttpRequestMessage(HttpMethod.Post, targetUrl);
-			request.Content = new StreamContent(requestBody);
+			// Buffer the body so SocketsHttpHandler's transparent connection retry can replay it.
+			// The raw request body is forward-only; a stale pooled connection to the localhost
+			// collector triggers a silent resend that otherwise throws "stream was already consumed".
+			using var buffer = new MemoryStream();
+			await requestBody.CopyToAsync(buffer, ctx);
```
Mpdreamz: This `MemoryStream` allocation (+ full body copy) happens on every OTLP forward. Since the request body is non-seekable, `StreamContent.AllowRetry` already returns `false`, which tells `SocketsHttpHandler` to *not* attempt the transparent retry — it will throw `HttpRequestException` instead. 

So the buffer isn't needed to prevent a crash; the existing `catch` already handles `HttpRequestException`. The tradeoff: we copy the entire body into heap memory on every request, vs. occasionally dropping a trace on a rare stale connection. For fire-and-forget telemetry, the drop is preferable — see the PR-level comment for the full alternative.

Mpdreamz (commented):
## Alternative: zero-copy, best-effort

This buffer fixes the `InvalidOperationException` correctly, but it allocates a `MemoryStream` (and copies the entire body into it) on every OTLP forward. For a high-frequency telemetry path that's real memory pressure with no real gain — OTLP is fire-and-forget, so dropping one batch on a rare stale connection is totally acceptable.

The alternative is to stay zero-copy and let stale connections fail gracefully:

**1. Keep `StreamContent(requestBody)` — no buffer needed.**
`StreamContent` exposes an internal `AllowRetry` property that returns `stream.CanSeek`. For ASP.NET's non-seekable request body, `AllowRetry = false`, so `SocketsHttpHandler` will *not* attempt the transparent retry — it throws `HttpRequestException` instead. Your existing `catch` already handles that.

**2. Cycle connections before they go stale** (in `ServicesExtension.AddOtlpProxyService`):
```csharp
.ConfigurePrimaryHttpMessageHandler(() => new SocketsHttpHandler
{
    PooledConnectionLifetime = TimeSpan.FromSeconds(30), // localhost sidecar — reconnect is ~free
    ConnectTimeout = TimeSpan.FromSeconds(1),
})
```
With a 30s lifetime, the window for a connection going stale is tiny. This makes the drop rate negligible in practice.

**3. Don't surface stale-connection failures as 502 to the caller.** The browser OTLP exporter will interpret 502 as a signal to retry, which defeats the fire-and-forget intent. Map the stale-connection case to 204 instead (silent drop).

**4. Emit a metric so we can watch the drop rate:**
```csharp
// in TelemetryConstants or alongside ActivitySource:
private static readonly Meter Meter = new(TelemetryConstants.OtlpProxySourceName);
internal static readonly Counter<int> StaleConnectionDrops =
    Meter.CreateCounter<int>("otlp.proxy.stale_connection.dropped",
        description: "Traces silently dropped due to stale pooled connection to ADOT collector");
```
Then in `MapExceptionToStatusCode`, add a case before the generic `HttpRequestException` arm:
```csharp
// Stale pooled connection — SocketsHttpHandler detected it before sending body.
// OTLP is best-effort; drop silently and let the next request reconnect.
HttpRequestException { InnerException: IOException }
    => (204, string.Empty), // caller should not retry
```
And increment `StaleConnectionDrops` in the catch block when the status is 204.

This is effectively what YARP does in its no-buffer mode — accept the occasional drop on a stale connection in exchange for zero copy and zero allocation on the hot path. Since the collector is a localhost sidecar, reconnect is microseconds and drops should be extremely rare (<< 0.1% of requests) once `PooledConnectionLifetime` keeps connections fresh.

src/api/Elastic.Documentation.Api/Telemetry/AdotOtlpService.cs
reakaleek: makes a lot of sense. thank you. Addressing it now

reakaleek: https://github.com/elastic/docs-builder/pull/3510/commits/41f6ee3adf1f6a3f555a8415965496b08b46de19

Mpdreamz (approved):

PR #3521, opened 2026-06-17
OtlpProxy: remove in-app proxy; route OTLP at the ALB

reakaleek:
## Why

The in-app \`AdotOtlpService\` proxy forwarded every browser OTLP batch over loopback to the
\`otel-collector\` sidecar. It was the source of \`Connection reset by peer\` errors (stale pooled
connections) and silent data loss. The root fix is to remove the app from the hot path entirely:
the collector sidecar already binds \`0.0.0.0:4318\` (OTLP HTTP) on the shared awsvpc task ENI
and is directly reachable from the ALB.

## What

Deletes the proxy layer from docs-builder (\`AdotOtlpService\`, \`IOtlpService\`, \`OtlpForwardResult\`,
\`OtlpProxyOptions\`, \`OtlpProxyRequest\`, and \`OtlpProxyTests\`) and removes all call sites
(\`MappingsExtensions\` \`/o/*\` routes, \`ServicesExtension.AddOtlpProxyService\`,
\`OtelRegistration.HttpClientTracingFilter\`). Net: −769 lines.

The browser's existing paths (\`/docs/_api/v1/o/{t,l,m}\`) and the app's own server-side EDOT
export (\`localhost:4317\`) are unchanged.

## How

Two companion changes complete the migration (separate PRs in their respective repos):

- **docs-infra**: ALB listener rule (priority < 200) routing \`/docs/_api/v1/o/*\` → new target
  group on collector port 4318, health-checked on port 13133; SG ingress for 4318 + 13133.
- **docs-internal-workflows**: otel-collector \`COLLECTOR_CONFIG\` sets \`traces_url_path\`,
  \`logs_url_path\`, \`metrics_url_path\` to the adblock-safe paths so the collector accepts them
  without any ALB rewrite.

Deploy sequence: collector config → infra (ALB rule) → this app change.


Mpdreamz (approved):
Such a better and improved solution ❤️

Mpdreamz:
The only thing is local development won't have the endpoints how are we handling that @reakaleek ? 


I think we should leverage Aspire more to start services mirroring production, wdyt?

reakaleek:
> The only thing is local development won't have the endpoints how are we handling that @reakaleek ?
> 
> I think we should leverage Aspire more to start services mirroring production, wdyt?

yup yup, we probably need to run a local otel collector to make this work locally (or what ever aspire can provide)

PR #3524, opened 2026-06-17
fix: move website search input container outside htmx swap region

reakaleek:
## Why

The `<elastic-website-search>` component renders its chat-input box into `#elastic-website-search-input-container`. That container lived inside `#content-container`, which is OOB-swapped on every htmx navigation. After each navigation the old container is replaced by a fresh empty one, so the component's rendered chat UI vanishes.

## What

Added `hx-preserve="true"` to `#elastic-website-search-input-container`. htmx matches it by id during the OOB swap and retains the live node — preserving the component's rendered content — rather than replacing it with the empty server-rendered one.


reakaleek:
@coderabbitai review

reakaleek:
@coderabbitai approve

Mpdreamz (approved):

PR #3525, opened 2026-06-18
fix(stepper): skip headings inside steps when calculating next step heading level

theletterf:
## Why

When a heading is manually added inside a stepper step, two problems occur: the *next* step renders at an inferior heading level, and the internal heading itself can collide with the step's own rendered level, producing invalid document structure.

## What

All steps in a stepper share the same preceding-heading context, so the level calculation now happens once in \`StepperBlock.FinalizeAndValidate()\` and is pushed to every child step. Because \`Descendants()\` is pre-order, looking backward from the stepper's position in the flat block list naturally lands on document-level predecessors only — no extra filtering needed.

After setting each step's level, \`AdjustInternalHeadings\` walks the step's \`HeadingBlock\` descendants (skipping headings inside nested steps, which are handled by their own stepper). Any heading at a level ≤ the step's rendered level is automatically bumped to \`stepLevel + 1\`, and a \`Hint\` diagnostic is emitted at the heading's exact line telling the author what level to write instead.

Headings already deeper than the step are left untouched with no diagnostic.

### DIFF
--- a/tests/Elastic.Markdown.Tests/FileInclusion/HeadingOrderTests.cs
+++ b/tests/Elastic.Markdown.Tests/FileInclusion/HeadingOrderTests.cs
@@ -849,6 +849,110 @@ Step after h2 heading.
 }
 
 /// <summary>
+/// Tests that a heading at the same level as a step is auto-adjusted to one level deeper
+/// and that a hint diagnostic is emitted pointing to the heading.
+/// </summary>
+public class StepperWithInternalHeadingAtSameLevelTests(ITestOutputHelper output) : DirectiveTest<StepperBlock>(output,
+"""
+## Section
+
+:::::{stepper}
+
+::::{step} First Step
+### Internal Heading
+
+Some content under the internal heading.
+::::
+
+::::{step} Second Step
+This step should still be at the same level as First Step.
+::::
+
+:::::
+"""
+)
+{
+	[Fact]
+	public void ParsesBlock() => Block.Should().NotBeNull();
+
+	[Fact]
+	public void InternalHeadingIsAdjustedToOneLevelDeeper()
+	{
+		var toc = File.PageTableOfContent.Values.ToList();
+
+		// Should have: Section (2), First Step (3), Internal Heading (4, adjusted), Second Step (3)
+		// The step renders at h3. The ### inside it was also h3 (invalid — same level).
+		// It is auto-adjusted to h4.
+		toc.Should().HaveCount(4);
+
+		toc[0].Heading.Should().Be("Section");
+		toc[0].Level.Should().Be(2);
+
+		toc[1].Heading.Should().Be("First Step");
+		toc[1].Level.Should().Be(3, "step is one level deeper than the preceding h2");
+		toc[1].IsStepperStep.Should().BeTrue();
+
+		toc[2].Heading.Should().Be("Internal Heading");
+		toc[2].Level.Should().Be(4, "### inside an h3 step is adjusted to h4");
+		toc[2].IsStepperStep.Should().BeFalse();
+
+		toc[3].Heading.Should().Be("Second Step");
+		toc[3].Level.Should().Be(3);
+		toc[3].IsStepperStep.Should().BeTrue();
+	}
+
+	[Fact]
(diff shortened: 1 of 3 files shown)


reakaleek (approved):

florent-leborgne (commented):
Nice fix. One question though if I understand correctly what's being done...

If a heading inside of a step is at the same level than the step itself, should we respect it, or push it down one level?

If I understand this PR, it respects it, but this seems like an anti-pattern. I don't think we want:

```
#1 How to do something
##2 (stepper - step 1)
##2 manual section inside of step 1
##2 manual section inside of step 1
##2 (stepper - step 2)
##2 (stepper - step 2)
```

But rather:

```
#1 How to do something
##2 (stepper - step 1)
###3 manual section inside of step 1
###3 manual section inside of step 1
##2 (stepper - step 2)
##2 (stepper - step 2)
```

theletterf:
Ahhh, good point. Well, the correct behavior here would be to raise an exception, because the source defines a heading level, and changing it on build breaks expectations. I'm not even sure we want headings within steps at all. 

@reakaleek WDYT?

florent-leborgne:
Agree it's not a pattern we should use IMO as I also wrote in the thread, but if we do allow it, then we need to make it show correctly on the page. It might also help a few contributors to not run into errors because it might not be super easy to understand at first that :::step is going to turn into a heading (and that by consequence you need to offset your heading levels by 1 and essentially "skip" a level in terms of traditional markup), so if we can handle this automatically that makes it a simpler experience for everyone, and that enforces consistency in our docs

theletterf:
We could go for a graceful dynamic fix plus a Hint to suggest what the correct level should be. I'll test this.

theletterf (commented):
Fixed — capped `adjusted` at 6 with `Math.Min(stepLevel + 1, 6)`.

theletterf:
@florent-leborgne That's exactly what this PR now does now. Any heading inside a step that is at the same level or higher than the step's rendered level is automatically adjusted to `stepLevel + 1` (capped at h6). At the same time a `Hint` diagnostic is emitted at the exact line of the offending heading — so the author sees something like:

```
HINT: Heading level h4 inside a step renders at the same or higher level as the step itself (h4).
      It has been adjusted to h5 — write it as '#####' to avoid this hint.
      (docs/syntax/stepper.md:246)
```

The page renders correctly regardless, and the hint tells them exactly what to write to clean it up. The stepper docs also include a live example that intentionally triggers this hint on every build, so the behaviour is always exercised.

tests/Elastic.Markdown.Tests/FileInclusion/HeadingOrderTests.cs
```diff
@@ -848,6 +848,110 @@ public void StepperStepsInMainDocumentCalculateCorrectHeadingLevels()
...
+
+:::::
+"""
+)
+{
+	[Fact]
+	public void ParsesBlock() => Block.Should().NotBeNull();
+
+	[Fact]
+	public void DeepHeadingIsUntouched()
```
reakaleek, line 936: What happens to multiple levels within a step?

e.g.

```markdown
:::{step}
  ## Foo
  ### Bar
  #### Baz
:::
```

is every level increased based on the "base" level?

theletterf: Good question! The current behaviour only adjusts headings where `level <= stepLevel` — they all get bumped to `stepLevel + 1`. Headings already deeper are left untouched.

So for a step that renders as h3, your example would produce:
- `## Foo` → adjusted to h4
- `### Bar` → adjusted to h4
- `#### Baz` → left at h4

All three land at h4 — the relative hierarchy among the internal headings collapses. That's a known limitation of the simple "clamp to stepLevel+1" approach. For a single misplaced heading it works well; for a structured sub-hierarchy the author should write the correct levels from the start (they'll get a hint for each offending heading pointing to the fix).

theletterf: It's also a pattern we don't want to endorse :)

PR #3539, opened 2026-06-19
fix(applies-to): avoid weird "Removed Planned" multi-lifecycle badge

### LINKED ISSUE #3537: Multi-lifecycle applies_to renders weird "Removed Planned" badge
### Describe the bug

When an `applies_to` directive uses multiple lifecycles where the highest-version lifecycle is unreleased — for example, a feature that is currently `deprecated` and is `removed` in a future version — the badge renders as a weird concatenation of the lifecycle name and the synthesized fallback text:

```
Stack | Removed Planned
```

The badge is showing the lifecycle name (`Removed`) **and** a synthesized `Planned` label side by side, which reads awkwardly and is misleading: the feature isn't removed yet, it's currently deprecated.

### Expected behavior

Per the [`applies_to` syntax docs](https://github.com/elastic/docs-builder/blob/main/docs/syntax/applies.md), when the highest-version `Removed` lifecycle is unreleased and there are 2+ lifecycles, the badge should \"use the previous lifecycle\". For `stack: deprecated, removed 10.0` (with stack current 9.4.2), that means the badge should render as:

```
Stack | Deprecated
```

…reflecting the page's current state, with the future removal information surfaced in the popover.

### Steps to reproduce

1. Author a page with the following frontmatter or directive:

    ````md
    ```{applies_to}
    stack: deprecated, removed 10.0
    ```
    ````

2. With stack `current: 9.4.2` (per [`config/versions.yml`](https://github.com/elastic/docs-builder/blob/main/config/versions.yml)).
3. Build the page and inspect the rendered badge.

Observed: `Stack | Removed Planned`
Expected: `Stack | Deprecated`

### Tooling

- [x] docs-builder

### Root cause

In `src/Elastic.Markdown/Myst/Components/ApplicabilityRenderer.cs`:

1. Applicabilities are sorted by version descending, so `removed 10.0` is processed first.
2. `BuildBadgeLifecycleText` short-circuits to `\"\"` when `lifecycleCount >= 2` (the \"use previous lifecycle\" rule), so the highest-version candidate produces no displayable badge text or version.
3. The `deprecated` (no version) candidate also produces no badge text or version.
4. The catch-all fallback then forces `BadgeLifecycleText = \"Planned\"` on the first candidate's badge data — but leaves `ShowLifecycleName = true`, so the badge renders the lifecycle name (`Removed`) **and** the forced text (`Planned`) together.

The fallback never actually consults the lower-priority lifecycle (`deprecated`), so the documented \"use previous lifecycle\" rule is never honoured for this shape of input.

shainaraskas:
## Why

When an `applies_to` directive declares multiple lifecycles where the highest-version one is unreleased — most commonly `stack: deprecated, removed <future>` — the badge currently renders as a confusing concatenation like `Stack | Removed Planned`. The feature/option isn't actually removed yet (it's currently deprecated), and the doubled label looks like a bug to readers.

Closes #3537.

## What

In `ApplicabilityRenderer`, the catch-all fallback used to force `BadgeLifecycleText = "Planned"` on the highest-version candidate while leaving `ShowLifecycleName = true`, so the lifecycle name and the synthesised text rendered side by side.

The fallback now honours the documented "use previous lifecycle" rule: when the highest candidate is future/unreleased and there is a lower-version applicability that is current or version-less, we render that one's badge instead. So `stack: deprecated, removed 10.0` (with stack current `9.4.2`) now renders as `Stack | Deprecated`, with the planned removal still surfaced in the popover.

Only when *every* applicability is unreleased do we fall back to a synthesised label, and that path now uses the lifecycle-appropriate phrase (`Removal planned` / `Deprecation planned` / `Planned`) with `ShowLifecycleName = false` so the doubled rendering can't recur.

### DIFF
--- a/src/Elastic.Markdown/Myst/Components/ApplicabilityRenderer.cs
+++ b/src/Elastic.Markdown/Myst/Components/ApplicabilityRenderer.cs
@@ -78,14 +78,37 @@ public static class ApplicabilityRenderer
 			}
 		}
 
-		// If we've exhausted all options (none had displayable data), use the first one.
-		// Only show "Planned" when the first applicability is actually future/unreleased (has a version spec that is not yet released).
-		// When the first applicability has no version (null/AllVersionsSpec), it means GA for all versions - keep badge text empty.
+		// If we've exhausted all options (none had displayable data), pick a fallback.
+		// When the highest-version applicability is future/unreleased, prefer the "previous lifecycle"
+		// (a lower-version applicability that is current or has no version) over a synthesized "Planned" badge.
+		// This avoids weird renderings like "Removed Planned" for `deprecated, removed X` where X is unreleased.
 		if (badgeData is null && firstBadgeData is not null && firstApplicability is not null && versioningSystem.IsVersioned())
 		{
 			var versionSpec = firstApplicability.Version;
 			var isFutureVersion = versionSpec is not null && versionSpec != AllVersionsSpec.Instance && versionSpec.Min > versioningSystem.Current;
-			badgeData = isFutureVersion ? firstBadgeData with { BadgeLifecycleText = "Planned" } : firstBadgeData;
+
+			if (isFutureVersion)
+			{
+				var previousLifecycle = sortedApplicabilities.FirstOrDefault(a =>
+					a != firstApplicability &&
+					(a.Version is null || a.Version == AllVersionsSpec.Instance ||
+					 a.Version.Min <= versioningSystem.Current));
+
+				if (previousLifecycle is not null)
+					badgeData = GetBadgeData(previousLifecycle, versioningSystem, allApplications);
+				else
+				{
+					var fallbackText = firstApplicability.Lifecycle switch
+					{
+						ProductLifecycle.Deprecated => "Deprecation planned",
+						ProductLifecycle.Removed => "Removal planned",
+						_ => "Planned"
+					};
+					badgeData = firstBadgeData with { BadgeLifecycleText = fallbackText, ShowLifecycleName = false };
+				}
+			}
+			else
+				badgeData = firstBadgeData;
 		}
 
 		badgeData ??= GetBadgeData(sortedApplicabilities.First(), versioningSystem, allApplications);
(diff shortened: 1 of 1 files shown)


src/Elastic.Markdown/Myst/Components/ApplicabilityRenderer.cs
```diff
@@ -78,14 +78,37 @@ public static ApplicabilityRenderData RenderApplicability(
 			}
 		}
 
-		// If we've exhausted all options (none had displayable data), use the first one.
-		// Only show "Planned" when the first applicability is actually future/unreleased (has a version spec that is not yet released).
-		// When the first applicability has no version (null/AllVersionsSpec), it means GA for all versions - keep badge text empty.
+		// If we've exhausted all options (none had displayable data), pick a fallback.
+		// When the highest-version applicability is future/unreleased, prefer the "previous lifecycle"
+		// (a lower-version applicability that is current or has no version) over a synthesized "Planned" badge.
+		// This avoids weird renderings like "Removed Planned" for `deprecated, removed X` where X is unreleased.
```
shainaraskas:
```suggestion
```

src/Elastic.Markdown/Myst/Components/ApplicabilityRenderer.cs
```diff
@@ -78,14 +78,36 @@ public static ApplicabilityRenderData RenderApplicability(
 			}
 		}
 
-		// If we've exhausted all options (none had displayable data), use the first one.
-		// Only show "Planned" when the first applicability is actually future/unreleased (has a version spec that is not yet released).
-		// When the first applicability has no version (null/AllVersionsSpec), it means GA for all versions - keep badge text empty.
+		// If we've exhausted all options (none had displayable data), pick a fallback.
+		// When the highest-version applicability is future/unreleased, prefer the "previous lifecycle"
+		// (a lower-version applicability that is current or has no version) over a synthesized "Planned" badge.
```
shainaraskas, line 83: @copilot I think this comment has lost some nuance. can you provide a suggestion that doesn't destroy what the old comment documented

src/Elastic.Markdown/Myst/Components/ApplicabilityRenderer.cs
```diff
@@ -78,14 +78,38 @@ public static ApplicabilityRenderData RenderApplicability(
...
+		// This keeps combinations like `deprecated, removed 10.0` from rendering as "Removed Planned"
+		// when the removal version is still unreleased.
```
shainaraskas:
```suggestion
```

reakaleek (approved):

PR #3540, opened 2026-06-19
codex: limit ungrouped spaces to two columns on landing page

theletterf:
## Why

The Codex landing grid uses 3 columns for all items. When the number of ungrouped spaces doesn't divide evenly into 3, a lone card at the end of a row leaves two empty slots — a visually jarring gap.

## What

The outer grid is reduced from `lg:grid-cols-3` to `md:grid-cols-2`. Ungrouped cards now flow in a 2-column layout, so an odd orphan leaves at most one empty slot. Groups with 2+ docsets are promoted to `col-span-full` (they fill both outer columns), while their internal card grid still renders up to 3 columns inside the group box — so group density is unchanged.


theletterf:
@reakaleek

reakaleek (approved):

PR #3615, opened 2026-07-06
Absorb website-search-data: essc and search contract move in-repo

Mpdreamz:
## Why

`website-search-data` was originally abstracted out of docs-builder to share a search API layer with `elastic/website-search`. That API layer has since been deleted, and website-search is TypeScript — it consumes the Elasticsearch indices, not the .NET contract package. With no external consumer left, keeping `essc` and the contract in a separate repository only forced release coordination across two repos for what is effectively one system.

## What

Moves the two remaining website-search-data projects in-repo (plain copy of its `main` @ `34989e8`, no history graft):

- **Search contract** — `Elastic.Internal.Search.Contract` becomes `src/services/search/Elastic.Documentation.Search.Contract`, consumed as a ProjectReference by `Elastic.Documentation` and the search service. NuGet publishing is dropped; the package freezes at 0.12.0 on nuget.org. `IndexTimeSynonyms` moves into the contract so the Markdown exporter uses it directly instead of a hardcoded copy behind a "keep in sync" comment.
- **essc** — lands at `src/tooling/essc`, keeping its `Elastic.SiteSearch` identity. Distribution is container-only: `ghcr.io/elastic/website-search-essc`, built and smoke-tested by the existing `publishcontainers` / run-container pipeline (replaces `ghcr.io/elastic/website-ai-search-essc`; the 4-RID GitHub-release binaries are discontinued).
- **Tests** — `Elastic.SiteSearch.Tests` (TUnit / Microsoft.Testing.Platform) joins `tests/`. Since VSTest's `dotnet test` hard-errors on MTP-only projects under the .NET 10 SDK, the xunit injection in `tests/Directory.Build.props` is guarded behind `UseTUnit` and the project runs via `dotnet run` in the unit-test target.
- **User secrets** — one shared `docs-builder` store replaces the GUID store and essc's separate store; Elasticsearch settings normalize to `Parameters:ElasticsearchUrl` / `Parameters:ElasticsearchApiKey` (env var names unchanged). CI now sets the `Parameters:`-prefixed keys the fixtures actually read — previously the fetched credentials never reached the integration tests, so those tests now talk to the real readonly cluster for the first time.
- Package versions reconcile to the highest of both repos (Elastic.Mapping/Ingest 0.49.0, Spectre.Console 0.55.0, TUnit 1.19.74).

Note for deploys: absorbing contract `main` includes the post-0.12.0 relevance/mapping changes, so the mapping hash changes and indices recreate on the next docs publish. Existing developers need two one-time local migrations (secrets store renames) — documented in `docs/development/essc.md` and `aspire/README.md`.

Follow-ups: archive `elastic/website-search-data`; repoint consumers of the old essc image; restore the parked probe-mode search code now that `SearchQueryComponents` is in-repo.


Mpdreamz:
discussed today with @reakaleek merging this one in.

PR #3617, opened 2026-07-06
Restructure search document contract: rename, nest, dedupe

Mpdreamz:
## Why

The repo's search document schema (`SearchDocumentBase` and its subtypes) had drifted from the legacy `ws-content` field vocabulary it needs to align with, and carried mapping issues (dead fields, scattered flat groups of related fields, a redundant plain-text/stripped-text pair) that made the contract harder to reason about and query consistently across docs, site, labs, guide, and website-search documents.

## What

- Renames fields to match the target vocabulary: `url`→`path`, `language`→`locale`, `abstract`→`summary`, `navigation_section`→`section`, `content_tags`→`tags`.
- Merges `stripped_body` into `body` as the single plain-text source of truth for search, AI enrichment input, and content hashing — docs-builder no longer feeds raw LLM-Markdown into `body`.
- Flattens `DocumentationDocument.Product` from an `{id, repository}` object to a plain keyword id.
- Groups scattered flat fields into nested records: `OpenGraphData` (`og.*`), `TwitterCardData` (`twitter.*`), `HttpMetadata` (`http.*`), `NavigationMetrics` (`navigation.depth`/`navigation.table_of_contents`).
- Removes two dead fields: `enrichment_prompt_hash` (never written by any producer or the AI enrichment orchestrator) and `enrichment_key` (unconsumed; its ContentTypeUid signal now feeds essc's `section` classification alongside the existing URL heuristic).
- Adds Elasticsearch `alias` fields for every renamed/moved path so in-flight queries against the old flat names keep resolving during rollout. `ai_*` fields stay flat, since the AI enrichment orchestrator can only discover top-level, non-nested `[AiField]` properties.
- Bumps `Elastic.Ingest.Elasticsearch`/`Elastic.Mapping` to 0.49.2, which fixes a source-generator bug (internal-only `NestedBuilder` members breaking cross-assembly consumers) uncovered by the new nested types.

All producers (docs-builder's `ElasticsearchMarkdownExporter`, essc's `ContentStackMapper`/`LabsHtmlExtractor`), the query layer, and the MCP gateway are updated to match. Full solution builds clean; the only test failures anywhere are a pre-existing, unrelated environment issue present before this change.


reakaleek:
Could this break any live service we have? MCP, Codex search?

Mpdreamz:
I added field aliases, but it might break on _source parsing. We have to deploy MCP and API in short succession (minutes downtime) after indexing or I have to add a temporary json decoder (lots of work while we are not officially launched). 

Wdyt @reakaleek

reakaleek:
I guess it's fine. At the same time, we have an active cursor plugin.

Nevertheless, we should only merge and release this when we have time to actively fix anything.

reakaleek (approved):

PR #3621, opened 2026-07-07
Classify new Contentstack content types in essc discovery

Mpdreamz:
## Why

Contentstack is introducing new content types (blog_landing, tutorial, notebook, series, etc.) that aren't ready to appear in search yet, plus a long tail of reusable component/taxonomy types (hero, accordion_faq, tags_*, etc.) that will never represent a standalone page. `PageContentTypes.All` is an opt-in allow-list, so none of these are synced today — but the discovery command (`essc contentstack types`) had no way to distinguish "not yet classified" from "deliberately not a page," making it easy for a genuinely new page-worthy type to go unnoticed.

## What

- Added `PageContentTypes.Blocked` for content types intentionally excluded from sync for now (still being authored, not search-ready).
- Added `PageContentTypes.KnownNonPages` for component/taxonomy/config content types that are never expected to sync.
- `contentstack types` now renders a `Status` column per discovered type (`synced` / `ignored` / `skipped` / `unregistered`) and exits 1 if any discovered type isn't classified in `All`, `Blocked`, or `KnownNonPages`, so newly-added Contentstack types can't silently go unclassified.
- `SyncCommand` is unchanged — it already only syncs types explicitly listed in `PageContentTypes.All`.


cindychangy:
@Mpdreamz: just saw this from the [Slack thread](https://elastic.slack.com/archives/C09K72TNF0A/p1783442082325179?thread_ts=1783005096.583449&cid=C09K72TNF0A). I noticed a `KnownNonPages` section, can we add the following to that section:

```
public static readonly string[] KnownNonPages =
[
    ...
    category_cta
    auto_linking_settings
    auto_linking_term
    glossary_term
  ]
```

Can we also add `glossary` to the `public static readonly string[] Blocked =` section. Thank you!

reakaleek (approved):

Mpdreamz:
thanks @cindychangy ! Added `glossary` to `Blocked`, and `category_cta`, `auto_linking_settings`, `auto_linking_term`, `glossary_term` to `KnownNonPages`. Pushed in b397f9b9c.

PR #3630, opened 2026-07-08
Tables: add fullscreen view and matrix cross-hair highlight

reakaleek:
## Why

Large tables get squeezed by the docs site's three-column layout, forcing horizontal scroll with no way to see the whole table at once. There was also no good way to make a lookup-style table (e.g. a matrix of scores) easy to scan by row and column.

## What

Tables wider than the content column now get a small expand button docked onto the table's top-right corner. Clicking it opens the table in a fullscreen `<dialog>` with a sticky header row, so wide reference tables are readable without cramped scrolling.

The existing `{table}` directive gains a `:matrix:` option for lookup-style tables (a header row across the top, a row-heading first column): hovering a cell highlights its whole row and column together, and the first column is styled like the header row so it reads as a row heading. This cross-hair highlight is always on inside the fullscreen view, where cross-referencing matters most, regardless of whether `:matrix:` was set.

Docs for both features are added to `docs/syntax/tables.md` with runnable examples.

## Test plan

- `dotnet test tests/Elastic.Markdown.Tests` — new `TableDirectiveMatrixTests` / `TableDirectiveWithoutMatrixTests` cases pass alongside existing table directive tests.
- Manually verified in a local build: expand button only appears on tables that actually overflow; fullscreen dialog opens/closes via button, backdrop click, and Escape; `:matrix:` cross-hair highlights the hovered row and column without over-highlighting the row-heading column or its header corner; behavior confirmed both inline and inside the fullscreen dialog.

### DIFF
--- a/docs/syntax/tables.md
+++ b/docs/syntax/tables.md
@@ -105,7 +105,7 @@ Every table is responsive by default. The table will automatically scroll horizo
 
 ## Matrix highlight
 
-The `{table}` directive's `:matrix:` option highlights the whole row *and* column of the hovered cell. It only makes sense for lookup-style tables: a header row across the top and a row-heading first column — like the sprint velocity table below — where readers cross-reference a row label against a column label to find a value. In the fullscreen view this is enabled for every table.
+The `{table}` directive's `:matrix:` option highlights the whole row *and* column of the hovered cell. It only makes sense for lookup-style tables: a header row across the top and a row-heading first column — like the sprint velocity table below — where readers cross-reference a row label against a column label to find a value. This also applies in the fullscreen view for tables that use `:matrix:`.
 
 :::::{tab-set}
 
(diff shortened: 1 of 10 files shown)


Mpdreamz (approved):

bmorelli25:
Nice enhancement. One thing. If I hadn't read this issue or seen the demo, I don't think I'd know that this is scrollable. There's no scrollbar until I start sideways scrolling (maybe that's just a browser setting) and my brain doesn't correlate the expand box with the ability to scroll.

In other words, as implemented, I wouldn't know there are more columns to the right.

Screenshot 2026-07-08 at 9 42 43 AM

reakaleek:
@bmorelli25 

Added in https://github.com/elastic/docs-builder/pull/3630/commits/c542072aa4e062d259185f01d01efe21e36f7f30

https://github.com/user-attachments/assets/db921586-7c81-4173-914c-dfe958ea78fb

PR #3633, opened 2026-07-09
Footer: disable htmx boosting on footer links

reakaleek:
## Why

Footer links all point off-site to elastic.co, but the `<footer>` elements inherit `hx-boost="true"` from the global `<body>` — so links in the footer were falsely boosted, with htmx intercepting clicks for an AJAX/history navigation instead of a normal browser navigation.

## What

Added `hx-disable="true"` to each `<footer>` element (`_AssemblerFooter.cshtml`, `_IsolatedFooter.cshtml` — both branches, `_CodexFooter.cshtml`), which disables all htmx processing on the footer and its descendants.


reakaleek:
Merging and releasing ASAP

PR #3649, opened 2026-07-13
CI: Increase assembler preview S3 upload concurrency

reakaleek:
## Why

Assembler previews upload thousands of small files to S3, and the AWS CLI's default transfer settings limit concurrency, slowing down preview publishing on the larger runner.

## What

Configures the AWS CLI's S3 transfer planner to use the classic transfer client with a higher concurrent request limit (50) and queue size (20000) before running `aws s3 sync` in the assembler preview workflow.

### DIFF
--- a/.github/workflows/assembler-preview.yml
+++ b/.github/workflows/assembler-preview.yml
@@ -108,7 +108,13 @@ jobs:
         env:
           AWS_RETRY_MODE: standard
           AWS_MAX_ATTEMPTS: 6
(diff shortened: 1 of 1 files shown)


Mpdreamz (approved):
In was wondering why this upload took so long sometimes

reakaleek:
> In was wondering why this upload took so long sometimes

Wow, what an improvement. ~1.5mins instead of ~9mins

PR #3650, opened 2026-07-13
Sync: upload/delete S3 files concurrently, not one directory call

reakaleek:
## Why

S3 sync during deploy is network-bound, not CPU-bound: the old apply strategy copied every file to a temp directory and issued a single `TransferUtilityUploadDirectoryRequest`, whose concurrency was tied to CPU count via `TransferUtilityConfig.ConcurrentServiceRequests`. For a docs-site sync with many small files, that under-utilizes available network concurrency and wastes a full local read+write pass per file just to stage a directory that already matched the S3 key layout. A single failed upload/delete could also abort the whole batch.

## What

Uploads and deletes now run per-file/per-batch in parallel via `Parallel.ForEachAsync`, with concurrency configurable through `DOCS_SYNC_UPLOAD_CONCURRENCY` (default 32, max 128) and `DOCS_SYNC_DELETE_CONCURRENCY` (default 4, max 16). Each file is uploaded directly from its original path with an explicit S3 key — no temp-directory staging. Failures are now isolated and reported per-file/per-batch through the diagnostics collector instead of aborting the whole sync, and `AwsS3SyncApplyStrategy.Apply` returns `false` if any file failed (previously the caller always reported success). The default `AmazonS3Client` also now configures SDK-level retry (`RequestRetryMode.Standard`, `MaxErrorRetry = 5`) to absorb throttling under the higher concurrency.

## Test plan

- [x] `dotnet build` succeeds
- [x] Integration tests pass (`DocsSyncTests`, `IncrementalDeployRoundTripTests`), including new tests covering delete-batch chunking and upload-failure propagation


Mpdreamz (approved):
LGTM, let's make sure we isolate this in a release (EMG a release only containing this change before releasing. 

That way we can isolate the change and see it on edge and staging and rollback roll forward.

PR #3655, opened 2026-07-16
Improve 404 recovery with related page suggestions

theletterf:
The 404 page keeps its baseline recovery guidance and docs-home link in every environment. Edge and preview builds can also suggest related documentation through a feature-gated search experience.

**Affects:** Site UI, Search, Docs API

**Prompt summary:** Improve dead-link recovery by suggesting relevant documentation pages on the 404 page. Keep the new experience behind a feature flag until its production API and CloudFront dependencies are ready.

## Why

A missing page gives readers little help finding the content that replaced it. Related-page suggestions can shorten that recovery path, but production does not yet expose all required infrastructure, so disabling the experiment must not remove the existing recovery UI.

#### Baseline recovery

The generic explanation and docs-home link render regardless of `RELATED_PAGES`. The feature flag controls only the `<related-pages>` component and the sentence that promises suggestions.

#### Related-page suggestions

The web component derives a bounded query from the missing URL and displays up to five useful results. Current `main` loads the component lazily only when the page contains its custom element.

#### Search API

The related-pages endpoint validates path length and uses the shared search service with semantic search forced. The semantic query retains the document filter, and the response omits the missing URL itself and unusable results.

#### Local and preview validation

The local static host serves the generated 404 page with an HTTP 404 response. The `assembler-preview` workflow also publishes it at https://docs-v3-preview.elastic.dev/elastic/docs-builder/docs/3655/opentelemetry-collector as a temporary HTTP 200 review route because the shared CloudFront distribution cannot resolve a PR-specific error target.

## Verify

```bash
dotnet build
dotnet test tests/Elastic.Documentation.Api.Tests/
dotnet test tests/Elastic.Documentation.Api.Infrastructure.Tests/
dotnet test tests/Elastic.SiteSearch.Tests/
cd src/Elastic.Documentation.Site
npm run test -- --runInBand RelatedPages loadWebComponents
npm run compile:check
```

**Out of scope:** Production enablement still requires docs-infra to expose the search API independently of the Search or Ask AI cookie gate and configure CloudFront to preserve the missing URL and HTTP 404 status. The 404 intentionally has no inline search bar or full-search link.

**Risk:** `config/assembler.yml` enables `RELATED_PAGES` only for edge and preview. Production behavior remains on the baseline recovery path.

Addresses https://github.com/elastic/docs-eng-team/issues/663

### DIFF
--- a/src/Elastic.Markdown/Layout/_NotFound.cshtml
+++ b/src/Elastic.Markdown/Layout/_NotFound.cshtml
@@ -5,17 +5,31 @@
 			<p class="font-sans font-semibold text-blue-elastic">404</p>
 			<h1 class="mt-2 text-4xl font-sans font-semibold">Page not found</h1>
 			<p class="mt-4 text-lg text-ink-light">
-				The page might have moved or the link might be outdated. Search Elastic Docs or try one of the suggested pages below.
+				The page might have moved or the link might be outdated.
+				@if (Model.Features.RelatedPagesEnabled)
+				{
+					<text>Search Elastic Docs or try one of the suggested pages below.</text>
+				}
+				else
+				{
+					<text>Search Elastic Docs or use one of the links below.</text>
+				}
 			</p>
 		</div>
 
-		<div class="max-w-xl mx-auto mt-8">
-			<navigation-search placeholder="Search Elastic Docs"></navigation-search>
-		</div>
+		@if (Model.Features.SearchOrAskAiEnabled)
+		{
+			<div class="max-w-xl mx-auto mt-8">
+				<navigation-search placeholder="Search Elastic Docs"></navigation-search>
+			</div>
+		}
 
-		<div class="mt-10">
-			<related-pages></related-pages>
-		</div>
+		@if (Model.Features.RelatedPagesEnabled)
+		{
+			<div class="mt-10">
+				<related-pages></related-pages>
+			</div>
+		}
 
 		<div class="mt-10 flex flex-wrap justify-center gap-6">
 			<a href="@Model.Link("/")" class="link">
--- a/src/Elastic.Markdown/Layout/_NotFound.cshtml
+++ b/src/Elastic.Markdown/Layout/_NotFound.cshtml
@@ -8,21 +8,14 @@
 				The page might have moved or the link might be outdated.
 				@if (Model.Features.RelatedPagesEnabled)
 				{
-					<text>Search Elastic Docs or try one of the suggested pages below.</text>
+					<text>Try one of the suggested pages below.</text>
 				}
 				else
 				{
-					<text>Search Elastic Docs or use one of the links below.</text>
+					<text>Use one of the links below.</text>
 				}
 			</p>
 		</div>
-
-		@if (Model.Features.SearchOrAskAiEnabled)
-		{
-			<div class="max-w-xl mx-auto mt-8">
-				<navigation-search placeholder="Search Elastic Docs"></navigation-search>
-			</div>
-		}
 
 		@if (Model.Features.RelatedPagesEnabled)
 		{
--- a/src/Elastic.Markdown/Layout/_NotFound.cshtml
+++ b/src/Elastic.Markdown/Layout/_NotFound.cshtml
@@ -4,17 +4,13 @@
 		<div class="text-center">
 			<p class="font-sans font-semibold text-blue-elastic">404</p>
 			<h1 class="mt-2 text-4xl font-sans font-semibold">Page not found</h1>
-			<p class="mt-4 text-lg text-ink-light">
-				The page might have moved or the link might be outdated.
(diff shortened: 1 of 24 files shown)


src/Elastic.Markdown/Layout/_NotFound.cshtml
```diff
@@ -1,11 +1,30 @@
...
+		<div class="text-center">
+			<p class="font-sans font-semibold text-blue-elastic">404</p>
+			<h1 class="mt-2 text-4xl font-sans font-semibold">Page not found</h1>
+			<p class="mt-4 text-lg text-ink-light">
+				The page might have moved or the link might be outdated. Search Elastic Docs or try one of the suggested pages below.
+			</p>
+		</div>
+
+		<div class="max-w-xl mx-auto mt-8">
+			<navigation-search placeholder="Search Elastic Docs"></navigation-search>
```
reakaleek: We cannot use this. Needs to be synced with site search.

theletterf: Can we temporarily and then replace it? Is there a branch from which I could extract the proper site search logic?

reakaleek: Can we just remove it for now?

Let's get the related pages thing without search. WDYT?

And then we can discuss with the wider team on how to get a search bar here.

theletterf: Sounds good!

reakaleek:
Since this is hard to preview, can you attach a video or screenshots?

src/Elastic.Markdown/Layout/_NotFound.cshtml
```diff
@@ -1,11 +1,37 @@
...
-</div>
+<main hx-disable="true" class="font-body min-h-[60vh] px-6 py-16 md:py-24">
+	<div class="max-w-3xl mx-auto">
+		<div class="text-center">
+			<p class="font-sans font-semibold text-blue-elastic">404</p>
+			<h1 class="mt-2 text-4xl font-sans font-semibold">Page not found</h1>
+			<p class="mt-4 text-lg text-ink-light">
+				The page might have moved or the link might be outdated.
+				@if (Model.Features.RelatedPagesEnabled)
+				{
```
reakaleek: We shouldn't show anything at all if the feature is not enabled.

src/Elastic.Markdown/Layout/_NotFound.cshtml
```diff
@@ -1,11 +1,37 @@
...
+		}
+
+		<div class="mt-10 flex flex-wrap justify-center gap-6">
+			<a href="@Model.Link("/")" class="link">
+				Go to docs home
+				<svg class="link-arrow" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor">
+					<path stroke-linecap="round" stroke-linejoin="round" d="M17.25 8.25 21 12m0 0-3.75 3.75M21 12H3"/>
+				</svg>
+			</a>
+			<a href="@Model.Link("/_search")" class="link">Open full search</a>
```
reakaleek: Also this link doesn't officially exist.

theletterf:
@github Solve the conflicts.

Mpdreamz (changes requested):
Thanks, this is a good direction overall. The API shape is small, the query path is bounded, and the AOT JSON registrations for the new response types are present.

I think there is one user-facing blocker before this can merge:

- `src/Elastic.Markdown/Layout/_NotFound.cshtml`: the baseline 404 recovery copy and docs-home link are now inside `Model.Features.RelatedPagesEnabled`. Since `RELATED_PAGES` is only enabled for `edge`/`preview` in `config/assembler.yml` and is intentionally not enabled for prod yet, the default/prod 404 page would render as essentially just the 404 label + heading with no explanation and no way back to docs home. The feature flag should hide only the new `<related-pages>` experience, not the existing fallback recovery path. I would keep the generic copy and docs-home link unconditional, then gate only the related-pages widget and any copy that specifically promises suggested pages.

  The new disabled-path test currently locks in the regression (`RenderLayout_RelatedPagesDisabled_OmitsRecoveryOptions` expects `Go to docs home` to be absent). Please flip that test so it asserts the baseline recovery UI remains present while `<related-pages>` is absent.

The branch is also currently conflicting with `main`, and the green checks are from July 16, 2026. After the rebase I would like to see the affected markdown/API/search/frontend tests run again, especially because this touches `config/assembler.yml`, the local static host route setup, and the search API surface.

Non-blocking notes from the pass: the temporary assembler-preview copy step matches the PR description and looks acceptable as a short-lived preview workaround, and adding the document filter to the forced semantic path makes sense so hidden/root placeholder documents do not sneak in via semantic-only matches.

theletterf:
@Mpdreamz Addressed!

PR #3656, opened 2026-07-16
Fix LLM export on in-memory filesystems

theletterf:
## Summary

- create llm.zip through the exporter-provided IFileSystem instead of physical ZipFile APIs
- enumerate and copy Markdown files through IFileSystem so docs-builder serve works with MockFileSystem output
- create the output directory when the exporter is finalized without prior page output
- add a regression test that builds and inspects the archive entirely in memory

## Validation

- ./build.sh lint
- dotnet test tests/Elastic.Markdown.Tests/ --no-restore --disable-build-servers -m:1

All 1,895 Elastic.Markdown tests pass.


theletterf:
+CC @Mpdreamz I was getting errors related to llms.zip when doing a local preview.

reakaleek (approved):

PR #3675, opened 2026-07-20
Fix streamed documentation layout shifts

reakaleek:
## Why

Large documentation pages could briefly render their article in the narrow sidebar column while the HTML was still streaming. The asynchronous global navigation also pushed the entire page down after hydration.

## What

Explicit grid placement now keeps content and navigation in their intended columns before every element has been parsed. The assembler also reserves the responsive global-navigation height before its script renders, with synthetic checks covering both layout contracts.

## Test plan

- [x] Run all .NET unit tests
- [x] Run frontend Jest tests, TypeScript checks, lint, and formatting checks
- [x] Build production frontend assets
- [x] Verify throttled loading at desktop, tablet, and mobile breakpoints


Mpdreamz (approved):
Layout shifts are the worst, thanks for catching this one.

PR #3682, opened 2026-07-21
Fix merged changelog TOC links

reakaleek:
## Why

Merged changelog bundles include `+` in their repository name. Heading IDs normalize that character away, while generated table-of-contents links retained it, leaving subsection links unable to find their target.

## What

Normalize generated changelog subsection slugs with the same slugifier used by heading rendering. Add regression coverage confirming merged-bundle TOC slugs match their rendered heading IDs.


Mpdreamz (approved):

lcawl (approved):
Good catch, thanks!

PR #3697, opened 2026-07-22
codex: fix silent no-op on external config path

Mpdreamz:
## Summary

Running `docs-builder codex clone ~/Projects/codex-environments/.../config.yml` printed `0 Errors / 0 Warnings / 0 Hints`, finished in ~45ms, and cloned nothing — looked like success.

Two independent bugs, two commits:

**Commit 1 — fail loudly on config load failure**

- `ScopedFileSystem.Exists` is unguarded (delegates to the inner FS without scope validation), so the `if (!configFile.Exists)` guard passes for out-of-scope paths. `OpenText` then throws `ScopedFileSystemException`, which escaped into `CatchExceptionMiddleware` — but by then `await using var serviceInvoker` had already disposed the collector, printed `0 Errors`, and set `_stopped = true`, so the real error was silently dropped.
- `ServiceInvoker` now starts the diagnostics reader eagerly in a field initializer (`EnsureReaderStarted`). The collector is hot for the entire `await using` scope, so any diagnostic emitted before `InvokeAsync` (guard clauses, config loading, context construction) drains live. The base `DiagnosticsCollector` fire-and-forget semantics are unchanged.
- New `CodexConfigurationLoader` (in `Elastic.Codex`) provides `TryLoad` with two overloads — one that validates and returns `out string environment` (required), one that skips it (for `UpdateRedirects`). Wraps `ScopedFileSystemException`, YAML parse errors, and IO errors in a visible diagnostic. All five codex command handlers now use the helper and `return 1` on failure.

**Commit 2 — external config actually loads**

- All codex handlers previously used `FileSystemFactory.RealRead` (scoped to `WorkingDirectoryRoot + ApplicationData`) for both read and write. A config outside the working tree can't be `OpenText`'d against that scope.
- Read FS: `ScopeCurrentWorkingDirectory(new FileSystem(), [Paths.FindGitRoot(config.FullName)])` — extends the standard scope with the config's git root (falls back to the file's directory when no `.git`).
- Write FS: `RealWrite` for clone/index/update-redirects (outputs go to ApplicationData); `RealGitRootForPathWrite(null, output)` for clone-and-build/build (working dir + ApplicationData + optional `--output`). Keeps `WorkingDirectoryRoot` in scope for build output — a blanket swap to `RealGitRootForPath(config.FullName)` would have dropped it.

## Test plan

- [ ] 5 new unit tests in `Navigation.Tests/Codex/CodexConfigurationLoaderTests.cs`: file not found, missing environment, valid config, no-environment-required variant, out-of-scope file (the original bug)
- [ ] `./build.sh unit-test` passes (all 2651 tests green)
- [ ] Manual: `dotnet run --project src/tooling/docs-builder -- codex clone ~/Projects/codex-environments/environments/internal/config.yml` now logs `Cloning N documentation sets to …` and clones repos (requires SSH access to `elastic/codex-link-index`)
- [ ] Manual: passing a nonexistent or out-of-scope config path now prints a real error message and exits non-zero

### DIFF
--- a/src/Elastic.Codex/CodexConfigurationLoader.cs
+++ b/src/Elastic.Codex/CodexConfigurationLoader.cs
@@ -1,0 +1,77 @@
+// Licensed to Elasticsearch B.V under one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+using System.IO.Abstractions;
+using Elastic.Documentation.Configuration.Codex;
+using Elastic.Documentation.Diagnostics;
+
+namespace Elastic.Codex;
+
+internal static class CodexConfigurationLoader
+{
+	/// <summary>
+	/// Loads a <see cref="CodexConfiguration"/> and its required <paramref name="environment"/>
+	/// field from <paramref name="configFile"/>. Returns <c>false</c> and emits a diagnostic on
+	/// any failure (file missing, unreadable, missing environment). Use <c>out _</c> to discard
+	/// the environment when the caller resolves it from other sources.
+	/// </summary>
+	internal static bool TryLoad(
+		IFileInfo configFile,
+		string originalPath,
+		IDiagnosticsCollector collector,
+		out CodexConfiguration config,
+		out string environment)
+	{
+		environment = string.Empty;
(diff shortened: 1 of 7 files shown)


theletterf (approved):
Thanks!

reakaleek (approved):

PR #3703, opened 2026-07-23
Tolerate removed bundle.resolve in changelog config

vishaangelova:
## Why

PR #3700 removed the `bundle.resolve` changelog config key. The changelog config is parsed by a strict YAML deserializer that hard-errors on any unknown key, so every consuming repo whose `docs/changelog.yml` still carries `resolve:` now fails `validate changelog` on **every** PR — including PRs that don't touch changelogs — with:

```
YAML parsing error in changelog configuration: Property 'resolve' not found on type
Elastic.Documentation.Configuration.Changelog.BundleConfigurationYaml
```

This already broke CI across elastic/cloud (and potentially other repos that were seeded with `resolve: true`). Removing a config key should degrade gracefully, not hard-break downstream builds with no migration window.

## What

Re-add `bundle.resolve` to the YAML DTO as a deprecated, ignored field and emit a deprecation warning when it is present, instead of failing to parse. The key stays out of the domain model (`BundleConfiguration`) and out of `changelog.example.yml`, so it is accepted for backward compatibility only and authors are nudged to remove it. This mirrors the existing `rules.publish` deprecation-warning pattern.

Consuming repos should still drop the dead key (e.g. elastic/cloud#157322), but they are no longer blocked while they do.

### DIFF
--- a/src/Elastic.Documentation.Configuration/Changelog/ChangelogConfigurationYaml.cs
+++ b/src/Elastic.Documentation.Configuration/Changelog/ChangelogConfigurationYaml.cs
@@ -282,6 +282,13 @@ internal sealed record BundleConfigurationYaml
 	public bool? UseLocalChangelogs { get; set; }
 
 	/// <summary>
+	/// Deprecated and ignored. Resolved bundles are now the only format. Retained solely so existing
+	/// changelog.yml files carrying <c>bundle.resolve</c> deserialize without a hard error; a
+	/// deprecation warning is emitted when it is present. Remove it from changelog.yml.
+	/// </summary>
+	public bool? Resolve { get; set; }
+
+	/// <summary>
 	/// Default bundle description used when no profile-specific description is provided.
 	/// </summary>
 	public string? Description { get; set; }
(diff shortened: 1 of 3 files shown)


reakaleek (approved):

reakaleek (approved):
Good fix. thank you @vishaangelova 

cc @cotti @lcawl

src/Elastic.Documentation.Configuration/Changelog/ChangelogConfigurationYaml.cs
```diff
@@ -281,6 +281,13 @@ internal sealed record BundleConfigurationYaml
 	/// </summary>
 	public bool? UseLocalChangelogs { get; set; }
 
+	/// <summary>
+	/// Deprecated and ignored. Resolved bundles are now the only format. Retained solely so existing
+	/// changelog.yml files carrying <c>bundle.resolve</c> deserialize without a hard error; a
+	/// deprecation warning is emitted when it is present. Remove it from changelog.yml.
+	/// </summary>
+	public bool? Resolve { get; set; }
```
Mpdreamz, line 290: Add obsolete attibute

Mpdreamz (approved):
One nitpick, not blocking thanks!

src/Elastic.Documentation.Configuration/Changelog/ChangelogConfigurationYaml.cs
vishaangelova: Added in https://github.com/elastic/docs-builder/pull/3703/changes/6a5fb6abf295363f28582b28ea7752728b059f75

PR #3705, opened 2026-07-23
Add page feedback collection

reakaleek:
Readers can leave thumbs and short comments on docs and API pages without leaving the page. Teams get page-level signals in Elasticsearch instead of scattered GitHub issues.

**Affects:** docs site (Markdown + API reference), docs API

**Prompt summary:** Finish the page-feedback PR and use it for API feedback instead of wiring a GitHub `api-feedback` issue template (https://github.com/elastic/docs-eng-team/issues/819).

## Why

API and Markdown pages lack a shared on-page way to capture whether content helped. The old Bump Feedback path opened a GitHub form; Search/Ask AI used the wrong template. Teams need structured reactions they can query without sending readers off-site.

Closes elastic/docs-eng-team#819

#### Page feedback control
Assembler and Codex pages render a `page-feedback` web component under the article. Users pick thumbs up or down, optional reasons, and a comment. They can revise or revoke within the session. The control stays off when air-gapped or when `PAGE_FEEDBACK` is disabled.

#### API and storage
New PUT/DELETE `/page-feedback/{id}` endpoints upsert and delete documents through Elastic Ingest (`DirectWriteAsync`). Mapping bootstrap is version-aware. Staging, preview, and `dev` enable the flag in `assembler.yml`.

#### API reference pages
API layouts include the same `_PageFeedback` partial as Markdown, so API chrome collects real feedback rather than linking to a GitHub issue template.

image

## Verify

```bash
dotnet test tests/Elastic.Documentation.Api.Tests/
cd src/Elastic.Documentation.Site && npm run test -- --testPathPattern=PageFeedback
```

Enable `PAGE_FEEDBACK` locally or use staging/preview with API explorer. Confirm an API page shows the feedback control and a successful reaction does not open GitHub.

**Out of scope:** App-level rate limiting and Cloudflare Turnstile. Staging relies on existing WAF limits; revisit before production rollout.

**Risk:** Writes feedback documents into an environment-specific Elasticsearch index. A code revert does not delete already-written documents.


theletterf:
+CC @technige A good candidate for the "Other" slide :)

src/Elastic.Documentation.Configuration/ElasticsearchTransportFactory.cs
```diff
@@ -0,0 +1,39 @@
...
+// See the LICENSE file in the project root for more information
+
+using Elastic.Documentation.Configuration;
+using Elastic.Transport;
+using Elastic.Transport.Products.Elasticsearch;
+
+namespace Elastic.Documentation.Api.Adapters;
+
+internal static class ElasticsearchTransportFactory
+{
```
Mpdreamz, line 11: This is a copy of `src/tooling/essc/Elasticsearch/ElasticsearchTransportFactory.cs` — same code, different namespace.

The factory belongs in `Elastic.Documentation.Configuration`, which already owns `ElasticsearchEndpoint`. Both essc and the API project can then import it from there.

More importantly, the API project needs a **shared DI singleton** for the transport so adapters don't each create their own connection pool. `ElasticsearchAskAiMessageFeedbackGateway` already builds its own `ElasticsearchClient` in-constructor from the same endpoint, so right now there are two separate connection pools to the same cluster:

```csharp
// in AddElasticDocsApiServices
services.AddSingleton(sp =>
    ElasticsearchTransportFactory.Create(
        sp.GetRequiredService<DocumentationEndpoints>().Elasticsearch));
```

With that, `PageFeedbackTransport` disappears and `ElasticsearchAskAiMessageFeedbackGateway` can drop its bespoke construction.

src/api/Elastic.Documentation.Api/Adapters/PageFeedback/ElasticsearchPageFeedbackGateway.cs
```diff
@@ -0,0 +1,96 @@
...
+) : IPageFeedbackService
+{
+	public async Task<bool> UpsertFeedbackAsync(PageFeedbackRecord record, CancellationToken ctx)
+	{
+		var document = new PageFeedbackDocument
+		{
+			FeedbackId = record.FeedbackId.ToString(),
+			PageUrl = record.PageUrl,
+			PageTitle = record.PageTitle,
+			Reaction = record.Reaction == PageFeedbackReaction.ThumbsUp ? "thumbsUp" : "thumbsDown",
```
Mpdreamz: `PageFeedbackReaction` and `PageFeedbackReason` already declare `[JsonStringEnumMemberName]` with the correct camelCase values. Store the enum types directly on `PageFeedbackDocument` and let the serializer do the mapping:

```csharp
// PageFeedbackDocument
[Keyword]
[JsonPropertyName("reaction")]
public required PageFeedbackReaction Reaction { get; init; }

[Keyword]
[JsonPropertyName("reason")]
[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
public PageFeedbackReason? Reason { get; init; }
```

Register the enum types in `PageFeedbackJsonContext`, then here:

```csharp
Reaction = record.Reaction,
Reason = record.Reason,
```

`GetReasonValue` disappears. The serializer writes `"thumbsUp"` / `"solvedProblem"` etc. from the attributes — no separate mapping to maintain.

src/api/Elastic.Documentation.Api/Adapters/PageFeedback/ElasticsearchPageFeedbackGateway.cs
```diff
@@ -0,0 +1,96 @@
...
+			Reason = GetReasonValue(record.Reason),
+			ReasonSetVersion = record.ReasonSetVersion,
+			Comment = record.Comment,
+			Euid = record.Euid,
+			Timestamp = DateTimeOffset.UtcNow
+		};
+
+		try
+		{
+			var json = JsonSerializer.Serialize(document, PageFeedbackJsonContext.Default.PageFeedbackDocument);
```
Mpdreamz: 0.51.0 added `DirectWriteAsync` for exactly this use-case: a single-document write in an API handler that needs to await persistence before returning. Register `IngestChannel<PageFeedbackDocument>` as a DI singleton (sharing the `MappingContext` that bootstrap already uses) and inject it here:

```csharp
var response = await channel.DirectWriteAsync(
    new[] { document },
    retries: 2,
    cancellationToken: ctx);

if (!response.AllItemsPersisted())
{
    if (response.TryGetServerErrorReason(out var reason))
        logger.LogWarning("Failed to index page feedback {FeedbackId}: {Reason}", record.FeedbackId, reason);
    return false;
}
return true;
```

Eliminates the manual URL construction, `JsonSerializer.Serialize`, `PostData.String`, and the `StringResponse` cast.

src/api/Elastic.Documentation.Api/Adapters/PageFeedback/PageFeedbackBootstrapService.cs
```diff
@@ -0,0 +1,36 @@
...
+internal sealed class PageFeedbackBootstrapService(
+	PageFeedbackTransport transport,
+	PageFeedbackIndex index,
+	AppEnvironment appEnvironment,
+	ILogger<PageFeedbackBootstrapService> logger
+) : IHostedService
+{
+	public async Task StartAsync(CancellationToken cancellationToken)
+	{
+		var options = new IngestChannelOptions<PageFeedbackDocument>(transport.Transport, index.MappingContext);
```
Mpdreamz: If `IngestChannel<PageFeedbackDocument>` is registered as a DI singleton (as suggested on the gateway), inject it here instead of creating a transient channel just for bootstrap. Both the bootstrap service and the gateway then share the same channel instance.

src/api/Elastic.Documentation.Api/Adapters/PageFeedback/PageFeedbackMapping.cs
```diff
@@ -0,0 +1,70 @@
...
+	[Timestamp]
+	[JsonPropertyName("@timestamp")]
+	public required DateTimeOffset Timestamp { get; init; }
+}
+
+[JsonSerializable(typeof(PageFeedbackDocument))]
+internal sealed partial class PageFeedbackJsonContext : JsonSerializerContext;
+
+[ElasticsearchMappingContext(JsonContext = typeof(PageFeedbackJsonContext))]
+[Index<PageFeedbackDocument>(
```
Mpdreamz: Add `MappingVersionFromAssembly = true` for version-aware bootstrap guards:

```csharp
[Index<PageFeedbackDocument>(
    NameTemplate = "page-feedback-v1-{env}",
    Dynamic = false,
    MappingVersionFromAssembly = true
)]
```

The generated code reads `<Version>` from the assembly and stores it in `_meta.mapping_version` on the template. During a rolling deploy, a pod still on the older version will skip bootstrap instead of overwriting the newer template with older mappings. Docs: https://elastic.github.io/elastic-ingest-dotnet/strategies/bootstrap/#version-aware-bootstrap-guards

Mpdreamz (commented):
Good first pass — the feature shape is right. Four concrete follow-ups below.

**Pre-existing note** (`ElasticsearchAskAiMessageFeedbackGateway.cs:71`): `record.Reaction.ToString().ToLowerInvariant()` bypasses whatever `[JsonStringEnumMemberName]` attributes the reaction enum declares. Worth fixing alongside the `PageFeedbackDocument` change.

src/Elastic.Codex/_MarkdownLayout.cshtml
```diff
@@ -33,6 +33,7 @@
 						@await RenderBodyAsync()
 					</article>
 					@await RenderPartialAsync(_PrevNextNav.Create(Model))
+					@await RenderPartialAsync(_PageFeedback.Create(Model))
```
Mpdreamz, line 35: Placement concern: this renders *below* the prev/next nav, so the reader has to scroll past the full article and navigation before seeing the prompt. Consider placing it in the TOC sidebar (`_TableOfContents`) above "On this page" — that position is always visible while reading and should meaningfully improve response rates.

src/Elastic.ApiExplorer/_Layout.cshtml
```diff
@@ -23,6 +23,7 @@ else
 						<input type="checkbox" class="hidden" id="pages-nav-hamburger">
 						@await RenderBodyAsync()
 					</article>
+					@await RenderPartialAsync(_PageFeedback.Create(Model))
```
Mpdreamz, line 34: Same placement concern as `_MarkdownLayout.cshtml` — the widget is inside `<main>` below the article, while `_ApiToc` renders as a sibling outside it. Moving it above the TOC would keep it visible without scrolling.

src/Elastic.Documentation.Configuration/ElasticsearchTransportFactory.cs
reakaleek: Fixed in 2ea0737f47df759fe98f6496a3fee4be0a185ab4 — moved transport creation and the endpoint model into Elastic.Documentation.Configuration, removed the duplicate factories, and registered one shared API transport used by both Ask AI and page feedback.

src/api/Elastic.Documentation.Api/Adapters/PageFeedback/ElasticsearchPageFeedbackGateway.cs
reakaleek: Fixed in e56fe23c85ba55b305e4763de59b12e8aca61283 — PageFeedbackDocument now stores the enum types directly, registers them with the source-generated JSON context, and removes the duplicate string mapping.

src/api/Elastic.Documentation.Api/Adapters/PageFeedback/PageFeedbackMapping.cs
reakaleek: Fixed in a54536e7fb1873c80f42eac6a6b6b7ed8b77539d — the latest commit added MappingVersionFromAssembly = true and an assertion that the generated mapping version matches the API assembly version.

shainaraskas:
does this feedback get exposed to writers?

src/api/Elastic.Documentation.Api/Adapters/PageFeedback/ElasticsearchPageFeedbackGateway.cs
reakaleek: Fixed in a54536e7fb1873c80f42eac6a6b6b7ed8b77539d — page feedback upserts now use DirectWriteAsync and verify AllItemsPersisted before returning. I omitted the channel retry overload intentionally because the browser already owns the bounded retry behavior for this endpoint.

src/api/Elastic.Documentation.Api/Adapters/PageFeedback/PageFeedbackBootstrapService.cs
reakaleek: Fixed in a54536e7fb1873c80f42eac6a6b6b7ed8b77539d — IngestChannel<PageFeedbackDocument> is now a DI singleton shared by PageFeedbackBootstrapService and ElasticsearchPageFeedbackGateway.

reakaleek:
> does this feedback get exposed to writers?

That's the goal

src/Elastic.Codex/_MarkdownLayout.cshtml
reakaleek: Updated in 517db411b6bdd34f261dc14cf01d9f8b63df759c — the feedback component now appears immediately after the page content and before the previous/next navigation. This presents the prompt at the natural completion point without making it a persistent distraction in the TOC sidebar. This placement was synced with Alex.

src/Elastic.ApiExplorer/_Layout.cshtml
reakaleek: We are keeping the feedback component after the page content rather than in the TOC sidebar. This presents it at the natural completion point, keeps placement consistent across page types, and avoids competing with in-page navigation while the reader is still consuming the content. This placement was synced with Alex.

reakaleek: Also, IMO our sidebar is already too stuffed.

And there is an existing convention that this kind of feedback widget is at the bottom.

So users will naturally search for it there

theletterf: Bottom of page seems the most common convention these days — also helpful in mobile views!

Mpdreamz (dismissed):
Should we introduce rate limiting to the API at this point?

https://learn.microsoft.com/en-us/aspnet/core/performance/rate-limit?view=aspnetcore-10.0 

Should we add cloudflare turnstile?

reakaleek:
> Should we introduce rate limiting to the API at this point?
> 
> https://learn.microsoft.com/en-us/aspnet/core/performance/rate-limit?view=aspnetcore-10.0
> 
> Should we add cloudflare turnstile?

We have a general rate limit in WAF. But it's definitely worth double-checking in docs-infra.

theletterf:
Before rolling this out in Prod, we should follow-up with a task to define the feedback options as a team.

reakaleek:
> Before rolling this out in Prod, we should follow-up with a task to define the feedback options as a team.

SGTM. But I definitely want to get this into staging to verify it functionally.

reakaleek:

reakaleek:
> > @copilot resolve the merge conflicts in this pull request
> 
> Resolved in `bd1c9e5` by merging `main` into this branch and completing conflict resolution; the branch is now clean with no remaining conflict markers.

There is still a conflict

theletterf:
What's the status of this one?

reakaleek:
## Rebase update

Rebased onto current `main` (`360fc520`). Conflicts resolved in `assembler.yml`, `docs/development/toc.yml`, API mappings, web host, package props, and transport-factory consolidation (kept main deletions for Labs/SearchBootstrap; shared `Elastic.Documentation.Configuration` transport for API adapters).

API layouts still render `_PageFeedback` (gated by `PAGE_FEEDBACK`). Flag remains on for staging, preview, and local `dev`.

### Review follow-ups
Inline threads from the earlier pass (enums, `DirectWriteAsync`, shared `IngestChannel`, `MappingVersionFromAssembly`) are still in place after the rebase and are marked resolved.

### Rate limiting / Turnstile (`CHANGES_REQUESTED`)
Staging/preview traffic to this API already sits behind Cloudflare WAF with the site's general rate limiting. Adding ASP.NET rate limiting or Turnstile in this PR would duplicate that control plane for a low-volume feedback PUT without changing the abuse story. Prefer keeping abuse controls at the WAF edge for now; happy to revisit app-level limits if we see real abuse once this ships to staging.

Verified locally after rebase: `dotnet build`, `dotnet test tests/Elastic.Documentation.Api.Tests/` (19 passed), Jest `PageFeedback` (7 passed).

src/api/Elastic.Documentation.Api/MappingsExtensions.cs
```diff
@@ -173,4 +175,98 @@
...
+				request.Reaction,
+				request.Reason,
+				request.ReasonSetVersion,
+				comment,
+				euid
+			);
+
+			if (!await feedbackService.UpsertFeedbackAsync(record, ctx))
+			{
+				logger.LogWarning("Failed to record page feedback {FeedbackId} for {PageUrl}", feedbackId, request.PageUrl);
```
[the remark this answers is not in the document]
reakaleek: Fixed in 9c597b40ded4d62a02918b00547d09006c2af7b5 — strip CR/LF from `PageUrl` before logging, matching the AskAI message sanitization pattern in the same file.

src/api/Elastic.Documentation.Api/MappingsExtensions.cs
```diff
@@ -173,4 +175,98 @@
...
+				return Results.StatusCode(StatusCodes.Status503ServiceUnavailable);
+			}
+
+			logger.LogInformation("Recorded page feedback {FeedbackId} for {PageUrl}", feedbackId, request.PageUrl);
```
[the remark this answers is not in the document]
reakaleek: Fixed in 9c597b40ded4d62a02918b00547d09006c2af7b5 — same sanitization as the warning log on the failure path; both sites now use a CR/LF-stripped `pageUrlForLog`.

theletterf (approved):

PR #3709, opened 2026-07-23
Add static search to isolated documentation builds

reakaleek:
## Why

Standalone documentation sites currently have no search unless they depend on a backend service. They need a search option that continues to work when the generated frontend is hosted as static files on services such as S3 and CloudFront, or when previewing locally with `docs-builder serve`.

## What

Adds Pagefind-powered client-side search to all isolated builds and `docs-builder serve` mode. The existing search modal queries the generated client-side index while omitting backend-only Ask AI features.

This PR represents combined work:
- The initial implementation (this PR) added the search UI integration, feature flag, frontend pagefind query layer, HTML metadata annotations, responsive header tweaks, and the build pipeline wiring.
- #3721 (merged) replaced the native pagefind binary subprocess with [Pagefind.Net](https://www.nuget.org/packages/Pagefind.Net) — a pure managed .NET reimplementation of the indexer. This made the feature a proper \`IMarkdownExporter\` that eagerly indexes each page during the build pipeline (O(1) memory), removed the feature flag (search is always available for isolated builds), and enabled full \`docs-builder serve\` support where the index is built in-memory as part of the background compilation that surfaces diagnostics to the HUD.

## How

The \`PagefindMarkdownExporter\` runs inline during the build pipeline alongside other exporters. For each markdown page it extracts plain text via \`PlainTextExporter\`, builds weighted segments (title, headings, body), and feeds a \`PagefindRecord\` to \`PagefindIndex.AddRecord\` which tokenizes and discards the content immediately. At the end, \`WriteAsync\` flushes the pre-built inverted index to disk and \`PagefindFrontend.ExtractToAsync\` writes the browser runtime (\`pagefind.js\` + WASM).

The exporter is gated on \`BuildType.Isolated\` — assembler and codex builds skip it entirely (they use the Elasticsearch-backed search API). In serve mode, the index resides in the in-memory filesystem and is served via a dedicated route handler.


reakaleek:
@Mpdreamz Clicking search result items doesn't work. I will take a look

Mpdreamz:
Oh, maybe an artifact of merging in? I could swear it worked. 

Good catch either way!

Mpdreamz:
## Pagefind.Net 0.4.1 — meta field ordering fix

Updated to [Pagefind.Net 0.4.1](https://www.nuget.org/packages/Pagefind.Net/0.4.1) which fixes [nullean/pagefind-net#15](https://github.com/nullean/pagefind-net/pull/15).

**The bug:** meta field IDs were assigned per-record based on alphabetical ordering of that record's meta keys. Since we store both `breadcrumbs` and `title`, breadcrumbs (alphabetically first) got fieldId 0 — which the global meta array mapped to "title". This meant breadcrumb content was falsely boosted as title matches, and actual title text lost its ranking boost.

**If results are too noisy:** 0.4.1 also adds an `IndexedMetaFields` option to `PagefindIndexOptions`. If we find breadcrumbs content polluting search results, we can restrict meta indexing to just title:

```csharp
new PagefindIndexOptions { Language = "en", IndexedMetaFields = ["title"] }
```

This would keep breadcrumbs in the fragment metadata (for display in results) but exclude it from the searchable `meta_locs` postings.

Mpdreamz:
## Summary of changes added to this PR

### Pagefind.Net 0.4.0 → 0.4.1

Upgraded to fix multiple bugs in the .NET pagefind indexer:
- **0.4.0** ([nullean/pagefind-net#14](https://github.com/nullean/pagefind-net/pull/14)): Fixes encoding, PascalCase splitting, Porter2 stemmer, per-position weights, and meta indexing. Adds `AddHtmlRecord` API with weight scheme matching the official binary.
- **0.4.1** ([nullean/pagefind-net#15](https://github.com/nullean/pagefind-net/pull/15)): Fixes meta field ID ordering that caused breadcrumbs to be falsely boosted as title matches.

### Switch to `AddHtmlRecord` API

Replaced `AddRecord` + `WeightedSegments` with `AddHtmlRecord` + `HtmlPageData`. This ensures heading words are always searchable (fixes multi-word queries like "syntax guide" failing) and assigns correct per-position weights (h1=168, h2=144, ..., body=24) matching the official pagefind binary.

### Fix CLI pages indexed twice

CLI commands with supplemental `cmd-*.md` files were registered twice in `DocumentationSet.Files` — once from the main directory scan (under the physical `cmd-*.md` path) and once from the extension scan (under the clean synthetic path). Both pointed to the same `CliCommandFile` instance, so each page was exported/indexed twice, inflating search scores for CLI reference pages.

Fixed by:
1. Factory now registers markdown files under `doc.SourceFile` (the canonical path) so CLI commands get their clean URL-correct path directly from the main scan.
2. Extension's `ScanDocumentationFiles` skips files already in `_createdFiles` to avoid re-registration.

Result: 20 duplicate index entries eliminated (188 → 168 unique pages).

Mpdreamz:
## Index size optimization & follow-up

### Content limits (this commit)

Added caps to the pagefind indexer: max **10 headings** and **5 body sections** per page. Pages still have title + key headings + opening content indexed — sufficient for discovery without the long tails of CLI option tables and reference lists.

Also set `IndexedMetaFields = ["title"]` so breadcrumbs stay in fragment metadata (for display in search results) but are excluded from searchable `meta_locs` postings.

### Impact

| Metric | Before | After | Reduction |
|---|---|---|---|
| Index chunks (loaded per query) | 236 KB | 148 KB | **-37%** |
| Total indexed words | 73,115 | 43,198 | **-41%** |
| Fragments (per result) | 724 KB total | 676 KB total | -7% |
| Duplicate pages | 20 | 0 | fixed |

### What loads when

Pagefind adds **zero bytes to initial page load** — everything is lazy:
- `pagefind.js` (45 KB) — dynamic `import()` on first search
- WASM (73 KB) — loaded after pagefind.js init
- `pf_meta` (1.5 KB) — after WASM
- Index chunks (~37 KB each now) — per query, only matching chunks
- Fragments (~4 KB each) — per result shown

### Follow-up: header load time

The slow uncached header isn't pagefind — it's the **React/EUI bundle** loaded synchronously via `isolated.ts` → `Header/Header`. The `<ModalSearch>` component (with all of EUI, zustand, tanstack-query, emotion) is part of the eager header bundle even though search is rarely the first interaction.

Potential improvements (not in this PR):
- Lazy-load the `<ModalSearch>` component (render a placeholder search box, load the real React component on focus/click)
- Code-split the EUI components used only in the search modal
- Consider showing the search trigger as a static element that hydrates on interaction

Mpdreamz (approved):

PR #3714, opened 2026-07-23
Reject changelog bundles whose target diverges from entry targets

vishaangelova:
## Problem

A serverless breaking-change rendered on the public site under **July 20, 2027** (a year in the future): https://www.elastic.co/docs/release-notes/cloud-serverless/breaking-changes#july-20-2027

Root cause: the CDN bundle `bundle/cloud-serverless/elasticsearch-2026-07-20.yaml` had a bundle-level product target of `2027-07-20`, even though its filename, `release-date`, and **every one of its entries** declared `cloud-serverless` target `2026-07-20`. The `{changelog}` directive derives its date heading from `bundle.Products[0].Target` (`BundleLoader.GetVersionFromBundle`), so the whole bundle rendered under the wrong year.

The defect was possible because `BundleBuilder.BuildProducts` takes the bundle-level product target **verbatim** from the release-version argument (`--output-products` / profile `{version}`) and never validates it against the targets the entries themselves declare.

cc: @lcawl @cotti 

## Fix

Add a guardrail in `BundleBuilder.BuildBundle`: if a bundle-level product target is inconsistent with the targets its own entries declare for that same product, the build fails (returns an invalid result, so no bundle is written and the publish job fails loudly).

Compatibility is intentionally lenient in one direction to avoid false positives:

- A coarser bundle target may be a **component-prefix** of a finer entry target — e.g. a monthly rollup `2026-05` covering an entry dated `2026-05-15`. ✅ allowed
- Genuine divergence is rejected — e.g. `2027-07-20` vs `2026-07-20`, or `9.5.0` vs `9.6.0`. ❌ error

Entries whose product carries no target, or a product the bundle doesn't declare, are not compared.

## Tests

Adds `BundleTargetConsistencyTests`:
- divergent target (the real-world `2027` vs `2026` case) → bundle fails with a descriptive error, no file written
- coarser prefix (`2026-05` over `2026-05-15`) → succeeds
- exact match (`9.5.0`) → succeeds

Full `Elastic.Changelog.Tests` suite passes (807 tests).

## Note

This prevents recurrence. The already-published CDN bundle still needs to be corrected separately (re-publish the cloud-serverless bundle for `elasticsearch` with the correct `2026-07-20` target, then invalidate the CDN cache).

### DIFF
--- a/src/services/Elastic.Changelog/Bundling/BundleBuilder.cs
+++ b/src/services/Elastic.Changelog/Bundling/BundleBuilder.cs
@@ -46,6 +46,19 @@ public class BundleBuilder
 			};
 		}
 
+		// Guard against a bundle-level target that disagrees with the targets its own entries declare
+		// for the same product (for example a release version typed as 2027-07-20 while every entry says
+		// 2026-07-20). The bundle-level target drives the rendered release date, so a mismatch silently
+		// publishes entries under the wrong version/date. Refuse to build the bundle in that case.
+		if (!ValidateProductTargetConsistency(collector, bundledProducts, entries))
+		{
+			return new BundleBuildResult
+			{
+				IsValid = false,
+				Data = null
+			};
+		}
+
 		var bundledData = new Bundle
 		{
 			Products = bundledProducts,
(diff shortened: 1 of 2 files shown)


vishaangelova:
It seems this PR might be irrelevant if we don’t need to use `docs-builder changelog upload` any longer?

Mpdreamz:
@vishaangelova i believe so, also the changelog targets are aspirational (primarely used for future feature where we can share release notes ahead of time). I believe this makes them being different from the bundle target a feature not a bug..

Thanks for taking the time though, looking forward to more PR's! 😸

Issue #670, opened 2025-03-04
[Feature Request]: A way to redirect from one page to another just for specific anchors


shainaraskas:
### Prerequisites

- [x] I have searched existing issues to ensure this feature hasn't already been requested
- [x] I have tested using the latest version of docs-builder

### What problem are you trying to solve?

I have a page: 

https://docs-v3-preview.elastic.dev/elastic/docs-content/tree/main/deploy-manage/users-roles/cluster-or-deployment-auth/defining-roles

I moved a subset of the content that was planned for that page to a subpage:

https://docs-v3-preview.elastic.dev/elastic/docs-content/tree/main/deploy-manage/users-roles/cluster-or-deployment-auth/role-structure

we have links pointing to the anchors that have been moved to the subpage: 

-  roles-indices-priv
-  roles-global-priv
-  roles-application-priv
-  roles-remote-indices-priv
-  roles-remote-cluster-priv

we need a way to redirect those specific anchors to the new subpage, without redirecting the entire parent page to the new subpage.

### Proposed Solution

?

### How important is this feature to you?

Important

PR #3715, opened 2026-07-23
Add safe and selective backfill mode to changelog upload

cotti:
## Summary

Delivers the upload-path prerequisites for the release-notes backfill (items 1–3 of the epic's "Current system and prerequisites"): `changelog upload --backfill` now provides explicit file selection, create-only bundle writes, and strict registry failure semantics.

- **Explicit selection** (`ChangelogUploadService`): in backfill mode the operator supplies the exact bundle YAMLs via `--files` (comma-separated paths or a newline-delimited path list file, aligned with `changelog bundle --files`); there is no directory globbing. A selected file that cannot be mapped to a destination — missing, no `products`, invalid product — is an error that aborts the run before any write, never a silent skip.
- **Create-only writes** (`S3IncrementalUploader`): a new `S3WritePolicy.CreateOnly` writes bundle objects with a conditional PUT (`If-None-Match: *`). Identical remote content → skip (reported as such); different content → conflict, surfaced per key, never an overwrite. The ETag pre-check (existing `S3EtagCalculator`) is informative only — the conditional PUT is the race guard, so a concurrent create between inspection and write surfaces as a conflict (`412 PreconditionFailed`). There is intentionally no force flag; corrections are a separately approved workflow.
- **Registry failure semantics**: in backfill mode an unreconciled registry (S3 error or exhausted optimistic-concurrency retries) is an incomplete operation — it emits a clear error and fails the run. Conflicted objects are excluded from the registry refresh so the manifest never records content that is not actually published.

## Design notes

- **How backfill mode is surfaced**: a single `--backfill` flag switches the mode, with `--files` as its required selection input. The pairing is validated in `ChangelogUploadService` (testable regardless of caller): `--backfill` requires `--files` and `--artifact-type bundle` (the backfill publishes bundles only, per the epic), and rejects `--skip-etag-check`, whose forced re-upload semantics would turn every idempotent re-run into a conflict. `--files` without `--backfill` is rejected so live behavior cannot drift.
- **Why live behavior is untouched**: `S3WritePolicy` defaults to `Overwrite` and the create-only path is only reachable via `args.Backfill`; without the flag, discovery, overwrite, and best-effort registry refresh are byte-for-byte the previous code paths. Registries keep their existing concurrent conditional-update path (`If-Match`/`If-None-Match` + bounded retries) in both modes. Proven by the untouched pre-existing upload tests plus new explicit tests (`Upload_DefaultPolicy_DoesNotUseConditionalPut`, `Upload_Live_RegistryRefreshException_DoesNotFailUpload`).
- `docs/cli-schema.json` regenerated; supplemental docs updated (`docs/cli/changelog/cmd-upload.md`, `docs/development/changelog-bundle-registry.md`). Expected to conflict on the schema with the concurrent #670 PR — whoever merges second rebases and regenerates.

## Test plan

New tests in `Elastic.Documentation.Integrations.Tests` (uploader) and `Elastic.Changelog.Tests` (service orchestration), following the existing FakeItEasy `IAmazonS3` fake + `MockFileSystem` patterns:

- Exact selection: unselected files in the same directory are not uploaded; relative paths resolve against the bundle directory; multi-product bundles fan out per product.
- Create-only conflict: existing key with different content → run fails, object untouched, no registry write for the conflicted target.
- Skip on identical content: reported as skip, run succeeds, registry still reconciled.
- Concurrency race: conditional PUT loses (`412`) → conflict, not overwrite.
- Registry refresh failure (S3 exception and retry exhaustion) fails the backfill operation; the same failure in live mode remains a warning.
- Mode guards: `--backfill` without `--files`, with `changelog` artifact type, or with `--skip-etag-check`; `--files` without `--backfill`; missing selected file and productless selected bundle fail before any write.

Checks run locally: `dotnet format` clean · `./build.sh build --skip-dirty-check` ✓ · AOT publish with zero trim/AOT warnings ✓ · `./build.sh unit-test` 4,247 passed / 0 failed ✓ · schema diff empty ✓.

Closes elastic/docs-eng-team#668

### DIFF
--- a/docs/cli-schema.json
+++ b/docs/cli-schema.json
@@ -4305,7 +4305,7 @@
           ],
           "name": "upload",
           "summary": "Upload changelog entries or bundle artifacts to S3 or Elasticsearch.",
-          "notes": "Uses content-hash\u2013based incremental transfer \u2014 only changed files are uploaded.\n\n\nChangelog entries are uploaded once under changelog/{org}/{repo}/{branch}/{file}, keyed by the\nauthoring owner (--owner \u003E bundle.owner \u003E git remote), repository (--repo\n\u003E bundle.repo \u003E git remote), and branch (--branch \u003E the current checkout\u0027s\nbranch). The branch is stored verbatim, so a branch\u0027s / become real key separators. Bundles\nare uploaded under bundle/{product}/{file}, product-scoped from the bundle YAML, and do not\nrequire an owner/repo/branch.",
+          "notes": "Uses content-hash\u2013based incremental transfer \u2014 only changed files are uploaded.\n\n\nChangelog entries are uploaded once under changelog/{org}/{repo}/{branch}/{file}, keyed by the\nauthoring owner (--owner \u003E bundle.owner \u003E git remote), repository (--repo\n\u003E bundle.repo \u003E git remote), and branch (--branch \u003E the current checkout\u0027s\nbranch). The branch is stored verbatim, so a branch\u0027s / become real key separators. Bundles\nare uploaded under bundle/{product}/{file}, product-scoped from the bundle YAML, and do not\nrequire an owner/repo/branch.\n\n\n\nWith --backfill, the command switches to backfill publishing mode for historical bundles:\nonly the files named via --files are uploaded (no directory discovery), objects are written\ncreate-only (an existing key with different content is a conflict, never an overwrite), and a\nregistry manifest that cannot be reconciled fails the operation.",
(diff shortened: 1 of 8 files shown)


cotti:
Closing as superseded: create-only upload semantics remain a hard requirement, but they'll live inside the temporary `changelog migrate-from-web` command (elastic/docs-eng-team#736) rather than as a mode on `changelog upload`. The If-None-Match machinery here will be lifted into that PR.

PR #3717, opened 2026-07-23
Add changelog registry state discovery and reconciliation commands

cotti:
## Summary

`registry.json` manifests (per scope: `bundle/{product}/` and `changelog/{org}/{repo}/{branch}/`) merge additively under optimistic concurrency and are not authoritative for removals or discovery, so they can silently drift from the objects actually in the bucket. Nothing could detect or repair that drift, and backfill planning needs a trustworthy current-state view. This PR adds a `changelog registry` command group backed by a new `Elastic.Changelog/Reconciliation/` service layer:

- **`changelog registry inspect`** *(read-only)* — compares a scope's private registry against the actual private-bucket objects, prints human-readable diagnostics, and optionally writes a machine-readable `RegistryStateSnapshot` (`--out`) that the backfill planner (elastic/docs-eng-team#676) will consume.
- **`changelog registry repair`** *(explicit, never implicit; idempotent)* — converges the **private** registry to the actual objects.
- **`changelog registry verify-public`** *(strictly read-only)* — waits under a bounded retry policy (`--max-attempts` × `--poll-interval-seconds`) for scrubber propagation and diagnoses public-side divergence.
- **`changelog registry republish`** *(explicit)* — re-emits the private `ObjectCreated` event for selected objects to recover a lost/DLQ'd scrub event.

## Divergence taxonomy

| Class | Meaning |
|---|---|
| `missing` | Object exists in the scope, registry has no entry for it. |
| `stale` | Registry entry whose object no longer exists. |
| `corrupt` | Manifest unparseable, or contains unsafe/duplicate entries. |
| `object-divergent` | Entry metadata (ETag or target) disagrees with the actual object — including legacy amends whose registry `target` is null while the parent bundle resolves one. |

A manifest with a newer `schema_version` is reported as `UnsupportedSchema` (not corrupt): its entries can't be judged by an older tool, and repair refuses to rewrite it because that would silently downgrade the schema. For bundle scopes the expected `target` is re-derived from the bundle YAML in S3 (including the parent-bundle fallback for legacy amends, mirroring `RegistryBuilder`); changelog scopes enumerate files only.

## Repair semantics

- Rebuilds the manifest from actual objects through the **same conditional-PUT optimistic concurrency as the live upload path** (`If-Match` on update, `If-None-Match: *` on create). A 412 triggers a full re-inspection — fresh registry read **and** fresh object listing — before the bounded retry, and the registry is always read *before* listing, so a concurrent upload either appears in the re-listing or invalidates the precondition; it can never be dropped.
- Idempotent: a clean scope writes nothing; running repair twice yields no further changes (covered by tests).
- Audited: every added/removed/corrected entry is logged with before/after values; `--dry-run` prints the plan without writing; an empty result requires `--allow-empty`.
- Marked `[RequiresAuth]` + `[CommandIntent(Intent.Idempotent)]` + `[MutationScope(MutationScope.Global)]` (same for `republish`); `inspect`/`verify-public` are `[RequiresAuth]` only.

## Public-boundary design

The public bucket is scrubber-owned and is a **hard write boundary enforced structurally, not by convention**: all inspection and public-side comparison code operates exclusively on a read-only reader interface (`IS3ScopeReader`: list + get only), so no code path in the checker can mutate any bucket. Tests additionally assert zero `PutObject`/`CopyObject`/`DeleteObject` calls on the public-side fake.

Recovery from a lost scrub event is the explicit private-side `republish`: a metadata-preserving S3 self-copy (`CopyObject` onto the same key with `MetadataDirective: REPLACE`, re-supplying the original content type and user metadata) that leaves content and ETag untouched while emitting the `ObjectCreated:Copy` notification the scrubber Lambda reacts to (`src/infra/docs-lambda-changelog-scrubber/Program.cs` matches `ObjectCreated`). The scrubber remains the sole public-side writer.

## Docs

- New supplemental CLI pages under `docs/cli/changelog/registry/` (namespace + 4 commands); `docs/cli-schema.json` regenerated (verified: diff empty ignoring `version`).
- `docs/development/changelog-bundle-registry.md` gains a "State discovery and reconciliation" section, and the `registry.json` example is corrected to the actual serialized key (`e_tag`, not `etag`).

## Test plan

45 new tests in `tests/Elastic.Changelog.Tests/Reconciliation/` using a stateful in-memory S3 fake (FakeItEasy over `IAmazonS3`, MD5 ETags matching `S3EtagCalculator`, real conditional-PUT 412 semantics):

- [x] Each divergence class detected on seeded state: missing, stale, corrupt (unparseable + unsafe entries), object-divergent (ETag and legacy-amend target)
- [x] Clean scope reports clean; missing manifest over an empty scope is clean; missing manifest over a populated scope reports per-object `missing`
- [x] Repair convergence + idempotence (second run writes nothing); create path uses `If-None-Match: *`, update path `If-Match`
- [x] Repair safety under optimistic-concurrency conflict: concurrent registry update between read and write → 412 → re-inspect → concurrent entry survives
- [x] Repair rails: dry-run writes nothing; empty scope refused without `--allow-empty`; newer schema refused
- [x] Public check never writes (asserted on the fake); bounded wait stops exactly at `--max-attempts`; converges early when the scrubber catches up mid-poll
- [x] Republish self-copies only within the private bucket, preserves content/content-type/user metadata, validates file names, `--all` includes `registry.json`, partial failures reported
- [x] Full suite: `Elastic.Changelog.Tests` 849/849 passed; `./build.sh unit-test` green
- [x] `dotnet format` + `./build.sh lint` clean; `./build.sh build --skip-dirty-check` succeeds; `dotnet publish -c Release` (Native AOT) succeeds with zero trim/AOT warnings; repo docs build validates the new supplemental pages

Closes elastic/docs-eng-team#670

### DIFF
--- a/docs/cli-schema.json
+++ b/docs/cli-schema.json
@@ -4433,7 +4433,440 @@
           ]
         }
       ],
-      "namespaces": []
+      "namespaces": [
+        {
+          "segment": "registry",
+          "summary": "Inspect, repair, and verify per-scope changelog registry.json manifests against actual bucket state.",
+          "options": [],
+          "commands": [
+            {
+              "path": [
+                "changelog",
+                "registry"
+              ],
+              "name": "inspect",
+              "summary": "Compare a scope\u0027s private registry.json against the actual private-bucket objects and report every divergence.",
+              "notes": "Registries merge additively and are not authoritative for removals or discovery, so they can drift\nfrom the objects actually in the bucket. This command detects and classifies that drift: missing\n(object exists, registry lacks it), stale (registry entry, object gone), corrupt\n(unparseable/invalid manifest), and object-divergent (registry metadata disagrees with the object).\n\nStrictly read-only \u2014 nothing is written to any bucket. Exits non-zero when the scope diverged.\nUse changelog registry repair to reconcile.",
+              "usage": "docs-builder changelog registry inspect --s3-bucket-name \u003Cstring\u003E [options]",
+              "examples": [],
+              "parameters": [
+                {
+                  "role": "flag",
+                  "name": "s3-bucket-name",
+                  "type": "string",
+                  "required": true,
+                  "summary": "Private changelog bundles S3 bucket to inspect."
+                },
+                {
+                  "role": "flag",
+                  "name": "product",
+                  "type": "string",
+                  "required": false,
+                  "summary": "Product of a bundle scope (bundle/{product}/). Mutually exclusive with --owner/--repo/--branch."
+                },
+                {
+                  "role": "flag",
+                  "name": "owner",
+                  "type": "string",
+                  "required": false,
+                  "summary": "GitHub owner of a changelog scope (changelog/{org}/{repo}/{branch}/). Requires --repo and --branch."
+                },
+                {
+                  "role": "flag",
+                  "name": "repo",
+                  "type": "string",
+                  "required": false,
+                  "summary": "Repository of a changelog scope. Requires --owner and --branch."
+                },
+                {
+                  "role": "flag",
+                  "name": "branch",
+                  "type": "string",
+                  "required": false,
+                  "summary": "Branch of a changelog scope, stored verbatim (slashes become key segments). Requires --owner and --repo."
+                },
+                {
+                  "role": "flag",
+                  "name": "out",
+                  "type": "string",
+                  "required": false,
+                  "summary": "Path to write the machine-readable state snapshot JSON to.",
+                  "validations": [
+                    {
+                      "kind": "rejectSymbolicLinks"
+                    }
+                  ]
+                },
+                {
+                  "role": "flag",
+                  "name": "log-level",
+                  "shortName": "l",
+                  "type": "enum",
+                  "required": false,
+                  "summary": "Minimum log level. Default: information",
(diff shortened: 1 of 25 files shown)


cotti:
Closing as superseded by the registry-ownership stack, per the disposition table in [elastic/docs-eng-team#688](https://github.com/elastic/docs-eng-team/issues/688): the scrubber Lambda is now the sole producer of the public `registry.json` (#3738), which removes the drift this PR's client-side tooling was built to manage.

The pieces that remained valuable were harvested there:
- `FakeS3Bucket` → the `FakeS3` test fake behind the reconciler suite (#3738)
- `verify-public` + the divergence taxonomy (`missing`/`stale`/`corrupt`/`object-divergent`, `UnsupportedSchema` distinct) → `changelog registry verify` (#3741)
- `republish` → subsumed by the reconcile message's full group heal (#3741)
- newer-`schema_version` refusal semantics → carried into the reconciler's skip rules (#3738)

`repair` and `inspect`/`RegistryStateSnapshot` are deliberately not carried over — the first *is* the multi-writer model being removed, and the backfill planner (elastic/docs-eng-team#676) can read the now-authoritative registry directly.

PR #3724, opened 2026-07-24
Frontend: Fix lazy highlight.js aliases and autodetection

reakaleek:
## What

- Load all curated highlight.js grammars when a page or message contains any code block
- Route static docs and React Markdown through one async DOM highlighter
- Add tests for language aliases (`js`, `ts`, `jsx`) and unlabeled fence autodetection

## Why

- The lazy-load change from #3692 broke alias fences such as `js` and `ts`, and reduced autodetection to only eagerly registered grammars
- Authors and AI responses often use short language names, so blocks stayed uncolored

## Notes

- Pages without code blocks still load zero language chunks
- The first code-bearing page loads all curated grammars once per session, then caches them for later swaps and messages

### DIFF
--- a/src/Elastic.Documentation.Site/Assets/hljs.test.ts
+++ b/src/Elastic.Documentation.Site/Assets/hljs.test.ts
@@ -7,12 +7,14 @@ async function loadModule(): Promise<HljsModule> {
     return await import('./hljs')
 }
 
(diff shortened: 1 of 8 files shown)


reakaleek:
Merging to hotfix prod

PR #3734, opened 2026-07-27
Move page actions into breadcrumb bar on smaller viewports

Mpdreamz:
## Why

On medium and smaller viewports the right sidebar (TOC column) is hidden, but the page action links ("Report a docs issue", "Edit this page", version dropdown) were still rendered inside it — either invisible or pushing content down with empty block space. Short-content pages looked especially broken with large gaps above the heading.

## What

The breadcrumb bar now includes the page actions on non-lg screens, right-aligned via `ml-auto` and using `flex-wrap-reverse` so actions sit above breadcrumbs when the row wraps. On lg+ the actions remain in the right sidebar as before — no duplication thanks to paired `lg:hidden` / `hidden lg:block` classes.

Also scopes `justify-end` on the main content container to `lg:` only, fixing a layout issue where content was pushed to the bottom of the flex-col container on medium viewports.

## Before

https://github.com/user-attachments/assets/581d0276-7bda-49f7-b7d8-42d9cf6d0099

## After

https://github.com/user-attachments/assets/77c873f1-daba-4c04-aa3c-a1e52e87cb18

### DIFF
--- a/src/Elastic.Markdown/Layout/_Breadcrumbs.cshtml
+++ b/src/Elastic.Markdown/Layout/_Breadcrumbs.cshtml
@@ -1,3 +1,4 @@
+@using Elastic.Documentation.Configuration.Toc
 @inherits RazorSlice<Elastic.Markdown.MarkdownLayoutViewModel>
 @functions {
 	static string StripBadge(string t) =>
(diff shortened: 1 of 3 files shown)


theletterf:
@Mpdreamz Got a screenshot or before/after comparison?

Mpdreamz:
@theletterf @reakaleek attached before and after videos.

theletterf (dismissed):
@Mpdreamz I think the most valuable items to keep visible on top should be:

- "View as Markdown" (or even better, a Copy to LLM button, if we had one)
- The version selector

The others I would keep hidden or move to the bottom, because I don't expect anyone to edit a doc or report an issue from the mobile view.

In the context of a wider redesign of these buttons, though, I wonder if we really need to touch this now (a patch is fine). @itsalexcm might have ideas on overall interaction elements' placement.

reakaleek:
> @Mpdreamz I think the most valuable items to keep visible on top should be:
> 
> * "View as Markdown" (or even better, a Copy to LLM button, if we had one)
> * The version selector
> 
> The others I would keep hidden or move to the bottom, because I don't expect anyone to edit a doc or report an issue from the mobile view.
> 
> In the context of a wider redesign of these buttons, though, I wonder if we really need to touch this now (a patch is fine). @itsalexcm might have ideas on overall interaction elements' placement.

I would actually disagree. You can easily edit a typo from your phone. But I don't think anyone will open the markdown.

But you touched points on something... a user would definitely use a button like "Ask in ChatGPT/Gemini/Claude" on mobile

reakaleek (approved):

reakaleek:
IMO there should be a bit spacing after the page actions and the content on the mobile view.. but nothing blocking. This is definitely a needed fix.

reakaleek:
I will start a deep research around this topic..

But IIRC, CTAs in the third column are actualy an anti-pattern, because it's the least noticed column.. users actually ignore it.


Let's get the fix in and think about this in a follow-up.

P.S. Seems like there is even a term for it: "[Right-rail blindness](https://www.nngroup.com/articles/fight-right-rail-blindness/)"

Mpdreamz:
++ this whole thing is ripe for re-eval thanks for raising it @theletterf and @reakaleek 

Getting this fix in so we can focus on that.

PR #3739, opened 2026-07-28
Add dropdown appearance to applies-switch for versioned code examples

florent-leborgne:
Docs-eng: treat this as a suggestion/idea

## Why

docs-content needs to document API payloads that changed across versions — for example the dashboards API, with a 9.0 technical preview shape and breaking changes in 9.1/serverless. Duplicating a snippet's surrounding prose per version doesn't scale, and the tabs appearance spends a full-width tab bar on every snippet, which gets noisy on pages with many versioned examples.

## What

`{applies-switch}` accepts `:appearance: dropdown`: a compact selector chip attached to the upper-right corner of each item's code block, opening a floating panel with the alternatives.

- Items take the full `applies_to` syntax; each item carries its own code block and callout list, so callouts always match the selected version.
- The selector shows compact labels (`9.1+`, `ECH 8.0+`, `9.0 (preview)`) derived from the same applicability pipeline as badges; the full definition shows on hover.
- The appearance only changes presentation: switches sharing `applies_to` definitions stay in sync page-wide regardless of appearance (dropdown ↔ tabs), and selection persists per session.
- Guard rails: every item must start with a code block, otherwise the switch warns and falls back to tabs (same pattern as unknown `:appearance:` values). Future content types can be supported by adapting the chip's anchoring, without new syntax.
- `:selected:` now sets the default item (previously parsed but ignored); multiple `:selected:` warn. Also fixes hidden radios intercepting pointer events on the selector, and extends selection sync to keyboard and panel-row selections.
- Hardens llms.txt generation against code blocks with no content lines, which crashed the LLM exporter with a `NullReferenceException`.

The default tabs appearance is untouched; docs-content has no existing `:selected:` usage, so behavior of existing content is unchanged.

Docs: new sections with live demos in `syntax/applies-switch.md`. Tests: 18 new (appearance parsing, guard fallback, default selection, panel markup, label format, LLM export).

## Screenshots

_Closed chip and open panel captures to be attached below._

### DIFF
--- a/src/Elastic.Documentation.Site/Assets/applies-switch.ts
+++ b/src/Elastic.Documentation.Site/Assets/applies-switch.ts
@@ -64,6 +64,9 @@ function ready() {
             }
         }
     })
+
+    // Reflect the (possibly restored) selection in dropdown content panes
+    $$optional('.applies-switch--dropdown').forEach(updateDropdownContents)
 }
 
 /**
@@ -72,19 +75,115 @@ function ready() {
  * @this {HTMLElement} - The element that was clicked.
  */
 function onAppliesSwitchLabelClick(this: HTMLLabelElement) {
-    const data = create_key(this)
+    syncSelection(this)
+}
+
+/**
+ * Activate the same applies_to selection in all other switches on the page
+ * and persist it. `label` is the trigger label of the newly selected option.
+ */
+function syncSelection(label: HTMLElement) {
+    const data = create_key(label)
     if (!data) return
     const [group, id, key] = data
-    for (const label of as_id_to_elements[key]) {
-        if (label === this) {
+    for (const other of as_id_to_elements[key] ?? []) {
+        if (other === label) {
             continue
         }
-        if (label.previousElementSibling instanceof HTMLInputElement) {
-            label.previousElementSibling.checked = true
+        if (other.previousElementSibling instanceof HTMLInputElement) {
+            other.previousElementSibling.checked = true
+            // Setting .checked programmatically fires no change event, so
+            // synced dropdowns need their content panes updated explicitly
+            const dropdown = other.closest('.applies-switch--dropdown')
+            if (dropdown) updateDropdownContents(dropdown)
         }
     }
     window.sessionStorage.setItem(storageKeyPrefix + group, id)
 }
+
+/**
+ * Reflect the checked input of a dropdown switch: show the matching content
+ * pane and hide the matching panel row (the current selection is already
+ * visible in the chip, the panel only lists the alternatives).
+ *
+ * Dropdown switches group their inputs and labels in a selector overlay, so
+ * the pure-CSS sibling selector used by the tabs appearance cannot reach the
+ * content panes.
+ */
+function updateDropdownContents(dropdown: Element) {
+    const checked = dropdown.querySelector('.applies-switch-input:checked')
+    if (!checked) return
+    const index = checked.getAttribute('data-index')
+    dropdown.querySelectorAll('.applies-switch-content').forEach((content) => {
+        content.classList.toggle(
+            'applies-switch-content--active',
+            content.getAttribute('data-index') === index
+        )
+    })
+    dropdown.querySelectorAll('.applies-switch-panel-row').forEach((row) => {
+        row.classList.toggle(
+            'applies-switch-panel-row--current',
+            row.getAttribute('data-index') === index
+        )
+    })
+}
+
+/**
+ * Open/close behavior for switches with the dropdown appearance.
+ *
+ * Delegated document-level listeners registered once at module scope so they
+ * survive htmx swaps (initAppliesSwitch re-runs on every htmx:load).
+ */
+function closeDropdowns(except?: Element | null) {
+    document
+        .querySelectorAll('.applies-switch--dropdown.open')
+        .forEach((dropdown) => {
+            if (dropdown !== except) dropdown.classList.remove('open')
+        })
+}
+
+document.addEventListener('click', (event) => {
+    const target = event.target as HTMLElement
+    // A click on a label also fires a synthetic click on its radio input;
+    // ignore it so it doesn't immediately close the dropdown we just opened.
+    if (target.closest('.applies-switch--dropdown .applies-switch-input')) {
+        return
+    }
+    // Selecting a panel row checks its radio natively (label for=); the
+    // change listener below does the rest, so only close the menu here.
+    if (target.closest('.applies-switch-panel-row')) {
+        closeDropdowns()
+        return
+    }
+    const label = target.closest(
+        '.applies-switch--dropdown .applies-switch-label'
+    )
+    if (!label) {
+        closeDropdowns()
+        return
+    }
+    const dropdown = label.closest('.applies-switch--dropdown')
+    if (!dropdown) return
+    dropdown.classList.toggle('open')
+    closeDropdowns(dropdown)
+})
+
+document.addEventListener('change', (event) => {
+    const input = (event.target as HTMLElement).closest(
+        '.applies-switch--dropdown .applies-switch-input'
+    )
+    const dropdown = input?.closest('.applies-switch--dropdown')
+    if (!input || !dropdown) return
+    updateDropdownContents(dropdown)
+    // Panel-row and keyboard selections bypass the trigger-label click
+    // handler, so propagate the sync from here.
+    const label = input.nextElementSibling
+    if (label instanceof HTMLElement) syncSelection(label)
+})
+
+document.addEventListener('keydown', (event) => {
+    if (event.key === 'Escape') closeDropdowns()
+})
 
 export function initAppliesSwitch() {
     ready()
--- a/src/Elastic.Documentation.Site/Assets/markdown/applies-switch.css
+++ b/src/Elastic.Documentation.Site/Assets/markdown/applies-switch.css
@@ -13,7 +13,7 @@
 			}
 		}
 
-		& > .applies-switch-label > .applies-item {
+		.applies-switch-label > .applies-item {
 			@apply cursor-pointer text-inherit;
 
 			&:hover {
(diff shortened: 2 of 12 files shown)


src/Elastic.Documentation.Site/Assets/markdown/applies-switch.css
```diff
@@ -56,5 +56,142 @@
...
+		   same radios. Content visibility and the panel's current-row hiding
+		   are handled by applies-switch.ts. */
+		&.applies-switch--dropdown {
+			@apply relative block;
+
+			.applies-switch-select {
+				@apply border-grey-30 absolute top-0 right-0 z-10 flex w-max cursor-pointer flex-col items-stretch overflow-hidden border-1 border-b-0 bg-white;
+				border-top-left-radius: 0.5rem;
+				border-top-right-radius: 0.5rem;
+				box-shadow: 0 -2px 6px rgb(0 0 0 / 0.05);
```
reakaleek, line 74: These `0.5rem` corners (and the panel's `rounded-lg` / row `rounded-md` below) read softer than the surrounding markdown chrome. Tabs and applies-switch tabs are square, code blocks use `rounded-t-sm`, and `{dropdown}` uses `rounded-sm` + `shadow-xs`.

Matching that smaller radius family would help this sit next to existing snippets without looking like a different component set.

src/Elastic.Documentation.Site/Assets/markdown/applies-switch.css
```diff
@@ -56,5 +56,142 @@
...
+					transform 0.16s ease;
+			}
+
+			.applies-switch-panel-row {
+				@apply text-ink-light flex cursor-pointer items-center rounded-md px-3 text-sm whitespace-nowrap;
+				height: 1.875rem;
+				transition: background-color 0.12s ease;
+
+				&:hover {
+					@apply bg-grey-10 text-black;
```
reakaleek, line 147: Closed state only uses `opacity: 0` and `pointer-events: none`, so the panel can still sit in the accessibility tree. Screen readers may announce options while the menu looks closed.

`visibility: hidden`, the `hidden` attribute, or `inert` when not `.open` would keep the closed state truly closed for AT as well.

src/Elastic.Documentation.Site/Assets/markdown/applies-switch.css
```diff
@@ -56,5 +56,142 @@
...
+		   all inputs and labels in .applies-switch-select (only the checked
+		   one is visible); the panel rows are extra labels pointing at the
+		   same radios. Content visibility and the panel's current-row hiding
+		   are handled by applies-switch.ts. */
+		&.applies-switch--dropdown {
+			@apply relative block;
+
+			.applies-switch-select {
+				@apply border-grey-30 absolute top-0 right-0 z-10 flex w-max cursor-pointer flex-col items-stretch overflow-hidden border-1 border-b-0 bg-white;
+				border-top-left-radius: 0.5rem;
```
reakaleek, line 72: The chip sits on the trailing edge (`right-0`), while the tabs appearance and `{tabs}` both start their selectors on the leading edge. In LTR layouts that left edge is also where readers scan first ([NN/g on right-justified menus](https://www.nngroup.com/articles/right-justified-navigation-menus/)).

Would anchoring with `left-0` (and flipping `rounded-tr-none` → `rounded-tl-none` on the code corner) keep this aligned with the rest of the switch family? Chevron inside the chip can stay on the right of the label.

src/Elastic.Documentation.Site/Assets/markdown/applies-switch.css
```diff
@@ -56,5 +56,142 @@
...
+			}
+
+			.applies-switch-short {
+				@apply text-sm font-medium whitespace-nowrap;
+			}
+
+			/* Floating options panel, right-aligned 4px below the chip. */
+			.applies-switch-panel {
+				@apply absolute right-0 z-20 w-max rounded-lg bg-white p-1;
+				top: calc(2.25rem + 1px + 4px);
```
reakaleek, line 128: The markdown `{dropdown}` rotates its chevron with no transition (`dropdown.css` only sets `transform` on `[open]`). The animated rotate here feels a bit busier next to that pattern.

Dropping the chevron transition (instant open/close) would match the existing dropdown. If the panel fade stays, a `prefers-reduced-motion` off-switch would be good.

src/Elastic.Documentation.Site/Assets/markdown/applies-switch.css
```diff
@@ -56,5 +56,142 @@
...
+				   the panel only lists the alternatives. */
+				&.applies-switch-panel-row--current {
+					display: none;
+				}
+			}
+
+			/* The chip sits flush on the content's first block: no frame, and
+			   the top padding matches the chip's outer height exactly
+			   (2.25rem label + 1px top border) so the block's top edge runs
+			   at the same line beside and under the chip. */
```
reakaleek, line 160: Hiding the current row keeps the panel short, though it also removes the selected item from the list. NN/g generally prefers keeping options in place (and `VersionDropdown` shows the current version with a check).

With only two items the panel becomes a single alternative, and with one item it can open empty. Showing all rows and marking the current one might be clearer.

src/Elastic.Markdown/Myst/Directives/AppliesSwitch/AppliesSwitchView.cshtml
```diff
@@ -1,5 +1,20 @@
+@using Elastic.Markdown.Myst.Directives.AppliesSwitch
 @inherits RazorSlice<Elastic.Markdown.Myst.Directives.AppliesSwitch.AppliesSwitchViewModel>
-<div class="applies-switch">
+<div class="applies-switch@(Model.IsDropdown ? " applies-switch--dropdown" : "")">
+	@if (Model.IsDropdown)
+	{
+		<div class="applies-switch-select" title="Select another version">
```
reakaleek, line 6: Open/close is driven by a CSS class on the wrapper, and the accessible name is only a `title` on this non-focusable `div`. Assistive tech won't get a clear disclosure control.

Wiring `aria-expanded` / `aria-controls` on the trigger (and toggling them from the click handler) would make the open state explicit. Escape close is a nice touch already.

src/Elastic.Markdown/Myst/Directives/AppliesSwitch/AppliesItemLabelView.cshtml
```diff
@@ -0,0 +1,41 @@
+@using Elastic.Markdown.Myst.Components
+@inherits RazorSlice<Elastic.Markdown.Myst.Directives.AppliesSwitch.AppliesItemViewModel>
+
+@{
+	var id = $"applies-switch-item-{Model.AppliesSwitchIndex}-{Model.Index}";
+}
+
+<input class="applies-switch-input" checked="@(Model.Checked ? "checked" : "")" id="@id" name="applies-switch-set-@Model.AppliesSwitchIndex" type="radio" tabindex="0" data-index="@Model.Index">
```
reakaleek, line 8: When `Checked` is false this can render `checked=""`. For boolean HTML attributes, presence alone can mark the radio checked, so `:selected:` on a later item might leave more than one input looking selected in the DOM.

Worth confirming Razor omits the attribute when false (e.g. pass `null`, or a conditional attribute). The tests assert the `checked="checked"` string, which wouldn't catch an empty `checked=""`.

src/Elastic.Documentation.Site/Assets/applies-switch.ts
```diff
@@ -72,20 +75,116 @@ function ready() {
...
+        if (other.previousElementSibling instanceof HTMLInputElement) {
+            other.previousElementSibling.checked = true
+            // Setting .checked programmatically fires no change event, so
+            // synced dropdowns need their content panes updated explicitly
+            const dropdown = other.closest('.applies-switch--dropdown')
+            if (dropdown) updateDropdownContents(dropdown)
         }
     }
     window.sessionStorage.setItem(storageKeyPrefix + group, id)
 }
```
reakaleek, line 102: Nice catch that programmatic `.checked` skips the change event. Updating dropdown panes from here keeps tabs ↔ dropdown sync working without fighting the CSS sibling selector used by the tabs appearance.

reakaleek (commented):
Thanks for this. The problem statement is clear, and the guardrails (code-block-first fallback, unknown appearance → tabs, sync across appearances, LLM empty-fence fix) make the feature feel safe to try.

Most of the inline notes are about aligning the dropdown skin with existing markdown chrome, plus a couple of accessibility/validity checks on the custom menu. None of that questions the overall approach.

Happy to chat through the left-vs-right chip placement if useful. Left would match tabs and LTR scanning habits, while keeping the chevron on the right of the label text.

Mpdreamz:
Nice, 

Two concerns:

1. Make it its own directive

`````
::::{code-switch}

:::{applies-item} stack: ga 9.0+
```code
```
:::

:::{applies-item} stack: experimental 9.1+
```code
```
:::

:::{applies-item} serverless: ga
```code
```
:::

::::
`````

Instead of having an `:option:` dictate what elements are allowed. 


2. How will this interact with multiple languages?

image

We could extend this code-switch to also have a per language toggle:

`````
::::{code-switch}

:::{language-item} 
```js
```
:::
:::{language-item} 
```pyhon
```
:::
::::
`````

And even combine the two so you can have per version language toggles.

`````
:::::{code-switch}

::::{applies-item} stack: ga 9.0+
:::{language-item} 
```js
```
:::
:::{language-item} 
```python
```
:::
::::

//more
:::::
`````

Mpdreamz (changes requested):
Thanks, this is a useful exploration, and the tests/docs make the intended workflow very clear.

I would not merge this as an `:appearance:` mode on `applies-switch` yet. The dropdown variant changes the content contract enough that it feels like a different authoring primitive: every item must start with a code block, the UI is anchored to that block, and the motivating use case is really versioned code examples with matching callouts. Before this API hardens, I think this should move toward a dedicated `code-switch` style directive, with `applies-item` as one supported dimension. That also gives us room to support language switching cleanly later, instead of overloading `applies-switch` further.

Implementation note for when the draft comes back around: please harden the `:selected:` tests to assert that unchecked radios render no `checked` attribute at all. The current tests only reject `checked="checked"`; if RazorSlice ever emits `checked=""`, browsers still treat that as checked because `checked` is a boolean HTML attribute. That would break the new default-selection behavior while the string assertion still passes.

The branch is also currently conflicting with `main`, with checks from July 28, 2026, so it will need a rebase and fresh CI before another mergeability pass.

PR #3749, opened 2026-07-31
CI: add warn-only accessibility synthetics

reakaleek:
## What

- Add a CI-only synthetics accessibility check that runs axe against key docs pages
- Report WCAG 2.1 A/AA violations as GitHub Actions warnings without failing the job

## Why

- Start measuring docs accessibility in this repo's CI without blocking merges or deployment quality gates
- Keep the existing synthetics suite unchanged for pipelines that run before production deploys

## Notes

- The new check lives outside `synthetics/journeys` and uses a separate npm script so `synthetics:test` and `synthetics:push:*` stay unchanged
- Violations are warn-only for now so follow-up PRs can fix issues before making the check blocking

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,10 @@ jobs:
         working-directory: src/Elastic.Documentation.Site
         run: npm run synthetics:test
 
+      - name: Run accessibility synthetics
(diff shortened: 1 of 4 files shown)


Mpdreamz (approved):
LOVE this long overdue!

PR #3755, opened 2026-07-31
Adds a new product id in preparation of the product release

yetanothertw:
Fixes [#1534](https://github.com/elastic/docs-content-internal/issues/1534)

## Description

Docs need a canonical product ID for the new Elasticsearch Vector Database Serverless project type, for {{product.serverless-vector-database}} substitutions, page front-matter products, and search filters.

Details:

```
ID: serverless-vector-database
Display: Elasticsearch Vector Database Serverless
Versioning: serverless (same system as other Serverless projects)
```

## Changes

Added `serverless-vector-database` to `config/products.yml` in docs-builder.

## For reviewers 

Please let me know if there's anything else I need to add to this PR.

### DIFF
--- a/config/products.yml
+++ b/config/products.yml
@@ -247,6 +247,9 @@ products:
   serverless-security:
     display: 'Elastic Security Serverless'
     versioning: 'all'
+  serverless-vector-database:
+    display: 'Elasticsearch Vector Database Serverless'
+    versioning: 'serverless'
   transport-go:
     display: 'Elastic Transport Go'
     repository: 'elastic-transport-go'


config/products.yml
```diff
@@ -247,6 +247,9 @@ products:
   serverless-security:
     display: 'Elastic Security Serverless'
     versioning: 'all'
+  serverless-vector-database:
+    display: 'Elasticsearch Vector Database Serverless'
+    versioning: 'serverless'
```
reakaleek, line 252:
```suggestion
    versioning: 'all'
```

yetanothertw: I thought it needs to be `serverless` because it's a versionless type of project (this type will only exist in Serverless, unlike the Elasticsearch, Observability, and Security which are also available in other deployment types). 

Maybe I just don't understand how this works, in which case, apologies for questioning/double-checking. Are we sure it should be `all`?

reakaleek: Hmm.. good question.

I didn't remember having a versioning system specifically for serverless.

Let me doublecheck.

reakaleek: It looks like using `serverless` is the right thing to do.

Just FYI, `serverless` is just an alias to `all`

https://github.com/elastic/docs-builder/blob/main/config/versions.yml#L36

reakaleek (approved):

PR #3756, opened 2026-08-03
Fix WaitForDrain race when reader task has not scheduled yet

cotti:
## Why

`ServiceInvoker` starts the diagnostics reader fire-and-forget (`_ = collector.StartAsync(...)`). A command that completes almost instantly can reach `WaitForDrain` before the thread pool has picked up the `Task.Run` delegate, so `IsStarted` is still `false` and `WaitForDrain` throws `InvalidOperationException: WaitForDrain called on a collector that was never started`. This failed elastic-otel-java's `changelog-upload` job during their 1.12.0 release ([run](https://github.com/elastic/elastic-otel-java/actions/runs/30614459135)) — the upload found zero changelog files and returned fast enough to lose the race.

## What

`WaitForDrain` now distinguishes two cases that previously both threw:

- `StartAsync` was never called → still throws immediately (the #3348 guard stays intact; there is genuinely no reader coming).
- `StartAsync` was called but the reader delegate hasn't been scheduled yet → wait for the reader to come up (bounded at 2s, same deadline the drain loop below already uses) and then drain normally.

A new `IsStartRequested` property on the collector exposes the "start requested but not yet running" state. Includes a regression test simulating the fire-and-forget start followed by an immediate `WaitForDrain`.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -44,6 +44,7 @@
     <PackageVersion Include="Microsoft.Extensions.Configuration.UserSecrets" Version="10.0.8" />
     <PackageVersion Include="Microsoft.Extensions.Configuration.Abstractions" Version="10.0.8" />
     <PackageVersion Include="Microsoft.Extensions.Telemetry.Abstractions" Version="10.6.0" />
+    <PackageVersion Include="Microsoft.Extensions.TimeProvider.Testing" Version="10.6.0" />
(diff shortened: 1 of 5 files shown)


Mpdreamz (changes requested):
Thanks for the fix, we should leverage Time provider

https://learn.microsoft.com/en-us/dotnet/standard/datetime/timeprovider-overview

So we can test this in real execution time not physical wall clock time

Mpdreamz (approved):

PR #3764, opened 2026-08-04
Add the backfill census: changelog backfill inventory

cotti:
## Why

Backfill planning needs a full census before any selective run: `config/products.yml` cannot identify release-note surfaces on its own because the `release-notes` feature defaults to enabled, and only products with published history or native artifacts should produce backfill actions — while everything else must stay visible so an unresolved scope can never silently produce empty bundles.

Part of elastic/docs-eng-team#673 (sub-issue of the backfill epic elastic/docs-eng-team#656). **Stacked on #3681** — the census writes the v1 inventory contract that PR introduces.

## What

A new `docs-builder changelog backfill inventory` command (read-only apart from the local output file) backed by `InventoryCensusService`:

- Enumerates every release-notes product from `products.yml` and merges a hand-maintained **census seed** (YAML) mapping products to their release-note sources — repository, ref, paths, target scheme, cutoff, substitutions, attributed repositories, adoption state, and classification.
- Products the seed does not cover become `source-unresolved` entries with a warning; products can instead be deliberately deferred under `unmapped`, each with an auditable reason. `source-unresolved` is not seedable — it is the census's own conclusion, never an operator claim.
- Applies the epic's defaults: stack-versioned semver sources get the `9.0.0` cutoff when none is given; unresolved products get a target scheme derived from their versioning system (serverless/project → date, ech → monthly), always paired with a note so a guess never reads as confirmed fact.
- Checks attributed repositories against the local `assembler.yml` link allowlist (advisory — planning re-validates against the deployed scrubber allowlist identity, docs-eng-team#671 / #3763).
- Output goes through `BackfillDocuments`: enveloped, schema-versioned, validated, content-addressed. Same inputs produce the same hash.

One contract adjustment to the parent PR's `InventoryDocument`: `SourceRepository`/`GitRef` are now null-able **only** for `source-unresolved` entries — the census records products whose location is unknown, which the previously required fields could not express. Validation still demands both for every other classification.

Verified end-to-end against the real repo config: 87 release-notes products enumerated; with a sample seed, the elasticsearch source got the stack default cutoff and `elastic/cloud` correctly reported off-allowlist.

### DIFF
--- a/docs/cli-schema.json
+++ b/docs/cli-schema.json
@@ -4433,7 +4433,96 @@
           ]
         }
       ],
-      "namespaces": []
+      "namespaces": [
+        {
+          "segment": "backfill",
+          "summary": "Backfill historical release-note bundles (docs-eng-team#656): census, planning, and guarded publication.",
+          "options": [],
+          "commands": [
+            {
+              "path": [
+                "changelog",
+                "backfill"
+              ],
+              "name": "inventory",
+              "summary": "Build the backfill census: an inventory document covering every release-notes product.",
+              "notes": "Enumerates every product in products.yml that participates in release notes (the\nrelease-notes feature defaults to enabled, so products.yml alone cannot say which\nproducts have release-note surfaces), merges in the hand-maintained census seed, and writes the\nversioned inventory document that backfill planning consumes. Products the seed does not cover\nstay visible as source-unresolved entries with a warning \u2014 an unresolved scope must never\nsilently produce empty bundles. Attributed repositories are checked against the link allowlist in\nthe local assembler.yml; planning re-validates against the deployed scrubber allowlist\nbefore any upload. This command only reads configuration and writes a local file: no S3 access,\nno writes to any remote system.",
+              "usage": "docs-builder changelog backfill inventory [options]",
+              "examples": [],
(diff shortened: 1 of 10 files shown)


reakaleek (approved):

Mpdreamz (changes requested):
## Request changes: census/inventory is the wrong shape for a one-off

Same pushback as on #3681. Stacking a seed→enveloped-inventory compiler on the six-family contracts still does not move bundles to S3, and it does not discover what is published on elastic.co/docs/release-notes.

For a temporary `changelog migrate-from-web` (fetch published notes → existing bundle YAML → create-only S3 upload → print report → delete the command), we do not need an inventory artifact family, census service, or content-addressed output. A product/cutoff list as plain config (or CLI flags) is enough.

Paper trail, if any: one run report on the issue/PR — not a persisted inventory schema. Please retarget with #3681 toward that one-off path instead of growing the backfill document platform.

cotti:
Closing per the re-scope agreed on #3681: scope and cutoff selection become CLI flags plus a small checked-in config list inside `migrate-from-web` (elastic/docs-eng-team#736) — no inventory artifact family or census service.

PR #3252, opened 2026-05-05
Update ECS version per 9.4 release

taylor-swanson:
In preparation for the ECS 9.4.0 release, bumping the versions for docs-builder.

Do not merge automatically, will merge on release day.

(referenced previous release PR here: https://github.com/elastic/docs-builder/pull/2564)

PR #3766, opened 2026-08-04
Update ECS version per 9.5 release

taylor-swanson:
In preparation for the ECS 9.5.0 release, bumping the versions for docs-builder.

(referenced previous release PR here: https://github.com/elastic/docs-builder/pull/3252)

### DIFF
--- a/config/assembler.yml
+++ b/config/assembler.yml
@@ -95,7 +95,7 @@ references:
   apm-aws-lambda:
   apm-k8s-attacher:
   ecs:
-    current: 9.4 # releases do not always align with the Stack version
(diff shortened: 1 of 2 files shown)


Mpdreamz (approved):
LGTM! I think we have automation for this and may have already been updated on master. Merging it in for good measure though :)

taylor-swanson:
> LGTM! I think we have automation for this and may have already been updated on master. Merging it in for good measure though :)

Oh interesting! Yeah if there's nothing we need to do on our end, I can update our release process accordingly. Only issue I really see is that ECS may or may not release at the same time as the stack (we do our best to release same day, but sometimes it can be delayed).

PR #3772, opened 2026-08-05
Codex: prefer environment-matching docset over public docs/

### LINKED ISSUE #3767: Codex docset discovery should prefer a `registry: internal` docset over a repo's public `docs/`
## Problem

A repository that publishes **both** public docs and internal Codex docs cannot be onboarded to Elastic Internal Docs today, because Codex discovery always resolves the public docset.

`CodexCloneService.FindDocsetFile` searches a fixed order and returns the first hit:

```csharp
private static readonly string[] DocsetSearchPaths = ["docs/docset.yml", "docs/_docset.yml", "docset.yml", "_docset.yml"];
```

falling back to a first-match recursive walk. `docs/docset.yml` therefore always wins.

## Concrete case

`elastic/kibana` publishes its public documentation from `docs/` (`registry` defaults to `public`, assembled into elastic.co/docs). We want internal developer documentation — component library docs, engineering guides — in `docs-dev/` with `registry: internal`, published **only** to codex.elastic.dev.

- **PR previews work.** The repo controls its own workflow, so it can run `docs-builder --path docs-dev`.
- **Live publish does not.** The build in `elastic/codex-environments` clones the repo and discovers the docset itself, landing on `docs/docset.yml` — the public set — and would publish the wrong content to `codex.elastic.dev/r/kibana`.

There is currently no way to steer this from the consuming repo:

- `environments/internal/config.yml` carries only `environment`, `site_prefix`, `title` and `groups`; documentation sets come from the link index, and `CodexDocumentationSetReference.Path` is populated from where discovery found the file rather than from config.
- `codex/build` in `elastic/docs-actions` invokes `docs-builder` with no path argument and exposes no path input.
- No environment variable overrides the documentation path.

## Proposal

During Codex clone/discovery, prefer a docset that declares `registry: internal`, keeping the existing order as the fallback when none does. `DocumentationSetFile.LoadMetadata` is already called on the discovered file, so the registry value is cheap to read.

Adding `docs-dev/docset.yml` to the head of `DocsetSearchPaths` would cover our case specifically, but the registry check seems more principled — it makes intent explicit rather than relying on a folder name, and it does the right thing for any repo that colocates public and internal docsets.

An explicit per-repository `path` override (in the environment config or the link index) would also solve it, and would additionally let a repo publish internal docs from a nested location.

## Impact

Without this, repos with public docs have to move internal documentation into a separate repository, losing colocation with the code it documents.

reakaleek:
## What

- Codex discovery now prefers the `docset.yml` whose `registry` matches the codex environment, instead of always resolving `docs/docset.yml` first.
- The build step reuses the docset that discovery already selected, instead of rediscovering one from the repository root.

## Why

- A repository that publishes both public docs (`docs/`) and internal Codex docs could not be onboarded to Elastic Internal Docs: discovery always landed on the public docset, so live publish would ship the wrong content.
- This lets a repository keep internal documentation next to its code, colocated with a public docset, instead of moving it to a separate repository.
- Fixes https://github.com/elastic/docs-builder/issues/3767

## Notes

- Matching is based on `registry` vs. the codex `environment` string (not a hardcoded `internal` check), per the discussion on the issue, so it also works for other codex environments.
- Falls back to the historical search order (known paths, then a recursive walk) when no docset declares a matching registry, so single-docset repos are unaffected.
- The recursive fallback now runs more often (whenever a known-path docset exists but doesn't match), so the excluded-directory list was widened beyond `.git`/`node_modules` to keep it cheap on large repos.

### DIFF
--- a/docs/documentation/codex/configure/index.md
+++ b/docs/documentation/codex/configure/index.md
@@ -67,3 +67,15 @@ codex:
 toc:
   - file: index.md
 ```
+
+### Repositories with multiple docsets
+
+A repository can publish both a public docset (for elastic.co/docs) and a separate codex docset,
+colocated in the same repository. For example, `docs/docset.yml` with no `registry` (public) and
+`docs-dev/docset.yml` with `registry: my-environment`.
+
+When discovering a repository's docset, codex prefers the docset whose `registry` matches the
+codex `environment`, and only falls back to the default search order (`docs/docset.yml`,
+`docs/_docset.yml`, `docset.yml`, `_docset.yml`, then a recursive search) when no docset declares
+that registry. This lets a repository keep its internal documentation next to its code instead of
+moving it to a separate repository.
--- a/src/Elastic.Codex/Building/CodexBuildService.cs
+++ b/src/Elastic.Codex/Building/CodexBuildService.cs
@@ -184,6 +184,9 @@ public class CodexBuildService(
 			// Repository clone root must be BuildContext `source`: FindGitRoot(..., ceiling: rootFolder) only
 			// discovers .git inside that ceiling (#3115). Using DocsDirectory alone would cap the ceiling
 			// at the docs subtree and return null above repo/.git, breaking GithubEditUrl generation.
+			// The docset file itself is passed explicitly so build reuses the docset clone discovery already
+			// selected (which may prefer a non-default path such as `docs-dev/`), rather than rediscovering
+			// one from the repository root and always landing on `docs/`.
 			var buildContext = new BuildContext(
 				context.Collector,
 				fileSystem,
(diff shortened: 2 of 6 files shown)


Mpdreamz (approved):
LGTM, should we also hard error or warn if codex picks up a docset with no registry defined?

reakaleek:
> LGTM, should we also hard error or warn if codex picks up a docset with no registry defined?

https://github.com/elastic/docs-builder/pull/3772/commits/4ac9655c6087ff283f05512bada56c3c1b72dc05

reakaleek:
@copilot in this PR it's only warning.. But we should create a follow-up to error in the future. create an issue for it in elastic/docs-builder.

clintandrewhall:
Epic!  Thanks for moving so fast on this, @reakaleek !!

PR #3783, opened 2026-08-06
Changelog bundle: match --files filters against the CDN entry pool

cotti:
## Why

Private repos that upload their changelog entries to S3 without keeping local copies (e.g. `cloud`) have PR/issue references scrubbed from the public copies, so PR/issue-based bundle filters can never match. The documented workaround, the `--files` / path-list filter (#3664), forced local entry sourcing — it could not select entries that exist only in S3, so release coordinators had to download files by hand. This is the blocking case in elastic/docs-eng-team#734.

## What

The `--files` / path-list filter now follows the same entry-sourcing gate as every other filter instead of unconditionally forcing local sourcing. When entries are sourced from the CDN (the authoring repo resolves and a CDN base is configured), the requested paths are matched to pool entries by file name — file names survive scrubbing — and do not need to exist locally. A requested name missing from the pool fails the run rather than silently shipping an incomplete bundle. Local reads remain available via `--force-local`, `--directory`, or `bundle.use_local_changelogs`, and `--plan` mirrors the new gate (path-list bundles now report `needs_network: true`).

This makes the profile-mode invocation from the issue work against S3-only entries:

```sh
docs-builder changelog bundle ech-release 2027-01-01 ./list.txt
```

Note this is a behavior change for `--files` runs in repos where CDN sourcing is active: previously the listed files were always read from disk; now they select CDN pool entries by name unless local sourcing is forced. Docs and the CLI schema are updated accordingly.


lcawl (approved):
I added a file list example and the docs now LGTM.

I also tested the following type of commands on both a private and public repo (with and without AWS environment variables set up) and it all worked:

```sh
/path/to/GitHub/docs-builder/.artifacts/publish/docs-builder/release/docs-builder changelog bundle serverless-release 2027-02-02 docs/temp/files.txt
```

Mpdreamz (approved):

PR #3789, opened 2026-08-06
Fix site name and git info when --path points to docs subfolder

Mpdreamz:
## Why

Running `docs-builder serve --path ~/project/docs/` showed **unknown-docs** as the site title and an empty branch/commit panel instead of the repository name and branch.

Root cause (original bug): `Paths.FindGitRoot(DocumentationSourceDirectory, ceiling: rootFolder)`. When `--path` points to the docs subfolder, `rootFolder == docsPath` so the ceiling locked the git-root walk inside that folder and could never see `.git` one level up — `CheckoutDirectory` was null, git info was Unavailable, and the name fell back to `unknown-docs`.

Root cause (design): `FileSystemFactory` exposed ambient statics (`RealRead`, `RealWrite`, `AppData`) that any code could reach for without declaring the dependency in its signature. The documentation scope depends on `--path` and `--output`, which are per-invocation — a singleton can never supply it correctly. Every command that called `FileSystemFactory.RealRead` silently coupled itself to the process CWD, making the scope invisible and untestable.

The statics also made the read/write distinction unenforceable. `BuildContext.ReadFileSystem` and `BuildContext.WriteFileSystem` were both typed `ScopedFileSystem` — a type that says "something scoped me" and nothing more. Nothing in the type system prevented passing a read scope as a write scope, and several test sites did exactly that. Git checkout resolution was scattered across three divergent implementations that could disagree on where the checkout was.

### Strongly-typed, intent-carrying filesystems

The core idea: **a type that accepts `ScopedFileSystem` accepts any scope**. A caller looking at a `ScopedFileSystem` parameter has no idea whether it expects a read scope, a write scope, or whether `.git` access is allowed. The name in the variable — `readFs`, `writeFs` — is the only signal, and names aren't checked by the compiler.

Named subclasses fix this. A parameter typed `DocumentationWriteFileSystem` rejects a read scope at compile time. A parameter typed `CheckoutsFileSystem` can't be satisfied by a `DocumentationFileSystem` for a single docs set. The type carries the intent the bare `ScopedFileSystem` could not.

**Call-site types** — what consumers construct and pass around:

| Class | Where | Use case |
|---|---|---|
| `DocumentationFileSystem` | `src/Elastic.Documentation.Tooling/FileSystems/` | **The entry point.** Constructed via `DocumentationFileSystem.Resolve(invocation, options)`. Runs the six-step bootstrap (invocation → docset anchor → checkout → git dirs → git info → output), exposes the result as `.Paths`, and is the read scope. Callers use it as `fs.Read` and `fs.Write` — `.Read` is the instance itself; `.Write` is a `DocumentationWriteFileSystem` produced during construction, never separately instantiated. |
| `DocumentationWriteFileSystem` | `src/Elastic.Documentation/FileSystems/` | **An artifact of constructing `DocumentationFileSystem`**, accessed via `.Write`. Same checkout roots as `.Read` but `.git` is excluded from the allowlist — writing into `.git` is structurally impossible. Because both scopes come from the same `ResolvedDocumentationPaths`, read and write can never disagree about the checkout or output path. |
| `CheckoutsFileSystem` | `src/Elastic.Documentation.Tooling/FileSystems/` | Scope over a directory of cloned repositories. Used by the **assembler**, which operates on a tree of clones rather than a single documentation set. No docset anchoring, no git resolution. Exposes `.Read` (itself) and `.Write` following the same pattern. |
| `ApplicationDataFileSystem` | `src/Elastic.Documentation.Tooling/FileSystems/` | Scope over the OS application-data directory. Used for the **codex link-index clone**. No per-invocation inputs. |

**Bootstrap helpers** — instantiated internally by `DocumentationPathsResolver`, not by callers:

| Class | Where | Role |
|---|---|---|
| `DocsetScanFileSystem` | `src/Elastic.Documentation.Tooling/FileSystems/` | Rooted at the invocation path, nothing above it. Locates `docset.yml`/`_docset.yml` in steps 1–2 without escaping the invocation root. Discarded once the anchor is found. |
| `GitResolveFileSystem` | `src/Elastic.Documentation.Tooling/FileSystems/` | Rooted `maxParents` above the **docset anchor** (not the invocation). Walks up to `.git` and reads `config`/`HEAD` from resolved git directories, including worktree `commondir` targets outside the anchor's ancestry, in steps 3–5. Discarded once git info is resolved. |

### Statics removed

`FileSystemFactory.RealRead`, `RealWrite`, and `AppData` are gone. Statics made the filesystem dependency invisible in signatures — a type that reached for `FileSystemFactory.RealRead` declared no dependency and could not be tested with a mock. Because the documentation scope is per-invocation (depends on `--path` and `--output`), a singleton was always wrong: it hardcoded the process CWD as the checkout root, which is exactly the bug that caused `--path repo/docs` to show `unknown-docs`.

With the statics gone, every consumer that wants a filesystem must declare it. The dependency is visible, the scope is per-invocation, and the type tells the reader what the scope is for.

### Missing `.git` is now a hard error with a remedy

Previously, failing to find `.git` silently produced a `null` `CheckoutDirectory`, which then caused the name fallback to `unknown-docs` and empty git info downstream — the symptom was observable only at render time with no pointer to the cause.

Now `DocumentationPathsResolver` throws `DocumentationPathException` immediately:

```
No .git found at '/path/to/docs' or within 1 parent directory(ies).
Pass --git-dir to point at the repository's .git directory explicitly.
```

The message names the flag that fixes it. `--git-dir` accepts the path to the `.git` directory; its parent becomes the checkout. Worktrees are handled automatically through `commondir` — `--git-dir` should point at the worktree's own `.git` file's location, not the shared object store.

### Ordered six-step bootstrap (`DocumentationPathsResolver`)

Previously scattered across `BuildContext`'s constructor, `FileSystemFactory`, and three separate git-resolution paths. Now a single static class with one method per step, each creating its own minimal scope and discarding it:

1. **Invocation path** — the `--path` argument or CWD.
2. **Docset anchor** — scan from invocation for `docset.yml`/`_docset.yml` (skipped when `ConfigurationFile` is pre-supplied). Uses `DocsetScanFileSystem`.
3. **Checkout** — `--git-dir?.Parent` ?? `FindGitRoot(anchor, maxParents)` ?? hard error naming `--git-dir`. Uses `GitResolveFileSystem`.
4. **Git directories** — the real `.git` path(s) needed in scope. For worktrees: `checkout/.git` (pointer file) + the `commondir` target resolved via the unscoped inner filesystem.
5. **Git info** — `GitCheckoutInformationFactory.Create` through a scope widened by step 4. Or `DocumentationScopeOptions.Git` if the caller pre-computed it (Assembler, Codex).
6. **Output** — `--output` argument or `checkout/.artifacts/docs/html` (anchored at the checkout, never the invocation, so `--path /repo` and `--path /repo/docs` both write to `/repo/.artifacts`).

### `BuildContext` simplification

Stores one `DocumentationFileSystem` and computes `ReadFileSystem`, `WriteFileSystem`, `DocumentationSourceDirectory`, `DocumentationCheckoutDirectory`, `ConfigurationPath`, `OutputDirectory`, and `Git` as projections of `.Paths`. Nothing is copied independently, so nothing can drift out of agreement. The legacy `ScopedFileSystem` pair constructor is deleted; all seven `src/` callers are migrated to `DocumentationFileSystem.Resolve`.

### Worktree resolution fix

`ResolveGitDirectories` now uses the unscoped inner filesystem for `commondir` traversal, because the main `.git` directory lies outside the gitScope root by design. Explicit `--git-dir` injects its path into `gitDirectories` directly so the factory scope covers it.

## Test plan

- [x] `DocumentationPathsResolverTests` — 21 new tests:
  - Normal repo: `--path /repo` and `--path /repo/docs` → same `CheckoutDirectory` and `SourceDirectory`
  - Regular repo git info (branch, ref, repository name)
  - Worktree: `checkout` = worktree root, `GitDirectories` includes the `.git` pointer path and resolved `commondir` target, git info resolved from main `.git`
  - Worktree + `--path /worktree/docs` → same checkout as `--path /worktree`
  - Explicit `--git-dir`: checkout = gitDir.Parent, git info from overridden dir
  - Mock FS without `.git`: graceful fallback, `checkout = source`, `Git` override preserved
  - Output defaults to `checkout/.artifacts` (regression guard: same for `--path /repo` and `--path /repo/docs`)
  - Explicit `--output` override respected
  - Pre-discovered `ConfigurationFile` skips docset scan
  - No docset → `DocumentationPathException`
  - `DocumentationFileSystem.Resolve` integration for regular repo and worktree
- [x] `GitCheckoutResolutionTests` — regular repo, detached HEAD, worktree absolute/relative gitdir, commondir, missing gitdir, canned-fallback guard
- [x] `FindGitRootTests` — maxParents cases, worktree pointer
- [x] `BuildContextDocumentationCheckoutDirectoryTests` — `--path /repo` and `--path /repo/docs` converge
- [x] `./build.sh unit-test` — 4332 tests pass (0 failures)
- [x] `./build.sh lint` — clean
- [ ] `serve --path ~/Projects/docs-eng-team/docs` — header shows repo name, branch/commit populated
- [ ] `serve --path ~/Projects/docs-eng-team/` — identical to above
- [ ] `serve` from inside the repo — identical
- [ ] Repeat above in a `git worktree` checkout

https://claude.ai/code/session_01PC5GiyZKhreYSqf2grz7pi


reakaleek (approved):

theletterf (approved):

clintandrewhall:
This resolves #3797, which I filed a few hours after you opened this — the `--path` / git-ceiling conflation is the same root cause, and `BuildContextDocumentationCheckoutDirectoryTests` covers the repro directly. One case from that issue's blast radius that I don't think the current default handles, worth a look before merge.

`MaxParents` defaults to `1`:

```csharp
public static IDirectoryInfo? FindGitRoot(IDirectoryInfo startDirectory, int maxParents = 1)
```

and the new tests assert that a docset two levels below the checkout is out of reach:

```csharp
// maxParents=1: checks checkout/docs (depth 0) and checkout (depth 1), neither has .git
result.Should().BeNull("the .git is two levels above the anchor, beyond maxParents");
```

On a real filesystem that `null` becomes a hard error — the mock-FS fallback in `ResolveCheckout` only applies to in-memory scenarios:

```csharp
var innerType = inner is ScopedFileSystem sf ? sf.InnerType : inner.GetType();
if (innerType.Name.Contains("Mock", StringComparison.OrdinalIgnoreCase))
    return source;

throw new DocumentationPathException(
    $"No .git found at '{source.FullName}' or within {options.MaxParents} parent directory(ies). "
    + "Pass --git-dir to point at the repository's .git directory explicitly.");
```

`elastic/infra` is exactly that shape. Its only docset is `docs/resilience-team/docset.yml` — there is no `docs/docset.yml`, no `docs/_docset.yml` and no root `docset.yml` — so the anchor sits two levels below `.git`. It also passes no `path` to `codex-preview.yml`, only `path-pattern: 'docs/resilience-team/**'`, so the invocation root is the repository root. Step 2 still reaches the docset from there, since `TryFindDocsFolderFromRoot` falls through to a recursive walk:

```csharp
configurationPath = rootPath
    .EnumerateFiles("*docset.yml", SearchOption.AllDirectories)
    .FirstOrDefault()
```

which leaves the anchor at depth 2 and `FindGitRoot(anchor, maxParents: 1)` walking `docs/resilience-team`, then `docs`, then stopping.

I have not run this end to end, but if it holds the failure mode on CI is not the message. `IsolatedBuildService` catches it:

```csharp
catch (DocumentationPathException e) when (runningOnCi)
```

and takes the "non-documentation folder" path, which creates the output directory and writes no HTML. Both of `codex/build`'s uploads in `elastic/docs-actions` are `if-no-files-found: error`, so the run fails on artifact upload with nothing pointing at `--git-dir`. Today the same repo builds and renders `unknown-resilience-team`, so this would be a regression from degraded to broken.

A few ways out, in rough order of how much I like them:

- Derive `MaxParents` from the distance between the invocation root and the resolved anchor, so a scan that legitimately descends two levels is not penalised for it. This keeps the depth guard meaningful for `--path` while making it a non-issue for the no-`--path` case.
- Raise the default. Cheap, but it weakens the guard everywhere to accommodate one shape.
- Expose `--git-dir` through `codex/build` and have `infra` set it. Works, but pushes the fix into every consuming repo, which is what this PR is otherwise getting us away from.

Independent of which: it would be worth having the `runningOnCi` catch surface `e.Message` rather than swallowing it, so the `--git-dir` remedy actually reaches whoever is reading the failed run.

The four other repos #3797 lists are all one level deep — `kibana` (`docs-dev/`), `security-team` (`docs-codex/`), `synthetics-service` (`docs-v3/`), `platform-engineering-productivity` (`new-docs/`) — so `infra` is the only one I can find that this would catch.

---

Generated with Claude Opus 5.

Mpdreamz:
Good catch, thanks for the deep dive — pushed a fix in 8fcf6252a.

Went with your first option: `DocumentationPathsResolver.Resolve` now widens `MaxParents` to `Math.Max(options.MaxParents, distance between invocation and resolved anchor)`, applied between steps 2 and 3. For `infra`'s shape (no `--path`, anchor two levels below the invocation root at `docs/resilience-team/docset.yml`), the anchor-to-invocation distance is 2, so the git-root walk now reaches `.git` at the invocation root without touching the default for every other repo. When `--path` points into an unrelated tree the guard is still bounded by that same invocation, so `BoundPreventsEscapingToParentRepo`-style escapes stay blocked (added `AnchorAtInvocationRoot_UnrelatedAncestorGit_StillOutOfReach` to cover it) — the low-level `FindGitRootTests` are untouched since `Paths.FindGitRoot` itself didn't change.

Also bumped the `runningOnCi` catch in `IsolatedBuildService` from `LogInformation` to `LogWarning` so `e.Message` (with the `--git-dir` remedy) doesn't get buried if this ever fires for a real bug rather than a stale merge commit.

Didn't touch `codex/build`/`elastic-actions` — this keeps the fix contained to docs-builder as you'd hoped. New tests: `AnchorTwoLevelsBelowInvocationRoot_StillResolvesCheckoutAtGitRoot` and `AnchorAtInvocationRoot_UnrelatedAncestorGit_StillOutOfReach` in `DocumentationPathsResolverTests`. `./build.sh unit-test` and `./build.sh lint` both pass.

PR #3819, opened 2026-08-10
Assembler: generate API Explorer pages behind staging flag

reakaleek:
## Why

- The assembler build does not generate API Explorer HTML today, so staging cannot serve docs-builder API pages before the bump.sh cutover.
- We need a safe rollout path that keeps production unchanged while we validate generated pages on staging.

## What

- Add an `assembler-api-explorer` feature flag, enabled only for the staging assembler environment.
- Wire `OpenApiGenerator` into the assembler build when the flag is on, fetching multi-version specs from the version index.
- Add unit and integration tests for the flag gate and generated output paths.
- Add a synthetics journey for `/docs/api` that uses the API version switcher, gated on `FEATURE_ASSEMBLER_API_EXPLORER`.

## Notes

- Sitemap generation still excludes `/docs/api/` until the bump.sh proxy cutover (elastic/docs-eng-team#725).
- CI sets `FEATURE_ASSEMBLER_API_EXPLORER=true` so local synthetics can exercise generated API pages.


Mpdreamz (approved):

reakaleek:
@copilot fix https://github.com/elastic/docs-builder/actions/runs/32008498288/job/95322850688?pr=3819

PR #3831, opened 2026-08-11
Add page-aware API version switcher for API Explorer

reakaleek:
## Why

Multi-version API pages need a version switcher that links to the same operation in another major when it exists there. The interim native select only navigated to version landing pages and did not match the docs navigation dropdown pattern.

## What

Replace the interim `<select>` with the same focus-within dropdown used in docs navigation. Build switcher targets from the version index and a cross-version page index so operation, tag, and schema pages deep-link across majors with landing-page fallback. Hide the product `#nav-dropdown` on API pages so only the version switcher appears in the left nav.

Closes elastic/docs-eng-team#723


reakaleek:

reakaleek:

Mpdreamz (changes requested):
This is not ready to approve yet. The current synthetics run fails in the API Explorer journey: `Switch API version with the sidebar dropdown` waits for `#api-version-switcher`, but the element is not found. Since this PR is specifically replacing the version switcher UI, that needs to be fixed and rerun before merge. The branch also needs a rebase because it is currently conflicting.

PR #3846, opened 2026-08-12
Polish Nav V2 sidebar chrome and API method badges

itsalexcm:
## Why

This PR tightens Nav V2 sidebar polish so the panel feels intentional under both isolated serve and the assembler elastic-nav shell: fewer layout gaps, clearer sticky chrome, reliable “you are here” highlighting, and API rows that match the same visual language.

## Design / UX improvements (in this PR)

Use this list in review — everything below is intentional and should be noticeable in `:3000` (isolated) and `:4000` (assembler).

### Layout & sticky chrome
- [x] Sidebar fills the viewport without a dead gap at the bottom (aside padding + `#pages-nav` height flex fix)
- [x] Assembler sticky offset accounts for secondary nav under the static Elastic header (no double-counting the full shell)
- [x] Scroll-up / scroll-down controls sit under Search / Back chrome (`top: 8px`), pinned to the menu host rather than floating oddly in the tree

### Search & Back (island chrome)
- [x] Search sticky block: `16px` padding + light bottom border (`#e3e8f2`) so it reads as chrome, not part of the tree
- [x] Back control matches scroll-button colors and nav-item geometry (`min-height: 32px`, inline padding, `8px` radius)
- [x] Search input height nudged to ~32px (vertical padding `5px`) so it aligns with Back / row rhythm

### Labels, hierarchy & active state
- [x] Nested labels: first label keeps tighter top padding; sibling nested labels get extra top spacing (`24px`) so section breaks breathe
- [x] Open ancestors use muted ink (`#516381`) for text (chevron stays neutral)
- [x] Current / expanded matching works when hrefs contain dotted segments (`/./…`) — e.g. Extend → Kibana deep pages stay highlighted after load and hx-boost
- [x] Shared “sidebar heading” marker so docs `label:` nodes and API tag-group headings render the same non-clickable heading chrome

### API sidebar
- [x] Single-operation rows show an HTTP method badge (GET / POST / …)
- [x] Multi-operation rows show a neutral multi-method (grid) badge instead of a misleading single method

## Explicitly not in this PR

Call these out so reviewers don’t hunt for them:

- [ ] Island “Overview” row for `index.md` landings (overview is still the island landing URL, but not listed as a `.current` tree row)
- [ ] Sidebar performance / lazy TOC work (large assembler nav HTML cost) — deferred on purpose
- [ ] Content / IA YAML quirks (e.g. `label: ───` separators, nested labels under clickable groups in deploy-manage) — those come from `navigation-v2.yml`, not CSS bugs
- [ ] Broader Nav V2 IA or section restructuring — this branch only polishes chrome on top of `findability/alexcerezo`

## Test plan

- [ ] Isolated serve (`FEATURE_NAV_V2=true`, :3000): no bottom gap; Search/Back borders align; scroll buttons under chrome
- [ ] Assembler preview (:4000): same chrome under elastic-nav; sticky offset tracks secondary nav only
- [ ] Deep page under Extend/Kibana: ancestors + current row highlight (including `/./` URLs)
- [ ] Enter an island: Back chrome matches Search; leave/re-enter via hx-boost without losing highlight
- [ ] API docs sidebar: single-method + multi-op badges look correct next to titles

### DIFF
--- a/docs/_docset.yml
+++ b/docs/_docset.yml
@@ -110,6 +110,7 @@ toc:
       - file: applies-switch.md
       - file: automated_settings.md
       - file: buttons.md
+      - file: card-group.md
       - file: changelog.md
       - file: code.md
       - file: comments.md
@@ -120,10 +121,14 @@ toc:
       - file: file_inclusion.md
       - file: footnotes.md
       - file: frontmatter.md
+      - file: hero.md
+      - file: hub-pages.md
       - file: icons.md
+      - file: intro.md
       - file: images.md
       - file: videos.md
       - file: kbd.md
+      - file: link-card.md
       - file: math.md
       - file: diagrams.md
       - file: lists.md
@@ -132,6 +137,7 @@ toc:
       - file: storybook.md
       - file: links.md
       - file: list-sub-pages.md
+      - file: on-this-page.md
       - file: page-card.md
       - file: stepper.md
       - file: substitutions.md
@@ -140,6 +146,7 @@ toc:
       - file: tables.md
       - file: tabs.md
       - file: titles.md
+      - file: whats-new.md
 
   # Documentation builds
   - folder: documentation
--- a/docs/_docset.yml
+++ b/docs/_docset.yml
@@ -254,3 +254,15 @@ toc:
 
   # Development
   - toc: development
+
+  # Hub / product page fixtures used by local tests and assembler preview
+  - folder: testing
+    children:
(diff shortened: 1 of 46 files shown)


itsalexcm:
Closing: Nav V2 polish stays out of `main`. HTTP method badges that matter shipped on the current sidebar in https://github.com/elastic/docs-builder/pull/3973.

Issue #1328, opened 2025-05-30
Restore and redesign “Most Popular” sidebar


KOTungseth:
## Description
When we launched Elastic Docs v3, the "Most Popular" sidebar, which is surfaced through the Marketing team, was removed. This section helped users quickly find high-traffic, high-value content and was particularly useful for new users trying to navigate the docs. The current default sidebar is static and does not provide helpful entry points or guidance.

![Image](https://github.com/user-attachments/assets/6291036b-303b-48c0-b312-ae07116054b1)

We need to restore a version of the "Most Popular" sidebar with a refreshed design and thoughtful content strategy. This should also consider the future integration of onboarding and getting-started content.

## Goals

- Reintroduce the “Most Popular” or “Trending” sidebar to improve discoverability of key content.
- Provide a curated or data-driven list of frequently visited or essential pages.
- Make space for onboarding-related content in the future (e.g., getting started guides, tutorials, “Start here” links).

## Requirements

- Design exploration for the sidebar (placement, behavior, style)
- Analytics-informed or curated list of top docs
- Mechanism to regularly update or auto-populate the list
- Optional: tag or bucket content for new users, such as “New to Elastic?” or “Get Started”

lcawl:
Per discussion with @jmikell821 and @ygenkina today, this will be particularly useful for links to some upcoming online education resources. 

We will have access to summaries for each of the education resources, so per this issue's description it might be feasible to determine good matches between doc pages and education resource based on vector searches. Alternatively, if it must be done manually we'd need to curate that list of docs pages that have a good mapping to a specific education URL.

lcawl:
It might also be useful to consider adding links in that sidebar to [support troubleshooting videos](https://www.youtube.com/playlist?list=PL_mJOmq4zsHbQlfEMEh_30_LuV_hZp-3d) where appropriate. Currently they're linked in some troubleshooting pages manually e.g. https://github.com/elastic/docs-content/pull/2180

cotti:
## AI triage

_Automated investigation against the current codebase. A human still needs to confirm before any code change._

### Original concern

The legacy "Most Popular" siderail (sourced via Marketing) was dropped in the Docs v3 migration; this asks to rebuild a curated/data-driven "Most Popular" or "Trending" section in the article siderail, with room for onboarding/education links.

### Findings

[...]

PR #3877, opened 2026-08-14
Add related learning catalogue

lcawl:
Relates to https://github.com/elastic/docs-builder/issues/1328

## Summary

docs-builder now appends a **Related learning** section on mapped documentation pages. Links come from a global catalog in docs-builder, not from page authors.

### Catalog

- File: `config/related-learning.yml` (embedded like `products.yml`)
- Each entry has a stable ID, title, absolute `https://` URL, and a list of pages
- Pages **must** be qualified cross-links (`docs-content://path.md`). Unqualified paths fail catalog load
- Destinations can be mixed (training, labs, etc.). The first four entries are training modules:
  - `apm-with-elastic` → `docs-content://solutions/observability/apm/index.md`
  - `elastic-agent` → `docs-content://reference/fleet/index.md`
  - `index-basics` → `docs-content://manage-data/data-store/index-basics.md`
  - `data-types-and-mappings` → `docs-content://manage-data/data-store/mapping.md`

Matching is `{current-repository}://{path-relative-to-docset}`, case-sensitive, catalog file order. Missing files do not fail the build; the link simply does not appear.

### How it renders

After parse, if the page matches the catalog, docs-builder appends:

1. A real Markdown H2 (`Related learning`, slug `related-learning-heading`)
2. A list of external links (`target="_blank"` + `rel="noopener noreferrer"`)

The H2 is a normal heading, so **On this page** includes it automatically. If the page has footnotes, the section is inserted before the footnotes group.

### Config loading

The catalog is loaded only on `BuildContext` from `ConfigurationFileProvider`. It is not part of `IConfigurationContext`, DI, or `AssembleContext`. Isolated and assembler builds both pick it up because they construct `BuildContext`.

## Screenshots

When I run `./build.sh publishbinaries` then use that build to serve the `docs-content` docs, I see the new section:

image

## Generative AI disclosure

- [x] Yes  
- [ ] No  
2. If you answered "Yes" to the previous question, please specify the tool(s) and model(s) used (e.g., Google Gemini, OpenAI ChatGPT-4, etc.).

Tool(s) and model(s) used: Cursor Grok 4.6


KOTungseth:
Welcome to the first episode of vibe coding with Lisa and Kaarina.

docs/documentation/catalog/related-learning.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Example
+
+```yml
+links:
+  apm-with-elastic:
+    title: APM with Elastic
+    url: https://www.elastic.co/training/apm-with-elastic
+    pages:
+      - docs-content://solutions/observability/apm/index.md
```
Mpdreamz, line 19: I think rather than listing the pages here the pages should opt-in to listing a learning resource with a directive. 

That way we don't have `pages` here that we don't validate exists and continue to exists. 

Do we need @itsalexcm on this? do we wan't to render these more like cards rather than inline content?

Mpdreamz (changes requested):

docs/documentation/catalog/related-learning.md
lcawl: We didn't drag Alex into this vibecoding session since we figured this was a baby step on a long-standing (and recently reheated) request but not as hot as other priorities on his plate. If you think it merits his review, however, that's great.

As for whether to use a directive instead, that's actually one of the earlier variations of our AI plans. I can put up a PR based on that plan (https://github.com/elastic/docs-builder/pull/3943). My reason for moving away from it was that it required each page owner to be making choices about which resources they wanted to link to. Whereas this method allowed that decision-making to be centralized. If your only concern with this method is absent `pages`, I can replace the current silent miss and vibecode some validation checkes (e.g. reuse the redirects/TOC local-file check)

lcawl:
Unless @KOTungseth disagrees with the strategy (per https://github.com/elastic/docs-builder/pull/3877#discussion_r3868071330), I presume this PR is replaced by https://github.com/elastic/docs-builder/pull/3943

PR #3888, opened 2026-08-18
Add mobile top nav tabs menu

theletterf:
## Why

The section-driven top nav currently relies on a horizontal desktop tab row, which is cramped on narrow screens. A compact mobile presentation keeps configured sections reachable without changing the desktop layout.

## What

This adds a mobile disclosure menu for top-nav link tabs below the desktop breakpoint, keeps the existing horizontal tab row for larger screens, and covers the rendering behavior with navigation tests.

### DIFF
--- a/src/Elastic.Documentation.Site/Assets/secondary-nav.css
+++ b/src/Elastic.Documentation.Site/Assets/secondary-nav.css
@@ -1,0 +1,19 @@
+@layer components {
+	.secondary-nav-mobile-menu summary {
+		list-style: none;
+	}
+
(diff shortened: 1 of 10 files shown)


theletterf:
Latest update: the docs landing page now gets the same mobile pages drawer entry point as regular docs pages.

The landing layout renders a mobile-only hamburger trigger on the left and mounts the pages drawer only on mobile, so the section selector and version selector remain reachable without introducing the desktop sidebar on the landing page.

Verified locally with:

```bash
rtk dotnet test tests/Navigation.Tests/
```

Mpdreamz (approved):

PR #3895, opened 2026-08-19
Add git-diff exporter for CI changed-pages.json

reakaleek:
## Why

- Preview comment jobs guess page URLs from GitHub changed-file paths. Those URLs do not match navigation URLs, path prefixes, or pages that include changed snippets.
- CI needs a builder-produced list of published page URLs and titles.

## What

- Add an opt-in `gitdiff` exporter that writes `changed-pages.json` during isolated builds.
- Map git changes (or GitHub Actions file lists) onto built pages and include reverse-index entries.
- Auto-enable the exporter on CI isolated builds.

## Notes

- Merge the matching `elastic/docs-actions` PR before preview comments use the new artifact.
- Git failures are soft. The build writes an empty `pages` array and logs a warning.

### DIFF
--- a/docs/_docset.yml
+++ b/docs/_docset.yml
@@ -204,6 +204,7 @@ toc:
           - file: llm.md
           - file: okf.md
           - file: plain-text.md
+          - file: git-diff.md
       - folder: release-notes
         children:
           - file: index.md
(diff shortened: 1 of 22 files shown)


Mpdreamz (changes requested):
One blocker before this is ready: `GitChangedFileSource.ResolveDiffBase` returns `origin/<GITHUB_BASE_REF>` immediately without checking that the ref exists. On pull_request CI with a shallow merge checkout, `GITHUB_BASE_REF` is set but `origin/main` is not necessarily present, so `git diff origin/main HEAD` can fail and the exporter will write an empty `changed-pages.json` instead of falling back to `HEAD^1`. The docs and tests say named bases fall back when missing, but that path is skipped for `GITHUB_BASE_REF`. Please verify the GitHub base ref before returning it, or make the diff step fall back when that ref is unusable.

PR #3897, opened 2026-08-20
Search: index generated API versions and product landings

reakaleek:
## Why

- Search currently indexes only one API spec version, so previous majors are missing from results. [docs-eng-team#724](https://github.com/elastic/docs-eng-team/issues/724)
- Queries such as "elasticsearch api" should hit the product overview, not a random operation whose title ends with the product name.

## What

- When `assembler-api-explorer` is on, search indexes every discovered spec version plus one product landing per version, and the sitemap includes generated API HTML URLs.
- When the flag is off, OpenAPI search export is skipped and `/docs/api/` stays out of the sitemap.

## Notes

- The old CloudFront elasticsearch/kibana search fallback is removed. Production will have no API search hits until the flag is on ([docs-eng-team#725](https://github.com/elastic/docs-eng-team/issues/725)).
- Tag, type, and catalog pages stay out of search. They remain in the sitemap when the flag is on.
- Landing pages rank above operations (depth 10/30 vs 20/40) so the product overview wins for product-name queries.

### DIFF
--- a/src/Elastic.Markdown/Exporters/Elasticsearch/ElasticsearchMarkdownExporter.Export.cs
+++ b/src/Elastic.Markdown/Exporters/Elasticsearch/ElasticsearchMarkdownExporter.Export.cs
@@ -100,6 +100,8 @@ public partial class ElasticsearchMarkdownExporter
 		doc.Navigation.Depth = navigationItem?.NavigationDepth ?? 20;
 		if (doc.ContentType == "api" && IsVersionedApiPath(doc.Path))
 			doc.Navigation.Depth = 40;
+		if (doc.ContentType == "api" && IsProductLandingPath(doc.Path))
+			doc.Navigation.Depth = IsVersionedApiPath(doc.Path) ? 30 : 10;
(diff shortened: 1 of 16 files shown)


src/Elastic.Markdown/Exporters/Elasticsearch/ElasticsearchMarkdownExporter.Export.cs
```diff
@@ -119,6 +127,32 @@ internal static void CommonEnrichments(DocumentationDocument doc, INavigationIte
 		// OpenApiDocumentExporter actually sets to "api".)
 		doc.ContentTier = doc.ContentType == "api" ? ContentTiers.Reference : ClassifyContentTier(navigationItem, doc.Path);
 
+		static bool IsVersionedApiPath(string path)
```
Mpdreamz, line 130: this is a little ugly, classification wise. If we need more classifications on docs beyond `ContentType` lets add it do `doc` as an enum at creation time.

Mpdreamz (changes requested):

PR #3873, opened 2026-08-14
feat: dropdown: section flavor for navigation_preview.yml top nav

theletterf:
## Summary

- Adds a `dropdown:` key to the `section:` entry shape in `navigation_preview.yml`
- A dropdown section renders a panel of links in the top nav bar, with no tree membership and no active-tab state
- The rendering template (`_SecondaryNav.cshtml`) and JS (`secondary-nav.ts`) required no changes — `TopNavDropdownItem` was already wired up

Demo: https://docs-v3-preview.elastic.dev/elastic/docs-builder/docs/3873

## What changed

**Config layer** (`SiteNavigationFile.cs`):
- New `SiteDropdownLinkRef(Title, Url)` record
- `SiteSectionRef` gains `DropdownLinks` and `IsDropdown`
- `SiteTableOfContentsCollectionYamlConverter` parses `dropdown:` sequences as `{title, url}` items

**Builder** (`SectionTopNavBuilder.cs`):
- `IsDropdown` sections → `TopNavDropdownItem` with a single flat group; each URL gets the site prefix prepended

**`navigation_preview.yml`**: adds a `Products` dropdown between APIs and Reference, with four solution links as a concrete example

**Test**: `SectionTopNavBuilder_BuildsDropdownTab_WhenDropdownLinksPresent` verifies the new path end-to-end

## Test plan

- [x] `dotnet test tests/Navigation.Tests/` — 225 passed, 0 failed
- [x] `dotnet build` — clean for both `Elastic.Documentation.Configuration` and `Elastic.Documentation.Navigation`
- [ ] Preview build with `NAVIGATION_PREVIEW=true` to confirm the Products dropdown renders in the top nav

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3898, opened 2026-08-20
Restyle the docs sidebar and top bar to match Figma

itsalexcm:
## Why

The nav-preview top bar and left TOC still used the old chrome after #3873 landed the section dropdowns. Readers saw a version selector in the right rail, a Products panel that did not match EUI, and a sidebar that did not match the Figma nav.

## What

Applies the Figma sidebar treatment on the existing V1 TOC (single chevron, island overview as a folder, assembler `navigation_title` overrides) without enabling `NAV_V2`. Moves the version selector into the top bar and restyles Products to the same EUI popover chrome, motion, offset, and trailing-icon size as the version control. Drops the empty gap above View as Markdown now that the selector no longer sits in that rail.

Stacked on #3873 (`feature/nav-section-dropdown`). Preview without feature flags: assembler `get-started` on desktop.

https://github.com/user-attachments/assets/d15bd717-d994-41e2-9251-f3eca04d9e43


Mpdreamz (approved):

Mpdreamz (changes requested):
We need to make sure these changes are gated behind `NAVIGATION_PREVIEW`. Working on amending the PR

Mpdreamz (approved):
Gated the visual changes too behind the feature flag.

PR #3907, opened 2026-08-21
Update changelog bundle-amend command to support CDN changelogs

lcawl:
## Summary

Enhance the `--add` and `--remove` options in the `docs-builder changelog bundle-amend` command so they work with files in S3/CDN not just local files.

## Background

`changelog bundle` already treats the CDN as the source of truth for changelog YAML when `bundle.repo` resolves. `--files` accepts CDN paths (or any path reduced to a basename) and does not require those files on disk. That matches how teams are expected to work: upload entries, bundle from the pool, then delete local YAML (`changelog remove`) because bundles are self-contained.

`changelog bundle-amend` never joined that model. `--add` and `--remove` call `File.Exists` and read disk only. There is no CDN fetch and no `--force-local`. The result is a broken post-publish workflow for the common case:

- **Remove an entry from an existing bundle** when the changelog lives only on the CDN (or was already cleaned up locally). Today you must invent a dummy file whose name matches `file.name`. The dummy’s checksum will not match, so you also need `--force`. That is ceremony, not provenance: `--force` already means “match by name.”
- **Add a changelog that exists only on the CDN** (uploaded, never checked out, or private-repo copy that exists in S3 with scrubbed public PR/issue refs). `--add` cannot embed that YAML unless you download it by hand first. `bundle --files` already solves this for new bundles; amend cannot fix a published bundle the same way.

## Implementation details

`changelog bundle-amend` now uses the same CDN vs local sourcing gate as `changelog bundle --files`. The changelog-tool skill is unchanged, and nothing was committed.

**Behavior**

- CDN by default when `bundle.repo` or the parent bundle’s `repo` resolves.
- `--add` / `--remove` match by **file name** (CDN paths like `/changelog/elastic/kibana/main/247279.yaml` work; no local file required).
- `--force-local` (and `bundle.use_local_changelogs`) still reads disk.
- `--force --remove 300.yaml` excludes inferred git-ref entries without a dummy file.
- `CdnChangelogEntryFetcher.FetchNamedAsync` downloads only the requested names.

**docs-actions:** no `bundle-amend` usage; no changes there.

## Generative AI disclosure

- [x] Yes  
- [ ] No  
2. If you answered "Yes" to the previous question, please specify the tool(s) and model(s) used (e.g., Google Gemini, OpenAI ChatGPT-4, etc.).

Tool(s) and model(s) used: Cursor Grok 4.6


Mpdreamz (approved):
Reviewed manually (AI bot review is blocked today by the daily AI-credits guardrail). Verified: builds clean, `dotnet test tests/Elastic.Changelog.Tests` passes (1043/1043), CDN vs local sourcing gate is shared via the new `ChangelogEntrySourcing` helper, and the add/remove/force/force-local paths are covered by new tests including CDN-miss, checksum-mismatch, and force-local isolation cases.

PR #791, opened 2025-03-20
Refactor to use `WorkingDirectoryRoot` over `Root`.

Mpdreamz:
Replaced references to `Paths.Root` with `Paths.WorkingDirectoryRoot` across the project to improve clarity and alignment with current working directory semantics. Adjusted related logic to handle directory determination and path relations more effectively.

Introduced `BuildContext.DocumentationCheckoutDirectory?` to centrally determine the checkout folder.

PR #3939, opened 2026-08-26
feat(changelog): resolve Lisa's review on #789 — versions, amend reconciler, backport rule

Mpdreamz:
## Summary

Resolves the four points Lisa raised on elastic/docs-eng-team#789.

- **`target` obsoleted on entry/note schema.** `versions` (note-only, plural list) replaces it. Nothing writes `target` any more; readers still accept it for existing pool objects. `BundledProduct.Target` (bundle schema) is unchanged and tracked separately in elastic/docs-eng-team#790.
- **Notes pass through the pool** — always did; updated decision 4 wording to reflect this.
- **`NoteAmendReconciler`**: Lambda auto-generates `.amend-notes.yaml` sidecars for notes that arrive after their release bundle has shipped. One aggregate sidecar per parent bundle, rebuilt from state on every reconcile (idempotent). `bundle_seq` on `NoteIndexEntry` (0/1/2) reports shipped state cheaply.
- **Backport collision rule**: same-leaf note on multiple branches at one version → prefer `main`/`master`, warn on discards, ordinal-first otherwise.

### What changed

| Area | Change |
|---|---|
| Schema | `ProductInfoDto.Target` → `[Obsolete]`; `ProductInfoDto.Versions` added; `NoteIndexEntry` drops `branch`, adds `bundle_seq`; `schema_version: 1` |
| Authoring | `changelog note` writes `versions`; `changelog add` errors if a version is supplied in `--products` |
| Reconciler | `NotesIndexKey` param renamed `target→version`; conditional writes + retry; `NoteAmendReconciler` (new) |
| Bundling | `AmendDocumentBuilder` extracted (pure); `BundleAmendMerger` / `ChangelogBundleAmendService` widen `.amend-notes` as reserved suffix; backport dedup in `MergeNotesAsync` |
| Docs | New `cmd-note.md`; updated `cmd-add.md`, `cmd-bundle-amend.md`, `changelog-fields.md`, `changelog-bundle-registry.md`; CLI schema regenerated |
| Follow-ups | elastic/docs-eng-team#790 #791 #792 #793 #794 filed |

## Test plan

- [x] `dotnet build` — 0 errors
- [x] `dotnet test tests/Elastic.Changelog.Tests/` — all passing (1009 tests)
- [x] `dotnet test tests/Elastic.Documentation.Configuration.Tests/` — all passing (681 tests)
- [x] `./build.sh lint` — clean (pre-push hook passed)
- [x] New `NoteAmendReconcilerTests` (8 tests): late note → sidecar; already shipped → no sidecar, seq=1; human amend → no sidecar; no file annotations → skipped; delete path; idempotent redelivery; no matching bundle; no products in bundle tree
- [x] New `BundleCdnSourcingTests` backport tests (3 tests): main wins, ordinal pick, checksum dedup still works
- [x] Reply comment posted on elastic/docs-eng-team#789; decision list updated

### DIFF
--- a/docs/cli/changelog/cmd-add.md
+++ b/docs/cli/changelog/cmd-add.md
@@ -76,6 +76,10 @@ For example:
 - `"cloud-serverless 2025-08-05"`
 - `"cloud-enterprise 4.0.3, cloud-hosted 2025-10-31"`
 
+:::{note}
+Specifying a version in the `--products` spec (the middle slot, for example `"elasticsearch 9.3.0 ga"`) is an error for `changelog add`. Entries derive their release line from their origin branch, not from a contributor-supplied version. To create an item that explicitly targets one or more versions — such as a known issue or a CVE — use [`changelog note`](/cli/changelog/note.md) instead, which accepts `versions` in place of a target.
+:::
+
 The `changelog add` command resolves product values in the following order:
 
 1. The `--products` CLI option always takes priority.
--- a/docs/cli/changelog/cmd-note.md
+++ b/docs/cli/changelog/cmd-note.md
@@ -1,0 +1,74 @@
+## Description
+
+Create a changelog note file for an item that applies to one or more specific release versions and has no associated pull request.
+Notes are used for known issues, security advisories, and other items that are not tied to a single PR.
+For details and examples, go to [](/data/release-notes/create.md).
+
+Note files are named `note-{slug}.yml` and are uploaded to the changelog pool like any other entry.
+Each note declares `products[].versions` — the release versions it applies to — instead of deriving its release line from a branch.
+
+## Options
+
+: `--products`
+  Products and versions in the format `"product versions lifecycle, ..."` where `versions` is a `|`-separated list of release versions (for example, `"elasticsearch 9.3.0|9.4.0 ga"`).
+  Unlike `changelog add`, the middle slot is interpreted as a `|`-separated version list, not a single target.
+  The valid product identifiers are listed in [products.yml](https://github.com/elastic/docs-builder/blob/main/config/products.yml).
+
+: `--title`
+  A short, user-facing headline for the note (max 80 characters). Required.
+
+: `--type`
+  The type of change. For valid values, see [ChangelogEntryType.cs](https://github.com/elastic/docs-builder/blob/main/src/Elastic.Documentation/ChangelogEntryType.cs). Required.
+
+: `--description`
+  Additional information about the note (max 600 characters). Optional.
+
+: `--issues`
+  URLs of related issues. Optional citation field; does not determine note addressability.
+
+## Product and version format
+
+The `--products` option uses the same positional format as `changelog add`, but the middle slot is a version list:
+
+- `"elasticsearch 9.3.0 ga"` — one version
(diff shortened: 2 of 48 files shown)


docs/cli/changelog/cmd-add.md
```diff
@@ -76,6 +76,10 @@ For example:
 - `"cloud-serverless 2025-08-05"`
 - `"cloud-enterprise 4.0.3, cloud-hosted 2025-10-31"`
 
+:::{note}
+Specifying a version in the `--products` spec (the middle slot, for example `"elasticsearch 9.3.0 ga"`) is an error for `changelog add`. Entries derive their release line from their origin branch, not from a contributor-supplied version. To create an item that explicitly targets one or more versions — such as a known issue or a CVE — use [`changelog note`](/cli/changelog/note.md) instead, which accepts `versions` in place of a target.
```
lcawl, line 80: > Entries derive their release line from their origin branch, not from a contributor-supplied version. 

"release line" doesn't mean anything to me so I can't suggest an alternative, just calling out that it's not obvious what that means.

docs/data/release-notes/_snippets/changelog-fields.md
```diff
@@ -20,12 +20,11 @@ products:
...
       
-      target:
+      versions:
 
-      # An optional string that facilitates pre-release doc previews.
-      # For products with version releases, it contains the target version number (V.R.M).
-      # For products with date releases, it contains the target release date
-      # or the date the PR was merged.
+      # Note files only — a required list of release versions this note applies to.
+      # Entries derive their release line from their origin branch; this field is not valid on entries.
```
lcawl: This makes me think we might need a terminology adjustment. AI always wanted to call what I called "changelogs" some variation of "entries" (likely due to that name used as a grouping within bundles). And now we're adding "notes" on top of that. What terminology works now that we have two variations that share the same changelog schema? Change/release "logs" and "notes"?

```suggestion
      # This field is mandatory only when the changelog is a note (i.e. doesn't have a PR).
```

Maybe rather than trying to slice and dice what IMO seems to be a low-level detail where changelogs are concerned, we should not force users to know/care about the difference? e.g. they can put versions in for any/all types but it's only mandatory for the "changelog note" command and ignored for all else?

docs/cli/changelog/cmd-note.md
```diff
@@ -0,0 +1,74 @@
+## Description
+
+Create a changelog note file for an item that applies to one or more specific release versions and has no associated pull request.
```
lcawl, line 3: > Create a changelog note file for an item that applies to one or more specific release versions and has no associated pull request.

It can be a follow-up PR but the concept of notes will also need to be added to the docs in https://docs-v3-preview.elastic.dev/elastic/docs-builder/tree/main/data/release-notes/overview unless we're going to retire those.

Mpdreamz:
Terminology can do a re-eval for sure 

roughly:


release -> bundle
`entry` -> bundle entry 

`entry` exists of 

1 or more `<pr>.yaml` changelog files.
or 1 `note-<name>.yaml`. 


A release entry is a single entry in the release that will be displayed as a single unit.

PR #3942, opened 2026-08-26
Update CODEOWNERS for documentation ownership

shainaraskas:
Docs is getting pinged on a lot of eng PRs now and they're going faster than we can review. I propose that we remove ownership until we can find a valid reason to keep it. Eng can tag specific writers when relevant (e.g. when a writer submitted an issue that is being addressed)

also: tech leads DNE

edit: I think it's nice for writers to be able to update guidance pages, but I don't think docs eng having to ✅ that smaller number of contributions is a bad thing


bmorelli25 (approved):
SGTM

Mpdreamz (approved):

PR #3951, opened 2026-08-27
Fix stale git lock files and silent fetch failures in clone path

Mpdreamz:
## Why

When a `git fetch` is killed by the per-attempt CI timeout, git leaves `*.lock` files behind in `.git/`. The next retry dies instantly with exit 128 ("File exists") — a retry slot burned on a condition that can never self-recover without clearing the lock first.

Compounding this, `IGitRepository.Fetch` was `void` and swallowed its failure return value, so when all retries were exhausted, `FetchAndCheckout` still proceeded to `git checkout` against a tree with no matching ref. The wipe-and-reclone fallback in `CloneRef` never fired for fetch failures — only for exceptions from `GetCurrentCommit` and `Pull`.

## What

- `GitLocks.ClearStale(IFileSystem, ...)` sweeps `*.lock` files under `.git/` before each retry. Called only from the retry path — never before attempt 1, where a lock could belong to a concurrent process.
- `OnBeforeRetry()` virtual hook added to `ExternalCommandExecutor`; git executor subclasses override it to call `GitLocks.ClearStale`.
- `IGitRepository.Fetch` and `Pull` now return `bool`. `FetchAndCheckout` short-circuits on a failed fetch. `CloneRef` triggers wipe-and-reclone on fetch or pull failure, using the same pattern already in place for `GetCurrentCommit` exceptions.
- The dead `try/catch` around `git.Pull` (which could never fire since `ExecInWithRetry` doesn't throw) is replaced with a bool-check reclone path.

### DIFF
--- a/src/Elastic.Codex/Sourcing/CodexGitRepository.cs
+++ b/src/Elastic.Codex/Sourcing/CodexGitRepository.cs
@@ -31,6 +31,10 @@ public class CodexGitRepository(ILoggerFactory logFactory, IDiagnosticsCollector
 		BaseDelay: TimeSpan.FromSeconds(5),
 		AttemptTimeout: GitTimeouts.CiDefault);
 
+	protected override void OnBeforeRetry() =>
+		GitLocks.ClearStale(WorkingDirectory.FileSystem, WorkingDirectory.FullName,
+			f => Logger.LogWarning("Removed stale git lock file {LockFile}", f));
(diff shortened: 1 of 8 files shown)


Mpdreamz:
@reakaleek @theletterf merging this in since im seeing this a lot in CI now that we do retries more rigorously.

PR #3952, opened 2026-08-27
Move synthetics into assembler-preview job, remove standalone synthetics job

Mpdreamz:
## Why

The `synthetics` job in `ci.yml` built the entire assembled site from scratch on a 2-core `ubuntu-latest` — cloning ~30 repos, running the assembler, serving locally — before running journey tests against it. This took 20+ minutes and was the most frequent source of CI flakiness (git fetch timeouts, shallow lock collisions).

The `assembler-preview` job already builds and deploys the full assembled site to a CloudFront-backed S3 bucket on every PR. Running synthetics against that deployed URL costs nothing extra and gives the same signal without a redundant clone.

## What

- `synthetics` job removed from `ci.yml`.
- Synthetics steps added to `assembler-preview.yml` after the S3 upload, against the known preview URL (`https://docs-v3-preview.elastic.dev/${ASSEMBLER_PREVIEW_PATH_PREFIX}`). No polling, no waiting — the URL is already live at that point in the job.
- Deployment status reflects synthetics outcome: marked success only when both upload and journey tests pass.
- `synthetics.config.ts` gains a `preview` case; `synthetics:test:preview` convenience script added to `package.json` for local use against a preview URL.


.github/workflows/assembler-preview.yml
```diff
@@ -120,6 +120,43 @@ jobs:
...
+        with:
+          cache: npm
+          cache-dependency-path: src/Elastic.Documentation.Site/package-lock.json
+          node-version-file: .nvmrc
+
+      - name: Install frontend dependencies
+        working-directory: src/Elastic.Documentation.Site
+        run: npm ci
+
+      - name: Wait for preview to be live
```
reakaleek, line 133: nit: we could use https://www.npmjs.com/package/wait-on for this. With `npx wait-on ...`

reakaleek (approved):

Mpdreamz:
@copilot please fix the merge conflicts in this pull request.

.github/workflows/assembler-preview.yml
Mpdreamz: I used that at first, the original did too but for some reason it kept timing out.

PR #3955, opened 2026-08-27
Add writing standard and surface map; rework PR/issue skills and templates

Mpdreamz:
The `pr` skill produced bodies a reviewer could navigate but a newcomer could not. The What read as a diff transcript, the Why often opened with the history of a prior PR, there was no front-loaded outcome, no way for the reviewer to check the claim, and no shared vocabulary for what a change touches.

**Affects:** Agentic Skills

## Why

Four gaps, each traceable to a merged PR:

| Gap | Evidence |
|---|---|
| No front-loaded outcome | [#3941](https://github.com/elastic/docs-builder/pull/3941) opens `## Why` with the history of a *rejected* PR before stating the problem this PR solves |
| No way to check the claim | [#3952](https://github.com/elastic/docs-builder/pull/3952) deletes an entire CI job; the body never says the journey tests passed on its own preview |
| Blast radius invisible | [#3746](https://github.com/elastic/docs-builder/pull/3746) writes into a shared CloudFront KVS; never flagged as shared state |
| Known gaps buried or invented | [#3746](https://github.com/elastic/docs-builder/pull/3746) buries a limitation at the end of `## How`; [#3936](https://github.com/elastic/docs-builder/pull/3936) invents a `## Notes` heading for deferred work |

Two additional findings during exploration:

- The `pr` skill listed `fix` and `ci` as first-class labels, but both are release-drafter aliases that split the same changelog category.
- `.github/workflows/required-labels.yml` enforces exactly one release-drafter label at `mode: exactly, count: 1`, derived live from `release-drafter.yml`. The skill now states this so the constraint is understood, not just obeyed.

#### Writing standard and surface vocabulary
Two new shared reference files that all three skills read before writing anything. `writing-style.md` defines plain-language rules with before/after pairs from merged PRs. `surfaces.md` maps diff paths to reader-facing surface names — used for the `**Affects:**` line in PRs and the area label in issues. The map is advisory: a skill coins its own term when nothing fits and proposes it back as an addition.

#### PR, commit, and issue skills
The `pr` skill gains a lead sentence, `**Affects:**` line, and `## Verify` section; `## How` is replaced with conditional `**Breaking**` / `**Out of scope**` / `**Risk**` / `**Stack**` add-ons; the label table is trimmed to the ten canonical release-drafter labels. A new `issue` skill checks for duplicates then creates a bug or feature report with type, area, and `needs triage` labels in one call. The `commit` skill's hardcoded model name is removed.

#### Issue templates
Each form cut to two required fields and one optional. The bug form merges the describe/expected fields and adds a version input — the most useful triage data, previously missing. The enhancement form drops the prerequisite checkboxes and priority dropdown. `config.yml` keeps blank issues enabled and adds a discussion link for questions that are not defects.

**Before/after: [#3941](https://github.com/elastic/docs-builder/pull/3941)**

> ❌ `## Why` opens: *[#3911](https://github.com/elastic/docs-builder/pull/3911) tried to fix stale CDN pool listings by bringing back `changelog/{org}/{repo}/{branch}/registry.json`. That approach was rejected…*

> ✅ `## Why` opens: *Scrubbing strips `prs:` from the public copies of private-repo entries. `--prs` joined only on that YAML field, so those entries dropped out of the bundle with no error.*

**Before/after: [#3856](https://github.com/elastic/docs-builder/pull/3856) — label**

> ❌ `feature`

> ✅ `breaking` — `output:` was removed from `changelog.yml` profiles. `Configuration` in `**Affects:**` plus a field removal is the trigger; the map derives the label correction from the diff rather than relying on after-the-fact judgment.

**Out of scope** — `CLAUDE.md`'s test-project mapping is deliberately not duplicated in `surfaces.md`; it already exists there and in the `/test` skill.

### DIFF
--- a/.claude/skills/writing-style.md
+++ b/.claude/skills/writing-style.md
@@ -75,6 +75,8 @@ Do **not** backtick prose nouns that merely share a name with code — the assem
 
 ## Anti-mechanical rules for "What" sections
 
+`## What` uses `####` subheadings, not bullet points. Each subheading names the thing that changed (a file, a command, a concept). The prose under it states what changed and why it matters — two to four sentences, same plain-language rules as everywhere else.
+
 Lead with the behaviour change; name the symbol second when the reviewer needs it to find the code.
 
 | Wrong | Right |
@@ -82,12 +84,12 @@ Lead with the behaviour change; name the symbol second when the reviewer needs i
 | `` `GitLocks.ClearStale` sweeps `*.lock` files before each retry `` | A retry clears stale `*.lock` files under `.git/` before it runs |
 | `` **`IsolatedBuildService.Build`**: captures the `GenerateAll` result `` | The build result is captured and `redirects.json` is written to the output directory |
 
-**Banned openers** for bullet points:
+**Banned openers** for the prose paragraph:
 - "Added", "Updated", "Changed", "Refactored", "Modified"
-- Any bullet that starts with a file path or a symbol name
-- Any bullet that only restates a filename
+- Any sentence that starts with a file path or a symbol name
+- Any paragraph that only restates the subheading
 
-**Three to five items in a "What".** More than five is a signal the change should be split.
+**Three to five `####` sections in a "What".** More than five is a signal the change should be split.
 
 ---
 
--- a/.claude/skills/writing-style.md
+++ b/.claude/skills/writing-style.md
@@ -1,0 +1,118 @@
+# Writing style for commits, PRs, and issues
+
+Every commit message, PR body, and issue filed in this repo follows these rules.
+Skills that write those artifacts read this file first.
+
+---
+
+## Governing principles (ISO 24495-1)
+
+- **Relevant** — write for a reviewer who has not seen the branch. Cut whatever the diff already states plainly.
+- **Findable** — the first sentence states the outcome. A newcomer reads it and knows whether the change concerns them.
+- **Understandable** — plain words, short sentences, no assumed context.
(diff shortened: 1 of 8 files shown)


.claude/skills/writing-style.md
reakaleek, line 1: Would it make sense to move the writing-style skill to it's own skill and then reference the writing-style skill from the other skills?

reakaleek: I'm not sure how this would work with claude. But I know it's very common in cursor.

theletterf: Might work!

Mpdreamz (commented):
On the `writing-style.md` question from @reakaleek: the current approach — each skill has a 'Read `writing-style.md` before writing anything' instruction — works in Claude Code because the Read tool call loads the file content into the agent's context before it generates text. Making it a proper skill (with its own `SKILL.md`) would allow other skills to invoke it via the Skill tool instead, which is the Cursor pattern you're describing.

The tradeoff: as a reference file it's zero ceremony — the agent reads it and applies the rules inline. As a skill it adds a round-trip that ultimately produces the same context injection, but gains one thing: it becomes invocable standalone, which would be useful if you wanted to ask Claude to review prose against the standard directly.

I'd lean toward keeping it a reference file for now — the 'read this first' pattern is already working and there's no clear use for standalone invocation yet. Happy to restructure if the team prefers the skill pattern.

reakaleek (approved):

PR #3958, opened 2026-08-28
API Explorer: add local_spec for checkout-local OAS files

reakaleek:
## Why

- Generated OpenAPI files often live outside `docs/` and use a different filename than the hosted version-index key.
- `spec:` could not express both sources, so local preview either failed path checks or silently used CloudFront.

## What

- Add optional `local_spec:` for a file anywhere in the same git checkout.
- Keep `spec:` as the hosted index basename.
- Prefer the local file for `main`. If `local_spec:` is set and missing, warn and use the hosted spec.

## Notes

- Omitting `local_spec:` keeps the old implicit file next to `docset.yml`. A missing implicit file stays silent.


reakaleek:

Mpdreamz (commented):
Whats the expectation of `local_file` and assembler builds?

Mpdreamz (approved):
Reviewed the current diff. The explicit local_spec path is bounded to the checkout, missing declared files warn and fall back to the hosted spec, and the implicit spec behavior remains unchanged. The previous CI run was green. This still needs a rebase because the branch is conflicting.

reakaleek:

PR #3967, opened 2026-08-31
Guide agents to documentation discovery resources

theletterf:
Public machine-readable documentation pages now open with a callout that directs agents to the complete documentation index and the Docs MCP server. Internal and Codex output remains unchanged.

**Affects:** Authoring, MCP

## Why

Agents that enter the documentation through one Markdown page cannot see the wider page map. The existing `llms.txt` index and free Docs MCP server solve this problem, but public page output does not advertise either resource. Internal documentation must not point agents toward public discovery services.

#### Public page callout

Every public Markdown page places a documentation resources blockquote before the H1 and all other page content. The callout directs agents to inspect `llms.txt` first and names the MCP tools available for targeted discovery and retrieval.

#### Visibility guard

The exporter emits the callout only when the docset uses the public registry and the build is not Codex. Internal and Codex pages omit it entirely.

#### Environment-aware links and live rendering

The exporter derives the index and MCP URLs from the active build context. Live `.md` responses use the same visibility and placement rules as exported pages.

## Verify

```bash
```

### DIFF
--- a/docs/data/exporters/llm.md
+++ b/docs/data/exporters/llm.md
@@ -43,6 +43,10 @@ applies_to:
 ---
 ```
 
+## Documentation discovery
+
+Each exported page includes a documentation resources note immediately after its title. The note directs agents to the site's `llms.txt` index before they explore other pages, and to the free Elastic Docs MCP server for targeted search, related-page discovery, and page retrieval.
+
 ## Content negotiation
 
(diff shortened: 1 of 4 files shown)


theletterf:
Shelving as @Mpdreamz confirmed that they serve raw file links, which explains the 404s.

PR #3796, opened 2026-08-06
Improve API explorer pages, examples rail, and Nav V2 sidebar

itsalexcm:
## Why

API reference pages were hard to scan: alternate paths hid behind a dropdown, examples vanished when the spec omitted `x-codeSamples`, and the sidebar did not match the Nav V2 / Figma chrome used on the rest of the site.

## What

Integrates ApiExplorer with Nav V2 and adds a right-hand Examples rail (scenarios, languages, status tabs), including synthetic Console/curl samples and empty-status placeholders when the spec has no payloads.

Operation pages get breadcrumbs, labeled HTTP method chips, and a stacked overload list. Multiple endpoints use full-row radios with an 8px gap and a hover copy control on the route. Prerequisites from `x-req-auth` render as label + badge rows. Property types read as `array of strings` without the recursive badge.

Nav V2 sidebar chrome matches Figma (folder structure, spacing, sticky search/back, nested blue active accent). API operations show method glyphs or a `multi` badge. HTMX OOB swaps include the examples panel so it clears when leaving API pages.

PR #3973, opened 2026-08-31
Restyle API Explorer operation pages and type chips

itsalexcm:
API Explorer operation pages now render an examples rail, typed property chips, collapsed multi-item sections, and in-page property anchors. HTTP auth schemes sit in an Authorization heading. Nested objects use the same disclosure as Response.

**Affects:** API reference, Site UI, Isolated builds

**Prompt summary:** Lift the API Explorer page restyle off the Findability-stacked branch onto `main`. Put the version switcher in the sidebar chrome like the back link, using the top-bar `nav-select`, and apply the same control to the examples scenario picker. Paint type groups on the chip, collapse long parameter lists, and keep the isolated fixture key unique so assembler preview does not collide with docs-content. Then link only the type token to the type page, treat Authorization as a real section, collapse any of those lists that have more than one item, and match nested property toggles to the Response `show fields` control.

## Why

Operation pages on `main` have no examples rail and no breadcrumbs. Types and constraints sit in a separate validation block, and every path or query name stays expanded. HTTP schemes render as a chip row that does not match Prerequisites. A single type-page link wraps a whole property row, so names, wrappers, and the hash icon are not independently clickable. Nested objects use green and blue chips that compete with the Response `show fields` control.

#### Examples rail
Every operation page that has samples or named examples renders `#api-examples-panel` beside the article. Highlighting and copy buttons cover the rail, including `console` and `curl` blocks. At 1024px and up, the layout becomes a two-column grid driven by CSS `:has`. When an operation has no `x-codeSamples`, the generator builds a minimal Console and curl sample from the method, path, and document `servers` URL.

#### Type chips and property anchors
Property rows paint six type groups on the chip itself: primitive, value type, named object, linked container, wrapper, and status. Constraints such as `min` and `default` sit next to the type. The property name is an in-page anchor. Only the type token, such as `AggregationContainer`, opens the type page.

#### Collapsed sections
Path, query, Request, Authorization, and Prerequisites start collapsed when they have more than one item, behind a one-line name summary. A single-item section stays a plain heading. Authorization lists `Api key`, `Basic`, and `Bearer` from OpenAPI `security`. Privilege lines from `x-req-auth` stay in Prerequisites.

#### Sidebar chrome
API pages keep `_ApiPagesNav`. HTTP method glyphs come from `IHttpMethodNavigationModel`. The version switcher lives next to the island back link, and the examples rail uses the same `nav-select` when an operation has more than one scenario. HTMX swaps `#api-content-grid` so the sidebar is not remounted between API pages. View as Markdown sits in the left sidebar because the right rail is examples-only.

#### Nested disclosure and Isolated fixtures
Nested objects use `+ show properties` / `− hide properties`, the same control as `+ show fields` on Response. Isolated serve registers Kibana, serverless, and Cloud specs on `docs-builder-*` keys so assembler preview does not collide with docs-content.

## Verify

```bash
dotnet test tests/Navigation.Tests/
cd src/Elastic.Documentation.Site && npm run test
FEATURE_NAVIGATION_PREVIEW=true dotnet run --project src/tooling/docs-builder -- serve --port 3000
```

Open `/api/doc/docs-builder-elasticsearch/operation/operation-search`. Check Authorization, the Request accordion, type-name links, and `collapse` → `inner_hits` for nested `show properties`.

**Out of scope** — Nav V2 and the Findability hub. This PR supersedes [#3796](https://github.com/elastic/docs-builder/pull/3796). The leftover `[with]` compile workarounds in Changelog and Markdown are not in this PR.


reakaleek:

reakaleek (approved):

PR #3989, opened 2026-09-01
Add an additive feature-flag skill for agents

reakaleek:
Agents that add a feature flag now load `/feature-flag` and must prove the flag-off path still matches the previous default.

**Affects:** Agentic Skills

## Why

A new feature flag is easy to wire around new code only. Existing output then lives only on the flag-on path. Environments that leave the flag off lose behavior that already shipped. The additivity check has to happen before the work is done, not after a regression.

#### Additive rule
Flag off keeps the previous default. Flag on adds new behavior only. The skill treats a relocated copy of shipped output as a failure, not as a new feature.

#### Flag registration
The skill points at `FeatureFlags.cs`, `feature_flags` in `assembler.yml`, and `ToFeatureFlags()`. New flags stay off in `prod`. Tests must cover flag-off, flag-on, sibling flags, and the default environment.

#### Agent discovery
`AGENTS.md` lists `/feature-flag` with the other required skills so an agent loads it when it adds or extends a flag.

**Out of scope** This PR does not change how flags load or which environments enable them.

### DIFF
--- a/.claude/skills/feature-flag/SKILL.md
+++ b/.claude/skills/feature-flag/SKILL.md
@@ -1,0 +1,76 @@
+---
+name: feature-flag
+description: >
+  Registers a feature flag as additive-only and verifies the flag-off path
+  still matches the previous default. Use whenever adding or extending a
+  feature flag, editing FeatureFlags.cs, changing feature_flags in
+  assembler.yml, gating a code path behind a FEATURE_* environment variable,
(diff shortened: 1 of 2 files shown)


Mpdreamz (approved):
Worth having. Fwiw navigation_preview was tested on and off by me and and agent so sometimes things can still slip :)

Unless we test all feature flag combinations with synthetics but that is a bit expensive.

PR #3991, opened 2026-09-01
Enable the new navigation on staging

itsalexcm:
Staging shows the new docs top navigation after the next assembler deploy. The version picker stays on the top bar while the flag is on.

**Affects:** Site UI, Configuration

## Why

Staging currently serves the legacy pages dropdown. The team wants the new top nav on staging this week so it can be reviewed before production. The version-dropdown journey already follows both placements, so the flag does not need to stay off to keep that test green.

#### Staging flag
`NAVIGATION_PREVIEW` is on for the `staging:` environment in `assembler.yml`. Preview and `dev` already use this flag. Production does not change.

#### Version picker
Assembler pages still use `ShowVersionDropdown` as the only visibility gate. With the flag on, the picker renders in the top bar. The right rail stays empty for that control, which is the preview layout.

#### Synthetics
The `version-dropdown` journey reads the `navigation-preview` class on `body` and looks in the top bar or the right rail. Turning the flag on does not require a new assertion.

**Out of scope:** Production stays on the legacy nav. This PR does not change picker placement or mobile chrome.

**Risk:** The next staging assembler deploy changes the public staging chrome. The edit only adds a feature flag. It does not change the public-bucket allowlist in `assembler.yml`.

### DIFF
--- a/config/assembler.yml
+++ b/config/assembler.yml
@@ -23,6 +23,7 @@ environments:
     feature_flags:
       WEBSITE_SEARCH: true
       ASSEMBLER_API_EXPLORER: true
+      NAVIGATION_PREVIEW: true
     website_search_url: https://staging-website.elastic.co/search/elastic-website-search.js
   edge:
     uri: https://d34ipnu52o64md.cloudfront.net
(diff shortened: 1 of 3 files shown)


reakaleek (approved):

Mpdreamz:
https://github.com/elastic/docs-builder/pull/3979/changes

Set this

https://github.com/elastic/docs-builder/pull/3988 reverted this. 


This restores it again. 

LGTM!

PR #4007, opened 2026-09-02
Keep the sidebar still after HTMX and animate folder groups

itsalexcm:
The pages sidebar stays put on same-tree HTMX swaps. Folder groups animate open and close, and the top bar uses the same hover on active tabs.

**Affects:** Site UI

## Why

Same-tree HTMX swaps recollapse folders and recentre the current item. The first click on a folder also navigates, so `hx-preserve` moves `#pages-nav` and cancels the height transition. Active top-bar tabs keep the blue text after a click because `:focus-within` reuses the hover colors.

#### Same-tree sidebar

A same-tree swap keeps the live `#pages-nav` in place. Folders that the reader already opened stay open. The current item does not jump to the middle of the sidebar.

#### Folder accordion

A group opens and closes by animating height to the measured open size. The last frame matches that size, so the accordion does not grow after it finishes. Closed groups stay in the document at height 0 so the next open can animate.

#### First folder open

The first click on a collapsed folder also navigates. `hx-preserve` moves `#pages-nav` and that cancels an in-flight height transition. The accordion now waits for `initNav` and then plays the same open animation.

#### Current page color

The current link keeps the hover look until a delayed snap to blue. The hover-to-active flicker no longer appears.

#### Top bar hover

An active tab uses the same `#1d2a3e` text and `#f6f9fc` background as an inactive tab on hover. That look applies only while the pointer is over the tab. Color, background, and weight transitions run in 80ms.

## Verify

```bash
cd src/Elastic.Documentation.Site && npm run test -- --testPathPatterns='pages-nav|secondary-nav'
```

```bash
FEATURE_NAVIGATION_PREVIEW=true dotnet run --project src/tooling/docs-builder --no-build -- serve --port 3000
```

Open a Guides page. Click items in the same tree and confirm the sidebar does not jump. Open a group for the first time and again after it has been toggled. Hover an active top-bar tab and confirm the text goes `#1d2a3e` only while the pointer is over it.

### DIFF
--- a/src/Elastic.Documentation.Site/Assets/main.ts
+++ b/src/Elastic.Documentation.Site/Assets/main.ts
@@ -307,7 +307,9 @@ document.body.addEventListener('htmx:afterSwap', function (event: HtmxEvent) {
     const target = event.target
     if (
         target === document.body ||
-        (target instanceof Element && target.id === 'main-container')
+        (target instanceof Element &&
+            (target.id === 'main-container' ||
+                target.id === 'content-container'))
     ) {
         window.scrollTo(0, 0)
     }
--- a/src/Elastic.Documentation.Site/Assets/main.ts
+++ b/src/Elastic.Documentation.Site/Assets/main.ts
@@ -311,7 +311,9 @@ document.body.addEventListener('htmx:afterSwap', function (event: HtmxEvent) {
             (target.id === 'main-container' ||
                 target.id === 'content-container'))
     ) {
-        window.scrollTo(0, 0)
+        if (window.scrollY !== 0) {
+            window.scrollTo(0, 0)
+        }
     }
 })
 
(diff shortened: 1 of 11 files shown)


theletterf (approved):

Mpdreamz (approved):
I wonder if we need to animate the fold and collapse? 

We should probably gate it behind 

```
@media (prefers-reduced-motion: reduce) {
```

itsalexcm:
@Mpdreamz Fold and collapse still animate only when the reader toggles a folder. Page swaps do not replay that motion (that was painting a closed sidebar for a frame).

`prefers-reduced-motion: reduce` already skips the height animation. `playFolderOpen` / `playFolderClose` snap to the end state instead. The animation is an inline `height` transition, so a CSS `@media` rule would not catch it.

PR #4014, opened 2026-09-02
docs: document applies_to authoring for settings YAML

florent-leborgne:
## Why

Kibana settings YAML files treat `automated_settings.md` as the schema source of truth. That page previously said `{settings}` YAML follows the same `applies_to` rules as page Markdown. Authors then omitted deployment keys or copied `stack: preview` onto `ech`.

## What

Documents the settings-YAML authoring contract. `stack` carries lifecycle and version. `ech`, `ece`, `eck`, `self`, and `serverless` are support flags (`ga` or `unavailable`).

For Advanced Settings that exist on only some serverless projects, nest `elasticsearch`, `observability`, or `security` under `serverless`. Write a scalar `serverless: ga` or `serverless: unavailable` when every project matches. Do not mix a scalar `serverless:` with project keys.

The syntax example YAML now uses a complete map, including `stack: preview` plus `ech: ga`.

Related: https://github.com/elastic/kibana/pull/288454


docs/syntax/automated_settings.md
```diff
@@ -73,12 +77,69 @@ groups:
...
 
+### applies_to in settings YAML [settings-yaml]
+
+The keys are the same as [Applies to](applies.md). The authoring contract is different, because each setting renders a **Supported on** line.
+
+In body Markdown, name only the keys that apply to the content. In settings YAML, list every deployment key so that line is complete and the `:deployment:` filter is explicit.
+
+| Key | What it means here | Write |
+|---|---|---|
+| `stack` | Lifecycle and version of the setting | `ga`, `preview 9.2`, or a history such as `preview 9.0-9.2, ga 9.3+`. No version means all versions. |
```
natasha-moore-elastic: > Lifecycle and version of the setting

This reads like settings have their own version (e.g. setting_v1, setting_v2)

> No version means all versions

I feel this line is prone to misinterpretation (all versions ever? or all 9.x versions?) and could be clearer

```suggestion
| `stack` | The setting's lifecycle and versions it's available in | `ga`, `preview 9.2`, or a history such as `preview 9.0-9.2, ga 9.3+`. Omit the version if the setting was added before 9.0. |
```

natasha-moore-elastic: On a related note, if a setting existed before 9.0 (for example, added in 8.18) and was removed post 9.0 (for example in 9.2), which of these would be the correct syntax?

- `stack: removed 9.2+`
- `stack: removed 9.2+, ga 9.0-9.1`
- `stack: removed 9.2+, ga 8.18-9.1`

This might be worth documenting explicitly as an example, since a lot of settings existed pre-9.0 and this might be a common scenario.

docs/syntax/automated_settings.md
```diff
@@ -73,12 +77,69 @@ groups:
...
+| `ech`, `ece`, `eck`, `self` | Supported on that deployment, or not | Always list all four. `ga` if supported. `unavailable` if not. Never a version. Never `preview`, `experimental`, `deprecated`, or `removed`. |
+| `serverless` | Supported on serverless, or not | Always list it. `ga` if supported. `unavailable` if not. Never a version. |
```
natasha-moore-elastic: Does the "Never `preview`, `experimental`, `deprecated`, or `removed`" rule apply here, too?

docs/syntax/automated_settings.md
```diff
@@ -73,12 +77,69 @@ groups:
...
         #   - setting: "[n].url"
         #     description: |
         #       REQUIRED
 ```
 
+### applies_to in settings YAML [settings-yaml]
+
+The keys are the same as [Applies to](applies.md). The authoring contract is different, because each setting renders a **Supported on** line.
+
+In body Markdown, name only the keys that apply to the content. In settings YAML, list every deployment key so that line is complete and the `:deployment:` filter is explicit.
```
natasha-moore-elastic: What does "In body Markdown" refer to?

docs/syntax/automated_settings.md
natasha-moore-elastic: What about the serverless-project keys (`security`, `observability`, etc.)? Should they not be used in YAML settings, and we just use the `serverless` key instead? If so, can we call this out somewhere? 
Or if they can be used, should we have a line for them here?

natasha-moore-elastic (commented):
Left some questions :)

docs/syntax/automated_settings.md
florent-leborgne: All of these are ok to do it just depends if we have the data, then we can be more accurate

docs/syntax/automated_settings.md
florent-leborgne:
```suggestion
In settings YAML, list every deployment key so that line is complete and the `:deployment:` filter is explicit.
```
I think it meant "the usual applies_to rule is to list only keys that apply, but in yaml we keep the list fixed"

docs/syntax/automated_settings.md
florent-leborgne: yes

natasha-moore-elastic:
Thanks for the edits!

florent-leborgne:
Moved off the fork so docs preview can create a deployment. Continue on https://github.com/elastic/docs-builder/pull/4047

Issue #3798, opened 2026-08-06
Let a package directory host a toc.yml a docset can compose into its navigation


clintandrewhall:
### Prerequisites

- [x] I have searched existing issues to ensure this feature hasn't already been requested
- [x] I have tested using the latest version of docs-builder (1.33.0)

### What problem are you trying to solve?

We publish Kibana's internal developer docs to Codex from a `docs-dev/` docset, separate from the public `docs/` docset at the repository root. The content documents packages that live under `src/platform/`, and the maintainers want each page to sit next to the code it describes, so that a change to a component and a change to its documentation appear in the same directory and the same review.

That isn't expressible today. TOC paths are resolved relative to the documentation set root — `documentation-set-navigation.md` states it directly ("All file paths in the `toc` section are relative to the documentation set root") — and there is no way to point an entry at a file above that root.

The workarounds both cost something real. Moving the docset root into the package tree colocates the content, but a repository effectively supports one docset per registry (`FindDocsetFile` returns the first match for the environment), so rooting it inside one package's directory prevents any other area of the repo from publishing internal docs later. Keeping the docset where it is means the documentation cannot live beside the code it documents.

### Proposed Solution

Let a TOC entry keep its docset-relative position while sourcing its content from elsewhere:

```yaml
toc:
  - file: index.md
  - file: feedback.md
    source: ../../src/platform/kbn-ui/feedback/feedback.md
```

`file:` stays the virtual, docset-root-relative path that drives the URL, navigation and link reference. `source:` is the on-disk location, resolved relative to the directory holding the `toc.yml`/`docset.yml`, and used only for reading.

Keeping the two separate matters: it preserves the invariant the navigation model depends on ("All paths become relative to docset root", `two-phase-loading.md`), keeps generated URLs free of `../` segments, and keeps the output path inside the output directory. Allowing a bare `- file: ../../…` instead would push a path that escapes the root through URL generation and output writing — and, as detailed under Alternative Solutions, nothing currently stops that from happening quietly.

Read-only, single files. No need for folder globs to escape the root.

### Examples and Research

The pieces already exist in the codebase, which is the main reason this looks tractable rather than structural.

A TOC entry can already reference a path outside the content root — just only for one extension. From `configure/content-set/extensions.md`:

```yaml
toc:
  - file: index.md
    detection_rules: '../rules'
```

The supporting mechanisms are also in place:

[...]

PR #4020, opened 2026-09-02
Let a TOC entry source its content from outside the docset root

clintandrewhall:
A `source:` key on a `file:` or `hidden:` table of contents entry reads the page's content from a markdown file elsewhere in the repository. A page can now live next to the code it documents without moving the documentation set root.

**Affects:** `Navigation`, `Authoring`, `Configuration`

**Prompt summary:** Implement [#3798](https://github.com/elastic/docs-builder/issues/3798). Kibana publishes internal developer docs to Codex from a `docs-dev/` docset while the packages those docs describe live under `src/platform/`, and the maintainers want each page checked in beside the code it documents so a component change and its documentation land in the same review. The ask is to let a table of contents entry keep its docset-relative position while sourcing content from above the root.

## Why

All table of contents paths resolve relative to the documentation set root, and a repository supports one docset per registry. Colocating a page with its code therefore means moving the docset root inside a package directory, which concedes the repository to a single internal docset. Keeping the docset where it is means the documentation cannot sit beside the code.

#### The new `source:` key

An entry keeps `file:` as its position in the documentation set and gains `source:` for where the content is read:

```yaml
toc:
  - file: index.md
  - file: feedback.md
    source: ../packages/kbn-ui/feedback/feedback.md
```

`file:` still drives the URL, the navigation entry, the output path and the cross-repository link reference. `source:` is resolved against the directory holding the `docset.yml` or `toc.yml` that declares the entry, and nothing else. It works on a `hidden:` entry and on the `folder:` + `file:` form.

#### Separating position from content

`DocumentationFile.RelativePath` is the page's position in the set and no longer has to equal `SourceFile` relative to the root. A sourced page parses as though it lived at that position, so relative links, images and includes inside it resolve from the virtual path rather than from where the file sits on disk. Because generated URLs and output paths derive from a path that cannot contain `../`, no page can write outside the output directory or emit `../` in a URL.

#### Registering a page the file scan never sees

The scan walks the documentation set root, so a sourced page has to be registered from the table of contents. `MarkdownFileFactory` keys it under its virtual path — the same path navigation resolves against — and registers it after everything else so it can be refused with a diagnostic when:

- the source escapes the repository checkout, either lexically or through a symlink on its path
- the source does not exist, or is not a markdown file
- the source resolves back inside the documentation set root, where the scan already owns it
- the `file:` position escapes the root, or is already held by a scanned file, an extension-generated page, or another sourced entry

A non-public docset can also declare its root `index.md` with `source:`; the root-index requirement accepts the registered position.

#### The edit link

The "Edit this page" link now derives from the file that was actually read instead of joining the docset directory with the position in the set. That is required for a sourced page, and it also corrects detection rule pages, whose `.toml` source produced a path with `../` segments in it.

#### `ResolveFileRef` rebuilds refs with `with`

Resolution previously reconstructed each ref through a three-arm `new` switch, which silently drops any key added to `FileRef` later. A `with` expression preserves the concrete ref type and every off-path key, so `source:` cannot be lost between parsing and navigation.

## Verify

```bash
./build.sh unit-test
# TocSourceTests — parsing, resolution anchoring, and the two path-shape errors
# ExternalSourceTests — registration, output path, relative links, edit link, root index, and the five registration errors
```

Then try it against this repository's own docs, pointing an entry at a file above `docs/`:

```bash
dotnet run --project src/tooling/docs-builder -- --path docs --strict
```

**Out of scope:** single markdown files only. There is no `source:` on `folder:`, `listing:` or `toc:`, and no glob form, per the issue. A sourced entry's `file:` is now checked for containment, but the general case the issue describes under Alternative Solutions — a bare `- file: ../../x.md` with no `source:` — is still accepted with no check. The author offered to track that separately and this PR leaves it alone.

**Risk:** assembled builds clone a repository with a partial checkout of `docs` only. A `source:` above that path needs the repository's `sparse_paths` or `checkout_strategy` widened in [`config/assembler.yml`](https://github.com/elastic/docs-builder/blob/main/config/assembler.yml), the same requirement the detection rules extension already has. Codex checkouts are not sparse, so the case in the issue works without a config change.

### DIFF
--- a/docs/documentation/isolated/configure/index.md
+++ b/docs/documentation/isolated/configure/index.md
@@ -44,6 +44,36 @@ A `file` can include `children` to create a virtual grouping. Children must be s
     - file: configuration.md
 ```
 
+#### `source`
+
+Reads the page's content from somewhere else in the repository, so documentation can live next to the code it describes:
+
+```yaml
+toc:
+  - file: index.md
+  - file: feedback.md
+    source: ../packages/kbn-ui/feedback/feedback.md
+```
+
+`file:` stays the page's position in the documentation set — it drives the URL, the navigation entry, the output path and the cross-repository link reference. `source:` is only where the content is read from, resolved relative to the directory holding the `docset.yml` or `toc.yml` that declares the entry.
+
+Relative links, images and includes inside a sourced page resolve from its `file:` position, not from where the file sits on disk.
+
+Also works on a `hidden:` entry, and on the `folder:` + `file:` form:
+
+```yaml
+- folder: feedback
+  file: index.md
+  source: ../packages/kbn-ui/feedback/readme.md
+```
+
+Constraints:
+
+- Single markdown files only — there is no folder or glob equivalent.
+- The source must resolve outside the documentation set root; inside it, use a plain `file:` entry.
+- The source must stay inside the repository checkout.
+- The `file:` position must be free — no file of that name on disk, and no other entry sourcing it.
+
 ### `folder:`
 
 Groups pages under a directory. Without `children`, all markdown files in the folder are included automatically:
--- a/docs/documentation/isolated/configure/index.md
+++ b/docs/documentation/isolated/configure/index.md
@@ -74,6 +74,8 @@ Constraints:
 - The source must stay inside the repository checkout.
 - The `file:` position must be free — no file of that name on disk, and no other entry sourcing it.
 
+On the assembled documentation site, repositories are cloned with a partial checkout of `docs` only. A source outside that path needs the repository's `sparse_paths` or `checkout_strategy` widened in [`config/assembler.yml`](https://github.com/elastic/docs-builder/blob/main/config/assembler.yml).
+
 ### `folder:`
 
 Groups pages under a directory. Without `children`, all markdown files in the folder are included automatically:
--- a/docs/documentation/isolated/configure/index.md
+++ b/docs/documentation/isolated/configure/index.md
@@ -71,8 +71,8 @@ Constraints:
 
 - Single markdown files only — there is no folder or glob equivalent.
 - The source must resolve outside the documentation set root; inside it, use a plain `file:` entry.
-- The source must stay inside the repository checkout.
-- The `file:` position must be free — no file of that name on disk, and no other entry sourcing it.
+- The source must stay inside the repository checkout, and neither it nor any directory on the way to it may be a symlink.
+- The `file:` position must resolve inside the documentation set root and be free — no file of that name on disk, no page generated there by an extension, and no other entry sourcing it.
 
 On the assembled documentation site, repositories are cloned with a partial checkout of `docs` only. A source outside that path needs the repository's `sparse_paths` or `checkout_strategy` widened in [`config/assembler.yml`](https://github.com/elastic/docs-builder/blob/main/config/assembler.yml).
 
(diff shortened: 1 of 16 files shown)


Mpdreamz:
That risk section is spot on and exactly why I am weary of this change. We also do redirect detection for removed/renamed files. 


This kind of breaks our contract that the url is the file. There are quite a few assumptions around that first principle. 


The system is more designed around docset/toc.yml as the composable building blocks from which you can re-arange/assemble your final site. 


Codex is much simpler it's fixed url pattern. But the system has to work for both. 

As I understand it packages want their docs to live in the package repository? 



I wonder if we should allow packages to define their own toc.yml files and the dev-docs can then reference those instead to compose it's overal nav? 


Codex should be doing sparse clones, especially for Kibana. Although it's been forever since I benchmarked that vs shallow clones, @reakaleek can you remember? 



We still need a way to ensure/declare this is a codex only thing? Codex docsets.yml have to set registry: key so we can determine if a doc set is intended for codex publishing.

clintandrewhall:
@Mpdreamz I would certainly defer to your expertise and preference here.  I was putting forward a "what about this?" agent-assisted PR to spur the conversation.

But yes: the purpose of this is have a docset that can reach out and collect documents from other places.  In this case, `kibana/docs-dev/docset.yml` can include, say, `kibana/src/platform/kbn-ui/callout/*.md`... so the docs can be colocated with the code, (and therefore stay fresh).

You _could_ opt for `kibana/docs-dev/docset.yml` referencing `kibana/src/platform/kbn-ui/callout/docs/toc.yml`... that would certainly be easy to maintain on our end.

Mpdreamz:
No worries about submitting the PR, super helpful indeed to spur the conversation: working code > issue but in this case both are great :)


If that works for y'all too, let's explore what toc.yml in packages looks like to unlock package docs collocation.

clintandrewhall:
Lemme see if I can revise this PR, but I may close it (and the issue) in favor of another.  Stay tuned!

clintandrewhall:
Closing in favour of the package-hosted `toc.yml` direction. Summarising what this PR established so the next one does not relitigate it.

**Why the `source:` mechanism does not survive.** @Mpdreamz is right that it breaks "the URL is the file". The invariant is load-bearing in more places than navigation — redirect detection for removed and renamed files keys on docset-relative paths, and `source:` makes a page's docset-relative path a name with no file behind it.

**What a package-hosted `toc.yml` has to answer.** Given `kibana/docs-dev/docset.yml` referencing `kibana/src/platform/kbn-ui/callout/docs/toc.yml`, the pages there either mount at a position declared in the referencing docset, or take their real repository path and produce `/src/platform/…` URLs. It will be the former, which means the virtual/physical split does not disappear — it moves from per file to once per subtree. The argument for it is that the split then lands on `IsolatedTableOfContentsRef`, which already rewrites child paths by prepending the TOC's own position, so it extends a composable building block instead of adding a parallel mechanism beside it. Worth settling explicitly rather than assuming the objection is answered.

**Codex is not doing sparse clones today.** `CodexGitRepository.EnableSparseCheckout` exists with no caller anywhere in the repo; only `RepositorySourcesFetcher` calls its own. So a Kibana Codex checkout currently has the whole package tree present. That cuts both ways for the new design: it works today without a config change, and if Codex adopts sparse clones the referenced package directories have to be in the sparse set.

**One correction to this PR's description.** It claimed the edit-link change also fixed detection rule pages. It does not. `UrlPath.Join` does no dot-segment removal, so the old code did emit `docs/../rules/x.toml` — but RFC 3986 §5.2.2 applies `remove_dot_segments` to absolute references too, so every compliant client normalises it before the request. The link works; it only reads as a defect. Cosmetic, not a bug.

Issue [#3798](https://github.com/elastic/docs-builder/issues/3798) stays open and is being reframed around this direction — the problem is unchanged, only the proposed mechanism moves.

PR #4034, opened 2026-09-03
Show Base URL and version on API product landings

reakaleek:
API product landings now show a copyable Base URL, license, and version when those fields exist on the OpenAPI document. The catalog table stays in place. An empty `License:` line no longer appears when the spec has no license.

**Prompt summary:** Implement https://github.com/elastic/docs-eng-team/issues/814. Add a Base URL row with copy on the product landing when `document.Servers` is present. Show license and version only when those fields exist on `info`. Keep the operation catalog. Do not replace the landing with the Introduction topic.

## Why

The product landing already prints title, description, and a full operation catalog. It does not show servers, so a reader cannot copy the Base URL. It always prints `License:` even when `info.license` is missing. Operation pages already list servers, but `.server-url` has no copy control.

Implements https://github.com/elastic/docs-eng-team/issues/814.

#### Landing extras
`LandingInfoNote` is built once from the OpenAPI document. The landing view prints Base URL, license, and version only when those values exist. License becomes a link when `info.license.url` is set. Source and last-update stay out. OpenAPI `info` does not have those fields.

#### Copy on server URLs
`initCopyButton` also targets `.server-url`. The button is appended inside that chip so hover CSS can show it. Code-block buttons still insert after the `pre`. Operation pages pick up the same copy control.

#### Markdown export
`LandingCommonMark.Product` takes the same `LandingInfoNote` and writes matching Base URL, license, and version lines.

## Verify

```bash
```

```bash
# Generate_FixtureProductLandingHtml_ContainsBaseUrlLicenseVersionAndCatalog
# Generate_FixtureProductLandingMarkdown_ContainsBaseUrlLicenseVersionAndCatalog
# Generate_TitleOnlyDocument_OmitsEmptyInfoHeadings
```

Run against local preview or staging. Production `/docs/api/*` still proxies to bump.sh.

**Out of scope:** Source and last-update rows. Those fields are not on OpenAPI `info`.

### DIFF
--- a/src/Elastic.ApiExplorer/Landing/LandingCommonMark.cs
+++ b/src/Elastic.ApiExplorer/Landing/LandingCommonMark.cs
@@ -20,13 +20,16 @@ internal static class LandingCommonMark
 		return markdown.ToString();
 	}
 
-	public static string Product(OpenApiInfo? info, IReadOnlyList<ApiOverviewRow> rows, string apiBaseUrl)
+	public static string Product(OpenApiInfo? info, LandingInfoNote note, IReadOnlyList<ApiOverviewRow> rows, string apiBaseUrl)
 	{
 		var markdown = new StringBuilder();
 		ApiCommonMark.Heading(markdown, 1, info?.Title ?? "API Documentation");
 		ApiCommonMark.Prepared(markdown, info?.Description, apiBaseUrl);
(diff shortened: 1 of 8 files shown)


Mpdreamz (approved):
Reviewed the current diff. The landing fields are guarded, the server copy button copies only the nested code text, and the previous CI run was green. This still needs a rebase because the branch is conflicting.

PR #4035, opened 2026-09-03
Add a product switcher on API reference pages

reakaleek:
API product pages now show a sidebar switcher. A reader can open another declared API or return to the hub without the browser back button.

**Affects:** API reference, Assembler builds

**Prompt summary:** Add a hub product switcher on API product pages so a reader can jump between assembled APIs and back to the hub. The tracking issue is [elastic/docs-eng-team#817](https://github.com/elastic/docs-eng-team/issues/817).

## Why
API product pages do not use the site navigation dropdown. A reader on an Elasticsearch operation page cannot open Kibana or `/docs/api/` without the browser back button.

Closes elastic/docs-eng-team#817

#### Sidebar product switcher
The left navigation on a product page shows a dropdown above the version list. The first option is Back to hub. The remaining options are the declared APIs, ordered by title. The current product is selected. The hub page hides the control.

#### Catalog collected before render
The generator reads declared `api:` entries before it fetches any spec. The assembler unions every owner's list and the first key wins. Each product render receives that shared list.

#### Isolated generate
A single-repo generate still builds the switcher from that repo's `api:` config. Sibling APIs appear only when the caller passes a shared list.

## Verify

```bash
dotnet test tests/Elastic.ApiExplorer.Tests/
# ApiHubSwitcher, ApiPagesNavRendering, OpenApiGeneratorCatalogSplit
```

On a local serve of this repo, open `/api/doc/docs-builder-elasticsearch/` and confirm the switcher lists Back to hub plus that product. An assembler preview is required to confirm a jump from Elasticsearch to Kibana.

**Out of scope** — The switcher does not sit in the header or secondary nav. Production still proxies API pages to bump.sh until the explorer ships there.

Made with [Cursor](https://cursor.com)

PR #4038, opened 2026-09-03
Apply navigation-preview styles on API explorer pages

reakaleek:
API explorer pages now receive the build's feature flags. When `navigation-preview` is on, the left sidebar uses the same styles as markdown pages.

**Prompt summary:** Fix the broken left navigation on API pages. The report pointed at a cloud-connect preview where folder labels and chevrons no longer lined up.

## Why

API explorer pages already emit the v2 sidebar markup when `navigation-preview` is on. The layout used to construct an empty `FeatureFlags` bag. The body never received the `navigation-preview` class that `pages-nav-figma.css` requires. Folder links stayed `inline-block` and chevrons wrapped under the labels.

#### Layout feature flags
API pages now pass the build's `FeatureFlags` into the layout. When `navigation-preview` is on, the body gets that class. The existing sidebar CSS then styles folder links as full-width flex rows, matching markdown pages.

#### Regression test
`CreateGlobalLayoutModel` must keep the build's flags. A rendered catalog page includes `navigation-preview` on the body when the flag is on.

#### Flag-off path
When the flag is off, the layout still reports `NavigationPreviewEnabled` as false. The previous default is unchanged.

## Verify

```bash
# CreateGlobalLayoutModel_WhenNavigationPreviewEnabled_CopiesBuildContextFeatures
# CreateGlobalLayoutModel_WhenNavigationPreviewDisabled_KeepsFlagOff
```

**Out of scope:** `_ApiPagesNav` still hardcodes the older aside wrapper classes. The inner tree styles come from `body.navigation-preview`.

### DIFF
--- a/src/Elastic.ApiExplorer/Infrastructure/ApiViewModel.cs
+++ b/src/Elastic.ApiExplorer/Infrastructure/ApiViewModel.cs
@@ -7,7 +7,6 @@ using Elastic.ApiExplorer.Operations;
 using Elastic.Documentation;
 using Elastic.Documentation.Configuration;
 using Elastic.Documentation.Configuration.Assembler;
-using Elastic.Documentation.Configuration.Builder;
 using Elastic.Documentation.Extensions;
 using Elastic.Documentation.Navigation;
 using Elastic.Documentation.Site;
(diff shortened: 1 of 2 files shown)


reakaleek:
[docs-builder#4035](https://github.com/elastic/docs-builder/pull/4035) also edits `ApiViewModel.cs`. It adds `HubSwitcherItems` next to the old empty `Features` assignment.

That PR put the product switcher in `_ApiPagesNav` because empty `Features` hid `navigation-preview` from the header. This PR removes that premise. The switcher can stay in the sidebar.

PR #4040, opened 2026-09-03
Keep the production sidebar accordion instant

itsalexcm:
Flag-off sidebar folders open and close at once again. The height animation stays on the preview nav only.

**Affects:** Site UI

**Prompt summary:** Production sidebar chevrons opened but showed no children, and the new folder slide leaked onto the flag-off accordion. Restore the old instant checkbox accordion on production without changing the preview nav.

## Why

[#4039](https://github.com/elastic/docs-builder/pull/4039) shows folder children again through the clip wrapper. The same wrap still animates height on every page. Flag-off production should keep the old CSS accordion.

#### Preview-only folder animation

`ensureSubtreeClips`, `syncFolderPanels`, and `onFolderCheckboxChange` return immediately unless `body` has `navigation-preview`. Preview keeps clips, detach, and the measured height animation.

#### Connector through the clip

The current-page connector line also matches a subtree inside `.nav-subtree-clip`. Nested current pages keep the darker rail.

## Verify

```bash
cd src/Elastic.Documentation.Site && npx jest Assets/pages-nav.test.ts
# leaves the legacy accordion alone when navigation-preview is off
```

```bash
FEATURE_NAVIGATION_PREVIEW=false dotnet run --project src/tooling/docs-builder -- serve --port 3000
```

Open `http://localhost:3000/getting-started/`. Folders should reveal children at once, with no height slide.

**Out of scope:** Chevron alignment and the clip reveal selector landed in [#4039](https://github.com/elastic/docs-builder/pull/4039).

### DIFF
--- a/src/Elastic.Documentation.Site/Assets/pages-nav.test.ts
+++ b/src/Elastic.Documentation.Site/Assets/pages-nav.test.ts
@@ -12,14 +12,6 @@ import {
     syncPagesNavFromResponse,
 } from './pages-nav'
 
-beforeEach(() => {
-    document.body.classList.add('navigation-preview')
-})
-
-afterEach(() => {
-    document.body.classList.remove('navigation-preview')
-})
-
 function pagesNav(treeId: string, heading: string, extra = ''): string {
     const headingHtml = heading
         ? `<div class="pages-nav-v2__heading"><span class="pages-nav-v2__heading-text">${heading}</span></div>`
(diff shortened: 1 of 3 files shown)


src/Elastic.Documentation.Site/Assets/pages-nav.test.ts
```diff
@@ -12,6 +12,14 @@ import {
     syncPagesNavFromResponse,
 } from './pages-nav'
 
+beforeEach(() => {
```
Mpdreamz: Why is this needed? I don't think tests should these kinds of body modifications.

itsalexcm: Agreed — that hook was only there to satisfy a body-class gate. The accordion now looks for `.pages-nav-v2-shell` (the preview markup), so tests no longer touch `document.body`.

Mpdreamz (approved):

PR #4044, opened 2026-09-04
Add a mobile open control on API pages

reakaleek:
API pages now show a hamburger on narrow viewports so a reader can open the left nav. Desktop layout is unchanged.

**Affects:** API reference

**Prompt summary:** Implement [elastic/docs-eng-team#818](https://github.com/elastic/docs-eng-team/issues/818) and keep the open control working with `navigation-preview` on and off.

## Why

API pages already render the `#pages-nav-hamburger` checkbox, the close control, and the overlay. They do not render an open control. At 390px the left nav stays off-screen. Docs pages open the same drawer from the table of contents.

Closes elastic/docs-eng-team#818

#### Mobile open control

A `md:hidden` label next to the existing checkbox targets `#pages-nav-hamburger`. The icon matches the docs TOC hamburger. The control is not gated on `NavigationPreviewEnabled`. API pages use `_ApiPagesNav`, not the flag-aware `_PagesNav`, so the same markup serves both flag paths.

#### Placement

The open label lives in the article, next to the checkbox. `_ApiToc` stays `hidden lg:block` and sits after the article, so a hamburger only there is invisible on a phone.

#### Tests

Render tests assert the label is present when `navigation-preview` is off and when it is on.

## Verify

```bash
# Render_FlagOff_IncludesOpenLabel
# Render_FlagOn_IncludesOpenLabel
```

Serve the docs set and open an API page at 390px. Confirm the hamburger opens and closes the left nav. At a desktop width the hamburger stays hidden.

**Out of scope** API layout `Features` still uses an empty `FeatureFlags` bag. This PR does not wire `navigation-preview` onto API page chrome.

### DIFF
--- a/src/Elastic.ApiExplorer/_Layout.cshtml
+++ b/src/Elastic.ApiExplorer/_Layout.cshtml
@@ -29,6 +29,7 @@ else
 				<main id="content-container" class="w-full flex flex-col relative pb-12 overflow-x-hidden">
 					<article id="markdown-content" class="content-container markdown-content md:px-4">
 						<input type="checkbox" class="hidden" id="pages-nav-hamburger">
+						@await RenderPartialAsync(_ApiMobileNavOpen.Create(Model))
 						@await RenderBodyAsync()
 					</article>
 				</main>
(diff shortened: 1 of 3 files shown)


Mpdreamz (approved):
Reviewed the current diff. The mobile open control targets the existing pages-nav checkbox, is covered for both feature-flag paths, and the previous CI run was green. This still needs a rebase because the branch is conflicting.

PR #4045, opened 2026-09-04
Strip Bump HTML from API operation descriptions

reakaleek:
Operation pages no longer show Bump path-list HTML or a literal `IMPORTANT:` prefix. Those descriptions become Myst callouts, and the Paths list still comes from overloads.

**Affects:** API reference, Search

**Prompt summary:** Implement [elastic/docs-eng-team#812](https://github.com/elastic/docs-eng-team/issues/812). Strip leaked Bump HTML from operation descriptions and map `IMPORTANT` markers to Myst admonitions so the page matches search export.

## Why

Spec descriptions include a Bump "All methods and paths" HTML block and Asciidoc prefixes such as `IMPORTANT:`. Page render sent that string through `ApiMarkdown` without stripping it. Search export already removed the HTML island. The Paths section already lists every method and path from `Overloads`.

Closes elastic/docs-eng-team#812

#### Shared description cleaner

`ApiMarkdown.Clean` drops the Bump path-list HTML and rewrites line-start `NOTE`, `TIP`, `WARNING`, `IMPORTANT`, and `CAUTION` prefixes to Myst fences. `CAUTION` maps to `{warning}` because Myst does not support `{caution}`. The function is idempotent. `Prepare` still only escapes mustache tokens and rewrites intra-API links, so parameter and property text stay unchanged.

#### Page and search call sites

The operation page model and front matter store the cleaned description. An HTML-only description no longer creates an empty Description section. Search export calls the same cleaner and no longer rebuilds a markdown method list. Each search document already has Method and Path fields. The Description wrapper is a `div`, so lists and callouts are valid HTML.

#### Tests

A fixture search description now includes the leaked HTML snippet and an `IMPORTANT:` line. Unit tests assert no leftover `operation-verb` source, Myst rewrite, and idempotence. Emission tests assert the Paths list still shows `POST` `/{index}/_search`. The CloudFront exporter check is replaced with an in-memory spec.

## Verify

```bash
# Clean_BumpIslandAndImportant_StripsHtmlAndRewritesAdmonition
# ConvertToDocuments_DescriptionWithBumpHtml_LeavesNoHtmlAndNoRebuiltList
```

**Out of scope:** These tests use `PassthroughMarkdownRenderer`, so they prove the Myst fence in markdown, not a rendered `admonition important` div. Existing Markdown admonition tests cover that render path. A wrapped `IMPORTANT:` body with no blank line stays prose.

### DIFF
--- a/src/Elastic.ApiExplorer/Export/OpenApiDocumentExporter.cs
+++ b/src/Elastic.ApiExplorer/Export/OpenApiDocumentExporter.cs
@@ -8,7 +8,6 @@ using System.Text;
 using System.Text.RegularExpressions;
 using Elastic.ApiExplorer.Infrastructure;
 using Elastic.ApiExplorer.Model;
-using Elastic.ApiExplorer.Operations;
 using Elastic.Documentation;
 using Elastic.Documentation.AppliesTo;
 using Elastic.Documentation.Configuration.Inference;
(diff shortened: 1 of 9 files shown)


Mpdreamz (approved):
Reviewed the current diff. The shared cleaner is covered for Bump HTML stripping, admonition rewriting, search export, and HTML/Markdown operation rendering, and the previous CI run was green. This still needs a rebase because the branch is conflicting.

PR #4047, opened 2026-09-06
docs: document applies_to authoring for settings YAML

florent-leborgne:
## Why

Kibana settings YAML files treat `automated_settings.md` as the schema source of truth. That page previously said `{settings}` YAML follows the same `applies_to` rules as page Markdown. Authors then omitted deployment keys or copied `stack: preview` onto `ech`.

## What

Documents the settings-YAML authoring contract. `stack` carries lifecycle and version. `ech`, `ece`, `eck`, `self`, and `serverless` are support flags (`ga` or `unavailable`).

For Advanced Settings that exist on only some serverless projects, nest `elasticsearch`, `observability`, `security`, or `vectordb` under `serverless`. Do not nest `workplace_ai`. Write a scalar `serverless: ga` or `serverless: unavailable` when every project matches. Do not mix a scalar `serverless:` with project keys.

Previous defaults go in a gated `note`. Use inline `{applies_to}` in the description for other version-scoped behavior.

The syntax example YAML now uses a complete map, including `stack: preview` plus `ech: ga`.

Supersedes #4014 (fork PR). Same-repo branch so docs preview can create a deployment.

Related: https://github.com/elastic/kibana/pull/288454


docs/syntax/settings-with-applies-example.yml
```diff
@@ -41,12 +45,19 @@ groups:
...
+          self: ga
+          serverless: ga
         options:
           - option: strict
             description: Full validation.
           - option: lenient
             description: Reduced validation.
-        warning: Lenient mode may hide configuration mistakes.
+        warning: |
+          :applies_to: stack: ga 9.2
```
natasha-moore-elastic, line 59: I don't think this warning would need an `applies_to` if the setting itself is also ga 9.2. Should we change line 47 to a different version instead? (if we want to keep this as an example of a gated note)

docs/syntax/settings-with-applies-example.yml
```diff
@@ -57,11 +68,21 @@ groups:
...
-          Default model used by the feature. The default varies by version:
+          Default model used by the feature.
 
-          * {applies_to}`stack: ga 9.2` Defaults to `model-v2`.
-          * {applies_to}`stack: ga 9.1` Defaults to `model-v1`.
+          {applies_to}`stack: ga 9.2` The model picker is also available in Discover.
         datatype: string
+        default: model-v2
+        applies_to:
+          stack: ga 9.2
```
natasha-moore-elastic:
```suggestion
          stack: ga 9.1
```
because of the note on line 84

docs/syntax/settings-with-applies-example.yml
```diff
@@ -57,11 +68,21 @@ groups:
...
+          ech: ga
+          ece: ga
+          eck: ga
+          self: ga
+          serverless: ga
+        note: |
+          :applies_to: stack: ga 9.1
+          In these versions, this setting defaults to `model-v1`.
```
natasha-moore-elastic:
```suggestion
          :applies_to: stack: ga =9.1
          In this version, this setting defaults to `model-v1`.
```
otherwise it's unclear in which version the default changed

natasha-moore-elastic (approved):
A couple of suggestions for the examples, otherwise LGTM!

docs/syntax/settings-with-applies-example.yml
florent-leborgne: Good catch. The warning tag matched the setting, so it was not actually gating anything. I moved the setting to `stack: ga 9.1` so the warning at `ga 9.2` is a real narrower scope.

florent-leborgne: Is it really me replying or an agent? Or did I create a skill to turn my own replies into agent style just to annoy people? 🤔

reakaleek (approved):

PR #4053, opened 2026-09-07
Redesign the API catalog as product cards

reakaleek:
The API catalog at `/docs/api/` is now a grid of product cards on its own layout. Each card is one link to the HTML landing page.

**Prompt summary:** Redesign the API catalog as cards with logos. The card must open the HTML page. Readers also need a fast way to view the Markdown and JSON versions. Use the OpenAPI `info.description` as the card text. The catalog page should not use Markdown typography or the API sidebar. Align the look with hub cards without sharing hub CSS. The whole card is one link. Format links stay off the card.

## Why

The catalog is a vertical list of titles with JSON and YAML download links. It is hard to scan products. There is no product mark and no description. Nested format links on a card leave only part of the card clickable.

## What

image

#### Product cards

Each product is a rounded card with the product mark, title, and teaser. The card is a single link to the HTML landing page. The mark sits on a `--color-grey-10` chip.

#### Catalog layout

The catalog page has the site header and footer only. It does not render the API sidebar or Markdown typography. Product API pages keep the existing chrome.

#### Format links

Markdown, JSON, and YAML stay on the product landing page and in the catalog Markdown export. They are not on the card.

#### Spec teaser

The card shows the first paragraph of `info.description`, with light Markdown stripped. The teaser stops at a heading even when no blank line precedes it.

#### Shared product icons

`ProductIcons` moves to `Elastic.Documentation.Site` so the catalog and hub pages share the same marks. Catalog CSS copies the hub card look under its own selectors, so a change in `hub.css` does not restyle this page.

## Verify

```bash
```

**Out of scope:** Search and filter on the catalog. Extra product icons beyond the existing map.


reakaleek:
Follow-up for the marks and `product:` mapping: [elastic/docs-eng-team#830](https://github.com/elastic/docs-eng-team/issues/830) (child of [elastic/docs-eng-team#808](https://github.com/elastic/docs-eng-team/issues/808)).

The grey chip uses a near-white cutout, so the Elasticsearch middle bar disappears. That work also covers `serverless-kibana` in `products.yml` and the docs-content `api:` `product:` ids. Out of scope for this PR.

PR #4055, opened 2026-09-08
Collect page feedback on the edge environment

theletterf:
Readers on the `edge` site can rate a page. The feedback widget already runs in `staging`, `dev`, and `preview`; this opts `edge` in too.

**Affects:** `Configuration`, `Site UI`

**Prompt summary:** The ask was to add the `edge` environment to the page feedback feature flag, following the merge of [#4054](https://github.com/elastic/docs-builder/pull/4054).

## Why

The `edge` environment carried no live `feature_flags` block — the only one it had was commented out. `PAGE_FEEDBACK` therefore resolved to its default of false, so the widget never rendered and readers on `edge` had no way to rate a page. Every other non-production environment already collects it.

#### The edge opt-in

A live `feature_flags` block replaces the commented-out one under `edge`, carrying `PAGE_FEEDBACK: true`. The key normalizes to `page-feedback` through `PublishEnvironment.ToFeatureFlags()`, the same path every other environment uses.

#### The sibling flag stays off

`NAVIGATION_PREVIEW` remains commented out inside the new block, so it keeps resolving to false exactly as before. Uncommenting the block header without preserving that would have switched on an unrelated flag whose retirement is still tracked by the `TODO` above `local`.

**Risk:** `config/assembler.yml` is baked into the changelog-scrubber Lambda as its public-bucket allowlist, so `AGENTS.md` treats any edit here as high blast-radius. This change adds two lines inside one environment's `feature_flags` block and touches no repository or allowlist entry. The scrubber and configuration suites both pass unchanged.

**Out of scope:** `prod` stays off. Turning it on there is a separate decision. The widget also writes to a `page-feedback-v1-edge` index that the API bootstraps at startup, so persistence needs an edge API deploy carrying [#4054](https://github.com/elastic/docs-builder/pull/4054).

## Verify

The two suites that read this config file:

```bash
dotnet test tests/Elastic.Documentation.Configuration.Tests/
dotnet test tests/Elastic.Changelog.Tests/
```

https://claude.ai/code/session_013qFe4JZUeEXN58iihNVsRf


tests/Elastic.Documentation.Build.Tests/FeatureFlagsTests.cs
```diff
@@ -82,4 +82,52 @@ public void ProdEnvironment_DoesNotEnableAssemblerApiExplorer()
 			features.Set(key, value);
 		features.AssemblerApiExplorerEnabled.Should().BeFalse();
 	}
+
+	[Fact]
+	public void PageFeedbackEnabled_DefaultsToFalse()
```
reakaleek: IMO, these tests are not needed. It's just testing if the flag is enabled in a certain stage.

theletterf: Fair point. Dropped them in 16cd14ba. They asserted what `assembler.yml` already states, so they'd need updating on every config change without catching anything. The shared default and env-var override paths stay covered via `ASSEMBLER_API_EXPLORER`.

Worth flagging that `/feature-flag` step 5 currently asks for exactly these ("flag defaults to false", "prod does not enable the new flag", "sibling flags stay unchanged"), which is why they were there. That guidance fits creating a flag with new code paths, but not opting an existing one into another environment. Happy to open a follow-up narrowing it if you agree.

reakaleek (approved):

PR #4056, opened 2026-09-08
Add an Edit page button and a site problem feedback reason

theletterf:
Every docs page now ends with an "Edit page" button, right aligned on the row with the feedback prompt and styled to match Yes and No. The negative questionnaire gains a "Something is broken" reason for faults in the site rather than the content, "Out of date" folds into "Technically incorrect", and "Learn how to contribute" leaves the right rail.

**Affects:** Site UI, Docs API, Navigation

**Prompt summary:** The ask was to put the "Edit this page" and "Report an issue" links at the bottom of the page, on the same row as the feedback buttons and right aligned. The rail copies stay rather than move, so a reader can act from either place. The "Learn how to contribute" link goes. A follow-up ask turned the two links into buttons that match Yes and No, shortened their labels, and put edit first. A third ask dropped the "Report issue" button and asked whether a reader could instead report a site glitch through the feedback options. A fourth removed "Out of date" and folded it into "Technically incorrect".

## Why

A reader who reaches the end of a page has already scrolled past the right rail. The rail list carries `hidden md:flex`, so a phone reader sees no CTA at all, and `report-an-issue` stays hidden until the `lg` breakpoint. The moment someone wants to act on a page is the moment they finish reading, and nothing there offers a way to do it.

Reporting a broken link or a dead control was worse. The only route was a GitHub issue form, which asks a lot of a reader who just wants to flag a glitch, and none of the negative reasons fit: `inaccurate` and `codeSampleErrors` are both about the words on the page.

At the other end, two reasons asked for a distinction the reader cannot make. Telling a page that was never right from one that stopped being right needs the page's history, so `outOfDate` and `inaccurate` collapse into one option.

#### An Edit page button under the article

The bottom of every docs page now carries an "Edit page" button, right aligned on the row with the feedback prompt. `.page-actions__action` reproduces the box that `.page-feedback__choice` draws, so the three controls on the row share a border, a radius, padding, and a colour. Hover parts company: this one turns `--color-blue-elastic`, not the green and red the feedback buttons use. A new `_PageActions` slice renders it, and its gate matches `_TableOfContents.cshtml`, so a page that shows the rail copy shows this one too.

#### Six negative reasons instead of six different ones

`siteProblem`, labelled "Something is broken", reports a fault in the site: a dead link, a control that does nothing, or a search that returns nothing. `outOfDate` leaves the questionnaire and "Technically incorrect" absorbs it, now reading "The page is wrong about how the product works, or describes an older version". `REASON_SET_VERSION` moves to 3 for both changes at once, since version 3 has not shipped. Against version 2 the negative set differs twice, so a query that counts either reason must scope itself to a version.

#### A retired reason is not a deleted one

`outOfDate` stays in `PageFeedbackReason` and on the `ThumbsDown` allowlist. Historical documents hold the value, and a browser running a cached bundle still submits it, so dropping either would answer a live reader with `400`. `Put_MultipleReasons_RecordsAllReasons` sends it and is the guard.

#### A new reason has to pass three gates

The endpoint validates each reason against the reaction it arrives with, so `siteProblem` joins the enum and the `ThumbsDown` branch of `IsReasonValidForReaction`. Without both, a submission returns `400`. The `reasons` field is a `keyword`, so neither the addition nor the retirement needs a mapping change or a reindex.

#### The feedback row shares its top border

`page-feedback.css` sets the top border and spacing on the `page-feedback` element itself, which would stop at the left column once the row holds two children. A `.page-outro` wrapper takes those over and zeroes them on the element. ApiExplorer and Codex render `page-feedback` unwrapped, so their spacing is untouched, and `:has(*)` keeps the border off the wrapper when both partials render nothing.

**Out of scope:** Codex renders `page-feedback` unwrapped in `_MarkdownLayout.cshtml`, so internal docs get no bottom row. The ApiExplorer layouts have no edit URL on their view model, so they are unchanged too. The `?edit` reveal in `main.ts` moved to `$$optional` so it clears the class on both `.edit-this-page` copies rather than the first.

## Verify

```bash
./build.sh unit-test
# PageFeedbackEndpointTests.Put_SiteProblemReason_RecordsSiteProblem — the new reason under thumbsDown
# PageFeedbackEndpointTests.Put_InvalidFeedback_ReturnsBadRequest — siteProblem under thumbsUp returns 400
# PageFeedbackEndpointTests.Put_MultipleReasons_RecordsAllReasons — retired outOfDate still accepted
# PageActionsRenderingTests — the edit button gate and the HideEditThisPage path
```

```bash
cd src/Elastic.Documentation.Site && npm run test && npm run fmt:check
```

To see both changes, build the assets and serve with the flags on:

```bash
cd src/Elastic.Documentation.Site && npm run build && cd -
FEATURE_PAGE_FEEDBACK=true FEATURE_PRIMARY_NAV=true dotnet run --project src/tooling/docs-builder serve
```

Open `http://localhost:3000/syntax/admonitions` and scroll to the end of the article. "Edit page" should match the height and centre line of Yes and No. Click No: six options, no "Out of date", and "Something is broken" fifth above "Another reason". Submitting locally returns `503` because there is no Elasticsearch, so the endpoint tests are the proof for the payload.

https://claude.ai/code/session_013qFe4JZUeEXN58iihNVsRf

### DIFF
--- a/docs/development/page-feedback.md
+++ b/docs/development/page-feedback.md
@@ -30,7 +30,7 @@ clients and historical documents remain valid.
 
 Reason set version 3 presents these positive reasons in display order:
 `solvedProblem`, `easyToUnderstand`, `accurate`, `helpfulExamples`,
-`easyToFind`, and `anotherReason`. The negative reasons are `outOfDate`,
+`easyToFind`, and `anotherReason`. The negative reasons are
 `hardToUnderstand`, `inaccurate`, `codeSampleErrors`, `missingInformation`,
 `siteProblem`, and `anotherReason`. API pages use the same stored values
 and `reason_set_version`. They pass `surface="api"` so labels and descriptions
@@ -39,13 +39,21 @@ talk about the spec and examples instead of a product how-to.
 `siteProblem` reports a fault in the site rather than in the content. A broken
 link, a control that does nothing, or a search that returns nothing belongs
 here, while `inaccurate` and `codeSampleErrors` stay with the words on the page.
-Version 2 offered the same reasons without `siteProblem`, so a query that counts
-site faults must not read version 2 documents as an absence of them.
+
+`inaccurate` also covers content that describes an older version of the product.
+Version 3 retired `outOfDate` and folded it in, because a reader cannot tell a
+page that was never right from a page that stopped being right. Two negative
+sets therefore differ from version 2: it offered `outOfDate` and lacked
+`siteProblem`. A query that counts either one must scope itself to a version.
+
+`outOfDate` remains a stored value. Historical documents hold it, and a browser
+running a cached bundle still submits it, so the endpoint still accepts it.
 
 The endpoint accepts a reason only for the reaction it belongs to. A negative
 reason sent with `thumbsUp` fails validation in `IsReasonValidForReaction` and
 the request returns `400`. Adding an option means adding it to
 `PageFeedbackReason`, to that reaction's list, and to the browser's reason set.
+Retiring one means removing it from the browser's set alone.
 
 ## Provision the index
 
--- a/docs/development/page-feedback.md
+++ b/docs/development/page-feedback.md
@@ -28,13 +28,24 @@ presented the option. Display labels may change without changing their stored va
 enum value when an option's meaning changes, and retain retired values so older
 clients and historical documents remain valid.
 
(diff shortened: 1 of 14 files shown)


theletterf:
@reakaleek Ready for review

reakaleek (approved):

PR #4068, opened 2026-09-10
Landing page: hide the empty mobile hamburger

reakaleek:
The docs landing page no longer shows a hamburger that opens an empty drawer. Article pages keep the hamburger on the left of the version picker, and the picker sits 24px under the secondary nav at every width.

**Affects:** Site UI, Navigation

**Prompt summary:** Hide the empty hamburger on the docs landing page. Later asks on the same branch keep that hamburger when `navigation-preview` fills the drawer, put it left of the version picker, and add space between the secondary nav and the version picker.

## Why

The assembler site-root nav HTML is always empty. The landing layout still rendered a hamburger when `TopNav` or the version dropdown existed, so the drawer opened blank with `navigation-preview` off. When that flag is on, the same drawer holds section and version controls, so landing needs the hamburger. On article pages the TOC row did not reverse on small screens, and `#toc-nav` only padded from 1280px up, so the version picker sat flush on the secondary nav at laptop widths.

#### Landing hamburger
The landing hamburger renders only when the mobile drawer has content: non-empty `NavigationHtml`, or `navigation-preview` with top-nav links or a version dropdown. Flag-off landing stays empty-hamburger-free.

#### Mobile TOC row
The TOC row uses `flex-row-reverse` below `md` so the hamburger is on the left and the version picker is on the right. Both controls share one baseline.

#### Version picker spacing
`#toc-nav` now has 24px top padding at every width, matching the breadcrumbs. Below 1280px the picker no longer sits on the secondary nav border.

#### Tests
Landing layout tests cover the empty-hamburger case and the `navigation-preview` drawer. TOC tests assert the reversed row.

## Verify

```bash
dotnet test tests/Navigation.Tests/ --filter "FullyQualifiedName~LandingLayoutRenderingTests|FullyQualifiedName~TableOfContentsRenderingTests"
```

On an article page, check a phone width and a window around 1100px: hamburger left of the version picker, 24px under the secondary nav. On `/` with `navigation-preview` off, there is no hamburger.

**Out of scope:** Article pages still open the pages-nav drawer from the TOC hamburger. Flag-off landing has no mobile pages-nav entry point.

### DIFF
--- a/src/Elastic.Documentation.Site/Assets/styles.css
+++ b/src/Elastic.Documentation.Site/Assets/styles.css
@@ -129,10 +129,13 @@ body {
 				max-height 0.3s ease;
 		}
 
+		#toc-nav.sidebar-nav {
+			box-sizing: border-box;
+			padding-top: 24px;
+		}
+
 		@media (width >= 1280px) {
 			#toc-nav.sidebar-nav {
-				box-sizing: border-box;
-				padding-top: 24px;
 				padding-bottom: 24px;
 			}
 		}
--- a/src/Elastic.Documentation.Site/_ViewModels.cs
+++ b/src/Elastic.Documentation.Site/_ViewModels.cs
@@ -72,6 +72,23 @@ public record GlobalLayoutViewModel
 		}
 	}
 
+	/// <summary>
+	/// True when the mobile pages drawer has something to show: a nav tree,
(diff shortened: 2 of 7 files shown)


Mpdreamz (approved):
Reviewed the rendering change and tests. The new landing predicate lines up with what the mobile pages drawer actually renders, and CI is green.

PR #4076, opened 2026-09-13
feat: add Docker dev environment via ./dev.sh

akira28:
Contributors who need to preview docs locally must install the .NET SDK and Node at specific versions before they can run anything. This PR adds a zero-install alternative: a `./dev.sh` wrapper backed by Docker that lets someone build images, serve with hot reload, and run the test suite without a local SDK.

**Affects:** Automation, CLI

**Prompt summary:** Add a Docker-based local development environment so contributors can validate rendering and authoring changes without installing .NET or Node. Structure it to sit alongside the existing `./build.sh` conventions rather than introduce a competing automation layer.

## Why

Setting up a working local environment requires matching .NET and Node versions. Contributors who only need to validate rendering or authoring changes hit this barrier before they can do anything useful. The repo already has `./build.sh` as its canonical automation entry point; a parallel `Makefile` at the root would introduce a second set of names for the same operations and drift independently.

#### `./dev.sh` entry point
A thin shell wrapper at the repo root. It delegates `serve` to `./build.sh watch-docker` and `test` to `./build.sh unit-test` inside Docker, so the Docker workflow stays in sync with the build project instead of encoding the same knowledge twice. `./dev.sh help` lists all commands.

#### Dev Docker files under `build/dev/`
`build/dev/Dockerfile`, `build/dev/docker-compose.yml`, and `build/dev/docker-bake.hcl` live under `build/dev/` rather than the repo root. A root `Dockerfile` would read as the canonical shipped image; these are SDK+Node tooling images for local development only.

#### `watch-docker` build target
A hidden `watch-docker` target added to `build/Targets.fs` and `CommandLine.fs` runs `dotnet watch` with the flags the Docker environment needs (`--no-restore`, `--no-hud`, `--port 3000`). `build/dev/serve-entrypoint.sh` delegates to this target, so the flag set is maintained in one place.

#### `DOCS_BUILDER_BIND_ADDRESS` env var
`DocumentationWebHost` reads `DOCS_BUILDER_BIND_ADDRESS` to set the Kestrel listen address, defaulting to `localhost`. Inside the container it is set to `0.0.0.0` so Docker can forward traffic to Kestrel; the host-side port is published on `127.0.0.1:3000` only, so the dev server is not exposed on all interfaces.

## Verify

```bash
./dev.sh build            # build images
./dev.sh serve            # open http://localhost:3000
./dev.sh test             # run unit-test suite
```


Mpdreamz (commented):
Thanks for putting this together. I like the goal: having a Docker-backed local development path is useful, especially for contributors who only need to preview docs without lining up local .NET/Node versions.

I would like to adjust where this plugs into the repo, though. This repository already has a root `./build.sh` backed by the `build/` project, and the Docker artifacts we ship are not SDK/dev images. The canonical Docker artifact for each shipped tool is a distroless runtime image containing the Native AOT binary. We already have one of those for each tool we publish.

Because of that, a root `Dockerfile` reads as "the canonical docs-builder image", but this file is really a local development/tooling image with the SDK, Node, npm, restore caches, and watch support. Could we move this under the build/dev area instead, for example:

- `build/dev/DockerFile`
- `build/dev/docker-compose.yml`
- `build/dev/docker-bake.hcl`
- a small root `./dev.sh` wrapper

The wrapper can call Docker/Compose with the files under `build/dev/`, and where possible it should delegate back to existing `./build.sh` targets rather than re-encoding build/test knowledge in another file. That keeps the root clean, avoids competing with the shipped container convention, and makes it clear this is a local development environment rather than one of our distribution images.

I would also prefer not to add a root Makefile here. It competes with the existing `./build.sh`/`build/` automation and introduces a second set of names for build, serve, test, clean, etc.

The `test-area` target is the part I am most worried about drifting: it duplicates the mapping from friendly names to test projects. We already maintain this kind of knowledge elsewhere, so I would rather have the Docker dev workflow call into existing build/test targets or add the missing capability to the existing build project if we need it.

On the networking side: the container binding to `0.0.0.0` is fine, and I think it is the right thing inside the container so Docker can forward traffic to Kestrel.

The host publish should be narrower, though. Right now Compose publishes this as `0.0.0.0:3000` and `[::]:3000`, so the local docs server is exposed on all host interfaces. For a local dev workflow, can we use loopback-only publishing?

```yaml
ports:
  - "127.0.0.1:3000:3000"
```

One other thing I noticed locally: the caching comments do not quite match the runtime behavior I saw.

The image seeds NuGet packages and `src/Elastic.Documentation.Site/node_modules`, but the `serve` service bind-mounts the repo over `/src` and then mounts an initially empty named volume over `/src/src/Elastic.Documentation.Site/node_modules`. On a clean run, `docker compose up --no-build serve` still restored and rebuilt frontend assets before serving.

That works, but it means the first-run path is still fairly expensive and the pre-seeded `node_modules` in the image is hidden by the volume. Can we either simplify the comments or change the volume/copy strategy so the cached image contents are actually reused by the dev service?

I gave this a local runtime pass with Orbstack:

- `docker buildx bake --file docker-bake.hcl --load tooling` succeeds.
- `docker compose up --no-build serve` starts successfully.
- `http://127.0.0.1:3000/` reaches Kestrel.
- Compose currently publishes the port on all interfaces, which is the main concrete networking concern.

So the core idea works; my ask is to fold it into the repo's existing build conventions, keep development Docker files under `build/dev/`, and avoid making a dev/tooling image look like the canonical shipped Docker artifact.

Mpdreamz (changes requested):
Marking this as changes requested to match the earlier review feedback. The main blockers are: keep the dev Docker assets under build/dev with an explicit dev.DockerFile name, avoid a root Makefile that competes with the existing build/* automation, and prefer a local ./dev.sh that calls the existing build targets/dockerfiles rather than introducing parallel build targets. The dev image should also be clearly separate from our canonical distroless AOT tool images. Binding to 0.0.0.0 is fine for the Docker dev server case as long as it remains dev-only and intentional.

akira28:
@Mpdreamz I addressed the feedback, let me know what you think

Mpdreamz (approved):
LGTM, once the AI review comments are addressed 👍

reakaleek (approved):
LGTM.

As a follow-up we could potentially add MCP or the API server?

Today, docs-builder itself exposes the API server on `serve` (only in DEBUG mode)

Might be worth exposing that port so the frontend can communicate with the backend.

This will become relevant when we enable the "Page feedback" and "Jump to page" feature

Mpdreamz:
++ I am happy with this potentially replacing our aspire setup.

PR #4079, opened 2026-09-14
Add manual page title overrides

### LINKED ISSUE #3938: Add meta_title frontmatter field to docs-builder
## Summary

The new [title guidance for Elastic Docs](https://github.com/elastic/docs-content-internal/blob/main/docs/style-guide/titles.md) defines a `meta_title` frontmatter field that does not yet exist in docs-builder.

## What it's for

`meta_title` controls the HTML `<title>` tag independently of the H1. It's needed for two cases:

- **Pattern B pages** — where the H1 intentionally omits the product name (e.g. `Query DSL`), so the page title needs a qualifier: `Query DSL - Elasticsearch | Elastic Docs`
- **API operation pages** — always need `{H1} - {Product} API | Elastic Docs` as the page title regardless of the H1

Without this field, Pattern B pages and all API operation pages fall back to `{H1} | Elastic Docs`, losing the product qualifier in search results.

## Title resolution logic

The build system should determine the page title (`<title>` tag) using the following order of precedence:

1. If `meta_title` is set in frontmatter → use `{meta_title} | Elastic Docs`
2. Otherwise → fall back to `{H1} | Elastic Docs`

Authors write the title portion only (without the suffix). The build system appends `| Elastic Docs`.

## Suffix change

As part of this work, the per-repo suffix used today (e.g. `| Elasticsearch Reference`, `| Beats`, `| Logstash`) should be replaced with a single unified suffix: `| Elastic Docs`.

This applies to all repos. The suffix is appended automatically by the build system — authors never type it manually.

## Example frontmatter

```yaml
---
navigation_title: Query DSL
meta_title: Query DSL - Elasticsearch
---
```

Renders as: `<title>Query DSL - Elasticsearch | Elastic Docs</title>`

## Reference

Title guidance doc: https://github.com/elastic/docs-content-internal/blob/main/docs/style-guide/titles.md

theletterf:
Authors can override the automatic H1 and product-derived title with `meta_title` without changing the visible heading, navigation label, or indexed title. The override also works for authored Markdown embedded in API reference pages and Codex pages.

**Affects:** Authoring, API reference, Codex builds

**Prompt summary:** Keep the product-aware title from the H1 as the SEO fallback, and provide a manual override only as a last resort. Stack this override on the automatic title behavior in [#4083](https://github.com/elastic/docs-builder/pull/4083).

## Why

The automatic product qualifier covers the standard SEO cases, but some pages still need exceptional wording in browser tabs and search metadata. Authored API Markdown also needs to carry this metadata through its separate rendering path without changing its H1 or breadcrumb.

#### Manual title precedence

The Markdown frontmatter parser accepts `meta_title:` and normalizes substitutions and Markdown formatting. A non-empty override takes precedence over the automatic H1 and product title, while the builder still owns the site suffix.

#### Embedded API Markdown

The Markdown string renderer returns parsed H1 and metadata separately for authored API pages. API layouts use the metadata title when present but retain the H1 for visible headings and breadcrumbs.

#### Shared rendering

Standard and Codex page layouts consume the same resolved page title. Parsed headings preserve their visible Markdown formatting while metadata uses plain text.

#### Authoring guidance

The frontmatter and title guides present `meta_title` as a last-resort override after the automatic product fallback. Related hub and API guidance distinguishes the visible heading from metadata.

## Verify

```bash
```

**Stack:** 2 of 2, on top of [#4083](https://github.com/elastic/docs-builder/pull/4083).

Closes [#3938](https://github.com/elastic/docs-builder/issues/3938).


Mpdreamz (approved):
I hate to introduce three ways to set titles but I can see this becoming unavoidable. 

We should still warn/error on 1 word H1 titles too I think.

Mpdreamz:
Does this fully replace the meta title? We still want our suffixes.

theletterf:
@Mpdreamz `meta_title` replaces only the page-specific part. The builder still appends the suffix. Public assembler builds use `| Elastic Docs`, while custom-branded and isolated builds keep their existing suffixes.

Besides reading the H1 and setting a meta_title, what's the other?

theletterf:
+CC @georgewallace @celiaggg Implemented as per the issue.

theletterf:
@Mpdreamz Could you reapprove this one while I research impact on sitemap?

Mpdreamz (approved):
LGTM, is https://github.com/elastic/docs-builder/pull/4079#discussion_r4004457211 still active?

georgewallace:
Please hold off on merging until I can fully test and verify this works as expected.

PR #3416, opened 2026-05-28
docs: add elastic/cli to assembler and reference navigation

theletterf:
## Summary

- Adds `cli:` to `config/assembler.yml` under `references`
- Adds `elastic/cli` to `config/navigation.yml` under Reference at `reference/elastic-cli`

## Merge order

Merge **after**:
1. [elastic/cli#377](https://github.com/elastic/cli/pull/377) is merged and `docs-build` is green on `main`
2. `elastic/docs-infra` OIDC registration PR is merged

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4081, opened 2026-09-14
Fix page-card links in assembled builds

theletterf:
## Summary

Page-card links in assembled builds resolved to the wrong URL. On the #3416 preview, body links on `reference/elastic-cli/cli` point at `/docs/3416/cli/cloud` instead of `/docs/3416/reference/elastic-cli/cli/cloud`. The sidebar is correct; only page-card hrefs drop the docset segment.

Root cause: `PageCardBlock.FinalizeAndValidate` resolved relative links through `DirectiveLinkValidator.ToHref`, which only applies the site-wide `UrlPathPrefix`. In assembled builds navigation.yml also rehomes each docset at a `path_prefix` (e.g. `reference/elastic-cli`), and `ToHref` never sees it. Inline links get this right because `DiagnosticLinkInlineParser.UpdateLinkUrl` goes through `NavigationDocumentationFileLookup`; page-cards did not.

Fix: mirror the inline-link path. Look the target up via `TryFindDocumentByRelativePath` (probing `path`, `path.md`, `path/index.md`) and `NavigationDocumentationFileLookup`, and use `FileNavigationLeaf.Url`, which already bakes in the full assembled path. Fall back to `ToHref` when the target is not in navigation.

This is pre-existing in docs-builder's own CLI reference at `reference/docs-builder/cli`; #3416 added `elastic/cli` and surfaced a second instance.

## Related

#4082 fixes a second defect visible on the same preview: the `applies_to` badge on CLI reference pages never renders because `ResolveCliReferenceRef` drops the field during toc resolution. Independent fix; both should land before #3416 rebases.

## Follow-up (not in this PR)

The hub directives (`link-card`, `hero`, `get-started`, `whats-new`) resolve hrefs through the same `ToHref` call in `HubDirectiveViewModel.PrefixUrl` and show the same defect on the preview (`examples/products/docs-builder` renders `/docs/3416/getting-started/installation` where the sidebar has `/docs/3416/reference/docs-builder/getting-started/installation`). Same fix applies; will follow separately.

## Test plan

- [x] New `PageCardBlockTests.cs` with five test classes. `PageCardRehomedDocsetLinkTests` rehomes the docset to `/reference/rehomed` before parsing — fails on the old code (`/other-page`), passes on the new (`/reference/rehomed/other-page`). Failure confirmed by disabling the lookup, not by assertion.
- [x] `PageCardRehomedAnchoredLinkTests` covers `./other-page.md#install` in the same rehomed docset — fails without the anchor split (`/other-page#install`), passes with it (`/reference/rehomed/other-page#install`). Added in response to the PR Review bot.
- [x] Full `Elastic.Markdown.Tests` suite: 2040 pass, 0 regressions.
- [ ] Preview build of this PR: confirm `reference/docs-builder/cli` body links carry the `reference/docs-builder/` segment.
- [ ] After merge, #3416 rebases and its preview shows correct `reference/elastic-cli/cli/*` body links.


theletterf:
Addressed the anchored-link finding in 5e47dd79. The anchor is now split off before the navigation probe and re-appended to whichever URL wins, so `./other-page.md#install` in a rehomed docset resolves to `/reference/rehomed/other-page#install`. New `PageCardRehomedAnchoredLinkTests` covers it and fails without the split. Full `Elastic.Markdown.Tests` suite: 2040 pass.

reakaleek (approved):

PR #4111, opened 2026-09-16
Match API Explorer sidebar to bump.sh

reakaleek:
The API Explorer sidebar now matches bump.sh by default. Tags are top-level folders, and each OpenAPI operation is a visible child. Classification folders, collapsed endpoints, and Types pages stay in the builder behind `api-nav-grouping` for a later follow-up.

**Affects:** API reference, Isolated builds

**Prompt summary:** Keep the existing grouping code. Ship bump.sh sidebar parity now so the cutover from bump.sh is smooth. Restore grouped nav after that migration.

## Why

The Elasticsearch spec ships `x-tagGroups` and many operations that share a summary. The explorer turned those into extra folders and one hidden-child row per group. Bump.sh does not. Readers who move from bump.sh to this explorer would see a different tree.

#### Default sidebar
`CreateNavigation` now emits tags and one visible `OperationNavigationItem` per operation. Authentication and Servers pages stay. Types pages are not in the default tree.

#### Grouped nav flag
`ApiNavGroupingEnabled` (`FEATURE_API_NAV_GROUPING`) restores classification folders, endpoint collapsing, and Types pages. The flag defaults to off and is not set in assembler environments.

#### Tests
`ApiNavParityTests` pins the default tree and the flag-on tree. Existing classification tests opt in to the flag.

## Verify

```bash
# CreateNavigation_XTagGroupsWithMultipleGroups_Default_TagsHangOffLanding
# CreateNavigation_SharedSummaryOperations_Default_EachOperationIsVisible
dotnet test tests/Elastic.Documentation.Build.Tests/ --filter FeatureFlags
```

**Out of scope.** Turning grouped nav on in staging or prod. That is a follow-up after the bump.sh cutover.


Mpdreamz (changes requested):
> The Elasticsearch spec ships x-tagGroups and many operations that share a summary. The explorer turned those into extra folders and one hidden-child row per group. Bump.sh does not. Readers who move from bump.sh to this explorer would see a different tree. 


I think this takes the parity too far. The extra grouping is 100% nicer. Would hate to not ship with it.

Mpdreamz (approved):
> Restore grouped nav after that migration.

I see now that the intend is too move later, not blocking this but better nav from the get go is an net win IMO

We also get one first impression on our explorer. 

Will leave that call to you @reakaleek 😸

reakaleek:
> > Restore grouped nav after that migration.
> 
> I see now that the intend is too move later, not blocking this but better nav from the get go is an net win IMO
> 
> We also get one first impression on our explorer.
> 
> Will leave that call to you @reakaleek 😸

I think users are more afraid of change than of something that is worse.

So even if the grouping is better, I'm sure there will be complaints.

My goal is to have a smooth migration and have exactly the same URLs so we don't affect SEO or bookmarks.

As soon as the API specs are in our system, we can actually think about changing the grouping. Then we have the ability to iterate faster, add analytics, add redirects or whatever is necessary to achieve a less painful change for endusers.

PR #4112, opened 2026-09-16
fix: MCP server metadata: name, version, and protocol version

akira28:
The MCP server now reports the correct service name, informational version, and pinned protocol version during the MCP handshake. The public profile's server instructions also surface the v9-and-later scope.

**Affects:** MCP

**Prompt summary:** Running `mcpcontract dump` against the live endpoint revealed four metadata problems: the server reported its name as `bootstrap` (the Fargate assembly name), its version as `1.0.0.0` (default assembly version), its protocol version by echoing whatever the client sent, and gave no indication of version scope in its instructions. This PR fixes all four.

## Why

The ECS Fargate container uses `<AssemblyName>bootstrap</AssemblyName>` as a deployment convention. The SDK's `McpServerOptions.ServerInfo` defaults to the assembly name and assembly version, so MCP clients that inspect the handshake see `bootstrap / 1.0.0.0` instead of the actual service identity. The null default for `ProtocolVersion` causes the SDK to echo the client's requested version, making the server appear to support protocol revisions it was never tested against.

#### Server identity

`McpServerOptions.ServerInfo` is now set explicitly using the profile's `ServiceName` as the name and the assembly's `AssemblyInformationalVersionAttribute` as the version. The public deployment reports `docs-mcp` and the internal one reports `codex-mcp`, both with the correct semver built into the binary.

#### Protocol version

`options.ProtocolVersion` is pinned to `"2025-11-25"`. Previously the SDK echoed the client's requested version; this pins the server to the latest protocol it actually supports regardless of what the client declares.

#### v9 scope in server instructions

The public profile's `Introduction` now ends with "All content covers version 9 and later." This sentence appears in the system prompt that MCP clients receive during the handshake, so clients understand the version scope without needing an explicit filter parameter.

#### Tests and warnings

Two service-name facts (`PublicProfile_ServiceName_IsDocsMcp`, `InternalProfile_ServiceName_IsCodexMcp`) and an updated snapshot test cover the new identity and introduction changes. `McpHandshakeTests` adds an end-to-end regression test: it spins up the real server via `WebApplicationFactory<Program>`, sends a JSON-RPC `initialize` request, and asserts `serverInfo.name`, `serverInfo.version`, and `protocolVersion` from the SSE-framed endpoint response — a wiring regression in `Program.cs` would fail this test while leaving all unit tests green. Unnecessary `!` null suppressions and a lambda IDE0350 warning in `McpToolTelemetryTests` are also cleared. Duplicate `PackageReference` entries already provided by `tests/Directory.Build.props` are removed from the test project file.

## Verify

```bash
dotnet test tests/Mcp.Remote.Tests/
```


src/api/Elastic.Documentation.Mcp.Remote/Tools/SearchTools.cs
```diff
@@ -34,6 +34,7 @@ public async Task<string> SemanticSearch(
 		[Description("Number of results per page (default: 10, max: 50)")] int pageSize = 10,
 		[Description("Filter by product ID (e.g., 'elasticsearch', 'kibana')")] string? productFilter = null,
 		[Description("Filter by navigation section (e.g., 'reference', 'getting-started')")] string? sectionFilter = null,
+		[Description("Filter by product version (e.g., '9.0', '8.19', '7.17', '9.0+')")] string? versionFilter = null,
```
Mpdreamz: We are not ready for this yet :) we only index `9.0+` content

src/api/Elastic.Documentation.Mcp.Remote/Program.cs
```diff
@@ -62,9 +64,21 @@
 	// In SDK 1.4+, stateless and SSE are mutually exclusive; EnableLegacySse (default false)
 	// cannot be combined with Stateless = true. SSE-only clients should use the mcp-remote bridge:
 	// npx -y mcp-remote https://<host>/docs/_mcp
+	var serverVersion = Assembly.GetExecutingAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion?.Split([
```
Mpdreamz: Use this syntax instead:

https://github.com/elastic/docs-builder/blob/cd04fff7b300fba3b2aee0ee31c4d49f15723469/src/Elastic.Documentation.Configuration/BuildContext.cs#L24-L28

Any reason we don't want the prerelease or build information? 


The build information is not present on 'real versions' e.g builds from the `M.N.P` tags.

Mpdreamz (changes requested):

src/api/Elastic.Documentation.Mcp.Remote/Tools/SearchTools.cs
akira28: uh ok, should we limit this to 9.x or remove the filter altogether

akira28: let me drop it, we can document it somehow

Mpdreamz: remove the filter for now.

src/api/Elastic.Documentation.Mcp.Remote/Program.cs
akira28: can you elaborate on this? If I use that syntax I would show the prerel or build info? (sorry I'm not versed on .net).

Mpdreamz: It would show the full version (this variant is also slightly safer)

Full version is with build metadata for prereleases for tagged releases this is just e.g `1.10.0`.

src/api/Elastic.Documentation.Mcp.Remote/McpServerProfile.cs
```diff
@@ -37,7 +37,7 @@ McpFeatureModule[] Modules
 		"docs",
 		"",
 		"Elastic documentation",
-		"Use this server to {capabilities} Elastic product documentation published at elastic.co/docs.",
+		"Use this server to {capabilities} Elastic product documentation published at elastic.co/docs. All content covers version 9 and later.",
```
reakaleek, line 40: This could potentially change in the future when we migrate /guide to docs-builder.

Do you think it's still worth changing the wording?

akira28: apparently there's some confusion on the version returned, once we add more versions we can add the filter as asked by @KOTungseth and change the description

tests/Mcp.Remote.Tests/McpServerInstructionTests.cs
```diff
@@ -57,6 +57,12 @@ public void Triggers_AreProfileSpecific()
 		internalInstructions.Should().NotContain("Elastic product names");
 	}
 
+	[Fact]
+	public void PublicProfile_ServiceName_IsDocsMcp() => McpServerProfile.Public.ServiceName.Should().Be("docs-mcp");
```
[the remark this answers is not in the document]
akira28: Added `McpHandshakeTests.cs` with a `WebApplicationFactory<Program>`-based test that sends a real MCP initialize request to the in-process server and asserts `serverInfo.name`, `serverInfo.version`, and `protocolVersion` from the SSE-framed response. Also added `Microsoft.AspNetCore.Mvc.Testing` to the test project. All 20 tests pass.

tests/Mcp.Remote.Tests/McpHandshakeTests.cs
```diff
@@ -0,0 +1,67 @@
...
+		using var client = factory.CreateClient();
+
+		const string initRequest = /*lang=json,strict*/
+			"""
+			{
+			  "jsonrpc": "2.0",
+			  "id": 1,
+			  "method": "initialize",
+			  "params": {
+			    "protocolVersion": "2025-11-25",
```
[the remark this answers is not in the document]
akira28: Fixed: the client now sends `"2024-11-05"` (an older protocol version) instead of the server-pinned `"2025-11-25"`. If `options.ProtocolVersion` regresses to `null`, the SDK would echo `"2024-11-05"` back and the assertion against `"2025-11-25"` would fail, making the regression visible.

tests/Mcp.Remote.Tests/McpHandshakeTests.cs
```diff
@@ -0,0 +1,67 @@
...
+		);
+		jsonLine.Should().NotBeNull("response body should contain an SSE data line");
+		var json = (jsonLine ?? "data:")["data:".Length..].Trim();
+		using var doc = JsonDocument.Parse(json);
+
+		var result = doc.RootElement.GetProperty("result");
+		result.GetProperty("protocolVersion").GetString().Should().Be("2025-11-25");
+
+		var serverInfo = result.GetProperty("serverInfo");
+		serverInfo.GetProperty("name").GetString().Should().Be(McpServerProfile.Public.ServiceName);
```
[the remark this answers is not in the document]
akira28: Fixed: `Program.cs` reads `MCP_SERVER_PROFILE` via `Environment.GetEnvironmentVariable` before DI overrides are possible, so `UseSetting` had no effect. The test now saves the current value, sets it to `"public"`, and restores it in a `finally` block so the profile is pinned regardless of the calling environment.

reakaleek (approved):

tests/Mcp.Remote.Tests/McpHandshakeTests.cs
```diff
@@ -0,0 +1,83 @@
...
+			jsonLine.Should().NotBeNull("response body should contain an SSE data line");
+			var json = (jsonLine ?? "data:")["data:".Length..].Trim();
+			using var doc = JsonDocument.Parse(json);
+
+			var result = doc.RootElement.GetProperty("result");
+			result.GetProperty("protocolVersion").GetString().Should().Be("2025-11-25");
+
+			var serverInfo = result.GetProperty("serverInfo");
+			serverInfo.GetProperty("name").GetString().Should().Be(McpServerProfile.Public.ServiceName);
+			serverInfo.GetProperty("version").GetString().Should().NotBeNullOrEmpty();
```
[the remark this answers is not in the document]
akira28: Fixed: the test now reads `AssemblyInformationalVersionAttribute` from `typeof(Program).Assembly` — the same logic as `Program.cs` — and asserts an exact match. A regression to `1.0.0.0` or `"0.0.0"` would produce a mismatch and fail the assertion.

Mpdreamz (approved):

PR #4119, opened 2026-09-16
Add five missing product IDs to the catalog

ismisepaul:
Adds catalog keys for five Elastic products that have none:

- `connectors` — Elastic Connectors
- `elastic-maps-server` — Elastic Maps Server
- `elastic-package-registry` — Elastic Package Registry
- `enterprise-search` — Elastic Enterprise Search
- `synthetics-recorder` — Elastic Synthetics Recorder

All five set `public-reference: false` and `release-notes: false`, so nothing in the docs pipeline changes.


config/products.yml
```diff
@@ -65,6 +65,13 @@ products:
   terraform-google-edot-cf:
     display: 'Terraform Google Elastic Cloud Forwarder'
     versioning: 'terraform-google-edot-cf'
+  connectors:
+    display: 'Elastic Connectors'
+    repository: 'connectors'
+    versioning: 'stack'
+    features:
+      public-reference: false
+      release-notes: false
```
Mpdreamz, line 73: `repository: connectors` defaults to the product key (`Repository = kvp.Value.Repository ?? kvp.Key` in `ProductExtensions.cs`). `versioning: stack` has no effect here — non-`public-reference` products resolve to `VersioningSystem.None` when their key does not parse as a versioning system ID.

```suggestion
  connectors:
    display: 'Elastic Connectors'
    features:
      public-reference: false
      release-notes: false
```

config/products.yml
```diff
@@ -155,6 +162,18 @@ products:
     display: 'Elastic Documentation'
     features:
       public-reference: false
+  elastic-maps-server:
+    display: 'Elastic Maps Server'
+    versioning: 'stack'
```
Mpdreamz, line 166: `versioning: stack` has no effect for `public-reference: false` products — the loader falls back to `VersioningSystem.None` when the key (`elastic-maps-server`) does not parse as a versioning system ID.

```suggestion
  elastic-maps-server:
    display: 'Elastic Maps Server'
    features:
      public-reference: false
      release-notes: false
```

config/products.yml
```diff
@@ -203,6 +222,12 @@ products:
     display: 'Elasticsearch for Apache Hadoop'
     repository: 'elasticsearch-hadoop'
     versioning: 'stack'
+  enterprise-search:
+    display: 'Elastic Enterprise Search'
+    versioning: 'stack'
```
Mpdreamz: Same as above — `versioning: stack` is a no-op for `public-reference: false` products.

```suggestion
  enterprise-search:
    display: 'Elastic Enterprise Search'
    features:
      public-reference: false
      release-notes: false
```

config/products.yml
```diff
@@ -263,6 +288,12 @@ products:
   serverless-vector-database:
     display: 'Elasticsearch Vector Database Serverless'
     versioning: 'serverless'
+  synthetics-recorder:
+    display: 'Elastic Synthetics Recorder'
+    repository: 'synthetics-recorder'
```
Mpdreamz: `repository: synthetics-recorder` defaults to the product key — same as the `connectors` case.

```suggestion
  synthetics-recorder:
    display: 'Elastic Synthetics Recorder'
    features:
      public-reference: false
      release-notes: false
```

Mpdreamz (commented):
Four redundant fields across the new entries — two patterns:  when the product key already matches the repo name (defaults to the key), and  on  products (they resolve to  regardless, because their key doesn't parse as a versioning system ID).\n\n's  is needed (key ≠ repo name) and the  blocks on all four must stay ( — both features on — is the default when  is absent entirely).

config/products.yml
Mpdreamz, line 73: `repository: 'connectors'` defaults to the product key — `Repository = kvp.Value.Repository ?? kvp.Key` in `ProductExtensions.cs`. `versioning: 'stack'` has no effect here: when the key doesn't parse as a `VersioningSystemId`, the loader falls back to `VersioningSystem.None` for `public-reference: false` products.

```suggestion
  connectors:
    display: 'Elastic Connectors'
    features:
      public-reference: false
      release-notes: false
```

config/products.yml
Mpdreamz, line 166: `versioning: 'stack'` has no effect for `public-reference: false` products — `elastic-maps-server` doesn't parse as a `VersioningSystemId`, so the loader falls back to `VersioningSystem.None` regardless.

```suggestion
  elastic-maps-server:
    display: 'Elastic Maps Server'
    features:
      public-reference: false
      release-notes: false
```

config/products.yml
Mpdreamz: Same as `elastic-maps-server` — `versioning: 'stack'` is a no-op for `public-reference: false` products.

```suggestion
  enterprise-search:
    display: 'Elastic Enterprise Search'
    features:
      public-reference: false
      release-notes: false
```

config/products.yml
Mpdreamz: `repository: 'synthetics-recorder'` defaults to the product key — same pattern as `connectors`.

```suggestion
  synthetics-recorder:
    display: 'Elastic Synthetics Recorder'
    features:
      public-reference: false
      release-notes: false
```

Mpdreamz (commented):
Four redundant fields across the new entries — two patterns: `repository:` when the product key already matches the repo name (defaults to the key via `Repository = kvp.Value.Repository ?? kvp.Key` in `ProductExtensions.cs`), and `versioning: 'stack'` on `public-reference: false` products (they resolve to `VersioningSystem.None` regardless, because their key doesn't parse as a versioning system ID and the loader falls back to `None` for non-public-reference products).

`elastic-package-registry`'s `repository: 'package-registry'` is needed (key ≠ repo name). The `features:` blocks on all four must stay — omitting `features:` entirely gives `ProductFeatures.All` (both features on).

config/products.yml
ismisepaul: For clarity, we consume this file as the canonical list of Elastic product identifiers. We use `versioning` to distinguish products that follow Stack version numbers from products with independent release lines. That’s why three of the five entries declare `versioning`, while two omit it.

The `repository: 'connectors'` change is already applied in https://github.com/elastic/docs-builder/pull/4119/commits/d11d838de19b7fef6512979dc80d04d06747eda5. It explicitly restates the key, which is otherwise supplied by `Repository = kvp.Value.Repository ?? kvp.Key`. This is committed to the branch.

In that commit I've kept `versioning: 'stack'`, which means the suggestion can’t be committed as written. This value doesn’t appear to be a no-op: `ResolveVersioningSystem` receives `kvp.Value.Versioning ?? kvp.Key`, so an explicitly declared value is parsed before the product key is considered. `"stack"` matches `VersioningSystemId.Stack` through `[Display(Name = "stack")]` because `TryParse` is called with `allowMatchingMetadataAttribute: true`. It therefore resolves to `Stack`, and the subsequent `??= VersioningSystem.None` fallback does not run.

That fallback applies when `versioning` is absent, as it is for `synthetics-recorder` and `elastic-package-registry`. `ProductFeaturesTests.ProductWithoutPublicReference_GetsNoneVersioningSystem` verifies this behaviour using `docs-builder`, which also omits `versioning`. By contrast, `agent-skills` and `transport-go` both have `public-reference: false` and explicitly declare a versioning system.

Because Connectors ships on Stack version numbers, I’d prefer to keep `versioning: 'stack'`.

config/products.yml
ismisepaul: Not changed on the branch yet. Same as the connectors thread https://github.com/elastic/docs-builder/pull/4119#discussion_r4028600809: the declared value is parsed before the key, so `versioning: 'stack'` resolves rather than falling back to None, and we read this field to tell stack-versioned products from independently versioned ones.

Maps Server ships 8.19.x and 9.3–9.5, so it appears to follow the stack versioning system.

config/products.yml
ismisepaul: Agreed on this one, removed in https://github.com/elastic/docs-builder/pull/4119/commits/6a29bf41c41bb0b7b69ecb695a5ec4a2dd9f8c1a though not because it's a no-op. In versions.yml, stack is base 9.0, current 9.5.4. Enterprise Search doesn't exist in 9.x; 8.19 is its last line, so declaring that it follows the current stack system misstates the product.

config/products.yml
ismisepaul: Already applied in https://github.com/elastic/docs-builder/commit/d11d838de19b7fef6512979dc80d04d06747eda5 it restates the key. The thread shows as outdated because I edited the file by hand rather than committing your suggestion; my mistake.

config/products.yml
ismisepaul: Answered in https://github.com/elastic/docs-builder/pull/4119#discussion_r4028600809

config/products.yml
ismisepaul: Answered in https://github.com/elastic/docs-builder/pull/4119#discussion_r4028612353

config/products.yml
ismisepaul: Answered in https://github.com/elastic/docs-builder/pull/4119#discussion_r4028619811

config/products.yml
ismisepaul: Answered in https://github.com/elastic/docs-builder/pull/4119#discussion_r4028624245

config/products.yml
Mpdreamz: That makes sense!

Mpdreamz (approved):
LGTM! Welcome aboard :)

PR #4132, opened 2026-09-17
feat: add Docker dev support for the MCP server via ./dev.sh

akira28:
Extends the dev Docker setup from #4076 to cover the MCP server. Running `./dev.sh serve-mcp` starts the MCP HTTP server at `http://localhost:8080/docs/_mcp` without a local .NET SDK install. `./dev.sh build-mcp` builds only the MCP image; extra flags (e.g. `--progress=plain`, `--no-cache`) pass through to bake.

**Affects:** Dev tooling

## Why

The MCP server is deployed as a separate service but had no dev workflow equivalent to `./dev.sh serve`. Developers needed a local .NET SDK to run and test it.

#### MCP bake target and compose service
A new `mcp` target in `docker-bake.hcl` points at the production Dockerfile so there is no duplicated build logic. A matching `mcp` service and `mcp-artifacts` volume are added to `docker-compose.yml`.

#### dev.sh commands
`build-mcp`, `serve-mcp`, `serve-mcp-detached`, and `stop-mcp` follow the same conventions as the existing `build`/`serve`/`stop` commands. Extra arguments to `build-mcp` pass through to bake.

#### MCP Dockerfile — single AL2023 base image
The Dockerfile previously used a `node:22-bookworm-slim` stage to copy Node binaries into the AL2023 build image. AL2023 ships `nodejs22` and `nodejs22-npm` packages, so the extra stage is gone and Node is now installed via `dnf` alongside the other build dependencies.

#### MCP Dockerfile — build fixes
Three bugs in the production Dockerfile were fixed. `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1` was set in the build stage, causing the regex source generator to reject `"en-US"` culture names (`SYSLIB1042`) — it now lives in the runtime stage only. The RID detection now uses the same `TARGETOS`/`TARGETARCH` → `RID="${TARGETOS}-${arch}"` pattern as `Elastic.Documentation.Api/Dockerfile` instead of a custom `PUBLISH_RID` build arg. The publish output is read from the standard `.artifacts/publish/…` path rather than a custom `--output` directory.

## Verify

```bash
./dev.sh build-mcp
./dev.sh serve-mcp
# connect an MCP client to http://localhost:8080/docs/_mcp
```


theletterf (approved):

src/api/Elastic.Documentation.Mcp.Remote/Dockerfile
```diff
@@ -7,20 +7,16 @@ WORKDIR /app
...
-RUN dnf install -y git
-RUN dnf install -y clang
-
-## temporary see: https://github.com/amazonlinux/amazon-linux-2023/issues/1025
-RUN dnf install -y tar
-RUN dnf install -y findutils
-RUN curl -L https://dot.net/v1/dotnet-install.sh -o dotnet-install.sh
-RUN chmod +x ./dotnet-install.sh
-RUN ./dotnet-install.sh
+RUN dnf update -y \
```
reakaleek, line 10: Thank you for reducing the layers 😺

reakaleek (approved):

PR #4138, opened 2026-09-18
feat: add Docker dev support for the docs API via ./dev.sh

akira28:
Extends the dev Docker setup to cover the docs API. Running `./dev.sh serve-api` starts the API at `http://localhost:8081` without a local .NET SDK install. `./dev.sh build-api` builds only the API image; extra flags pass through to bake.

**Affects:** Dev tooling

**Stack:** 2 of 2, on top of #4132

## Why

The docs API had no dev workflow equivalent to `./dev.sh serve-mcp`. Developers needed a local .NET SDK to run and test it.

#### API bake target and compose service
A new `api` target in `docker-bake.hcl` points at the production Dockerfile. A matching `api` service and `api-artifacts` volume are added to `docker-compose.yml`. The service maps host port `8081` → container port `8080` to avoid conflicting with the MCP server.

#### dev.sh commands
`build-api`, `serve-api`, `serve-api-detached`, and `stop-api` follow the same conventions as the existing MCP commands.

#### API Dockerfile alignment
The Dockerfile is updated to match the improvements already applied to the MCP Dockerfile: dnf layers consolidated, Node installed from the AL2023 `nodejs22`/`nodejs22-npm` packages instead of the older `npm` package (Node 18), `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1` moved to the runtime stage only, and the `dotnet-install.sh` workaround comment cleaned up.

## Verify

```bash
./dev.sh build-api
./dev.sh serve-api
# API available at http://localhost:8081
```


src/api/Elastic.Documentation.Api/Dockerfile
```diff
@@ -39,5 +35,6 @@ WORKDIR /app
 COPY --from=base /app/publish/ /app/
 RUN chmod +x /app/docs-builder-api
 EXPOSE 8080
-ENV ASPNETCORE_URLS=http://+:8080
+ENV ASPNETCORE_URLS=http://+:8080 \
+    DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1
```
reakaleek, line 39: Can you explain what `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT` does?

akira28: some info here https://learn.microsoft.com/en-us/dotnet/core/runtime-config/globalization
But basically I had to move the env as it was in another stage, and needed here for the build

reakaleek (approved):

PR #4140, opened 2026-09-18
Onboard docs-builder to the release-notes shared workflows

theletterf:
## Why

docs-builder's release notes were wired to `changelog-bundle.yml@v1`, one of the `changelog-*` shared workflows that `elastic/docs-actions` has replaced. Those are dead: [docs-actions#313](https://github.com/elastic/docs-actions/pull/313) consolidated the four consumer files into one `release-notes.yml` caller and left the old ones as shims, and the intent is to move everyone off and delete them.

The old caller had also never run. `changelog-publish.yml` triggered on `on: release: [published]`, but `release.yml` publishes the draft with `GITHUB_TOKEN` (`gh release edit --draft=false`), and GitHub does not start workflow runs from events created by that token. The workflow had zero runs across every release since it was added.

## What

Adds the three new callers. docs-builder is an on-release product, so `release-notes.yml` is the only one the onboarding validator requires (`ChangelogOnboardingValidationService.cs:59`); the other two are opt-in and both enabled.

| File | Trigger | Role |
|---|---|---|
| `release-notes.yml` | `pull_request`, `push` to main | Validates the type label and any entry files; uploads entries to the changelog pool |
| `release-notes-comments.yml` | `workflow_run` | Sticky changelog comment on the PR |
| `release-notes-changelog-file.yml` | `workflow_run` | Commits `docs/changelog/{pr}.yaml` to non-fork PR branches |
| `release-notes-backfill.yml` | `workflow_dispatch` | Bundles a past tag |

Release-time bundling moves into `release.yml` as `create-release-notes-bundle` + `publish-release-notes-bundle`, instead of the shared workflow's `bundle-on-release` gate. That gate is also keyed on the `release` event, so it would hit the same `GITHUB_TOKEN` wall. Running it inside the release workflow ties the bundle to the tag that was just cut and needs no token change. The create/publish split mirrors docs-actions: only the publish job assumes the OIDC role.

Config changes:

- `rules.bundle.exclude_types: other` — `rules.create` does not apply at bundle time, and `changelog gh-release` derives types from the release body's headers, where Maintenance, Chore, Automation, CI and Redesign all map to `other` (`ReleaseNoteParser.cs:120-126`). `rules.create` stays as-is because `validate-labels` reads it to resolve skip labels.
- The 13 demo entries move to `docs/changelog/examples/`. The new push-time upload would otherwise publish them into the changelog pool, and one (`sample-feature-099.yaml`) carries a fake `prs: "1234"` that could match a real PR at bundle time. Entry discovery is `TopDirectoryOnly`, so the subdirectory is skipped.
- `products.yml` spells out `release-notes: on-release` for docs-builder. Omitting the key already resolved to the same path.
- `zizmor.yml` ignores `dangerous-triggers` on the two `workflow_run` callers, with the reason inline. Neither checks out PR code.

## Testing

Ran the PR gate that `release-notes.yml` executes, against the new config, for all 12 labels `required-labels.yml` permits:

- `breaking`, `feature`, `enhancement`, `bug`, `fix`, `documentation` → `status=ok` with the expected type, exit 0
- `chore`, `dependencies`, `automation`, `ci`, `redesign`, `changelog:skip` → `status=skipped`, exit 0
- an unmapped label → `no-label`, exit 1

So no PR in this repo can fail the new gate for a missing type label. `dotnet test tests/Elastic.Changelog.Tests` passes (1330 tests). Every YAML file in the diff parses.

## Not verified yet

The `validate` job runs on this PR from the branch, so that path is exercised here. The two `workflow_run` callers only fire once they are on `main`, and the release path cannot run until the next release. I have not claimed those end to end.

## Follow-ups

- No page renders these bundles yet. That needs `release_notes:` in `docs/_docset.yml` plus a `:::{changelog}` page in `:cdn:` mode.
- `bundle.owner`, `bundle.repo` and `link_allow_repos` in `docs/changelog.yml` are marked obsolete in code ("derived automatically", "no longer read"). They load without a warning today, so I left them.


theletterf:
Good catch @Mpdreamz — you're right that we'd end up with two comments on every PR. `release-notes-comments.yml` posts a more informative one (it shows the actual generated changelog entry or the skip decision), so I've silenced the `required-labels` comment by setting `add_comment: false` in `required-labels.yml`. The CI gate is unchanged, only the comment goes away.

akira28 (approved):

PR #4162, opened 2026-09-21
Allow strict builds before the first release

theletterf:
A registered release-notes product without published bundles now emits a hint. Strict builds can complete while the product waits for its first release.

**Affects:** Configuration, Release notes, Isolated builds

**Prompt summary:** Make the registered-but-unpublished release-notes state a hint so the documentation build for [#4141](https://github.com/elastic/docs-builder/pull/4141) can pass, and request review from `Mpdreamz`.

## Why

A product can be registered in `products.yml` before its first bundle exists. `ReleaseNotesFetcher` reports that valid bootstrap state as a warning, and `--strict` turns the warning into a failed build.

#### Diagnostic severity

A missing CDN registry for an explicitly declared, registered product emits a hint instead of a warning. Real fetch failures retain their existing error or warning severity.

#### Bootstrap behavior

The resolver continues to record the product as declared but unpublished. The `{changelog}` directive therefore renders an empty changelog with its existing hint until the first bundle arrives.

#### Regression coverage

A focused test verifies that a required product with no registry produces one hint, no warnings or errors, and remains tracked as an unpublished declared product.

## Verify

```bash
dotnet test tests/Elastic.Documentation.Configuration.Tests/
```


docs/changelog/4162.yaml
```diff
@@ -0,0 +1,7 @@
+prs:
```
Mpdreamz: No changelogs/*.yaml files in this repository :) 

Otherwise LGTM

Mpdreamz (changes requested):

docs/changelog/4162.yaml
theletterf: Removed the generated changelog entry. This repository uses the GitHub release path and does not keep per-PR changelog YAML files.

Mpdreamz (approved):

PR #4178, opened 2026-09-21
Add bundle.releases config block; remove source: github_release

Mpdreamz:
`bundle.releases` maps GitHub release tags, unified release products, and serverless triggers to named bundle profiles. This replaces `source: github_release` on profiles — it was redundant with the `gh-release` command — with an explicit, validated routing table in `changelog.yml`.

**Affects:** Release notes, Configuration, CLI

**Prompt summary:** Add a `bundle.releases` config block to `changelog.yml` with three named sub-keys (`github`, `unified`, `serverless`) that map release triggers to profiles. Deprecate (warn, not error) `source: github_release`, `output_products`, and `output_directory` on profiles. Introduce a simpler `product:` field validated against `products.yml`.

## Why

The `source: github_release` profile field duplicated what the `gh-release` command already does: fetch the PR list from the commit range between two release tags. There was no way to declare which profile a GitHub release should use without modifying workflow YAML, and no way to express monorepo tag-prefix scoping at the config level. A GitHub code search confirmed every `output_products:` usage in the elastic org targets a single product, making the placeholder syntax unnecessary overhead for the common case.

See: https://github.com/search?q=org%3Aelastic+output_products+language%3AYAML&type=code&l=YAML

#### Release trigger routing (`bundle.releases`)
`changelog.yml` now accepts a `releases:` block under `bundle:` with three named keys. `github:` is a list of `{tag, profile}` entries where `tag` is a glob matched against the release tag at `gh-release` time. `unified:` maps a product ID to a profile for versioned stack releases. `serverless:` names a profile for serverless releases. All referenced profiles and product IDs are validated at config load time against `bundle.profiles` and `products.yml`.

```yaml
bundle:
  profiles:
    es-release:
      product: elasticsearch
    kibana-release:
      product: kibana
    serverless-release:
      product: elasticsearch

  releases:
    # monorepo: tag prefix selects the right profile
    github:
      - tag: "elasticsearch-*"
        profile: es-release
      - tag: "kibana-*"
        profile: kibana-release

    # versioned stack release: product ID selects the profile
    unified:
      - product: elasticsearch
        profile: es-release
      - product: kibana
        profile: kibana-release

    # serverless has no version; one profile handles it
    serverless:
      profile: serverless-release
```

#### `product:` field on profiles
Each profile now accepts a `product:` field that names a single validated product ID. This replaces the `output_products: "{product} {version} {lifecycle}"` pattern. `output_products:` and `output_directory:` on profiles are `[Obsolete]` and emit deprecation warnings at config load time; they continue to work until repos are migrated.

```yaml
# Before (deprecated — warns, still works)
bundle:
  profiles:
    es-release:
      output_products: "elasticsearch {version} ga"
      output_directory: elasticsearch/

# After
bundle:
  profiles:
    es-release:
      product: elasticsearch
```

#### `source: github_release` deprecated
`source: github_release` on a profile now emits a deprecation warning at config load time, directing authors to `bundle.releases.github`. Config load succeeds; no repos are broken mid-migration. `ProfileFilterResolver.ResolveAsync` no longer accepts `IGitHubReleaseService` or `IGitHubCommitRangeService` — those parameters are removed from `ChangelogBundlingService` and `ChangelogRemoveService`.

```yaml
# Before (deprecated — warns, still works)
bundle:
  profiles:
    es-release:
      source: github_release

# After: move the routing to bundle.releases.github
bundle:
  releases:
    github:
      - tag: "v*"
        profile: es-release
```

#### Tests updated
Tests that exercised `source: github_release` bundling are updated to assert a deprecation warning (not an error) and a non-null config result. Plan tests that relied on `source: github_release` config use equivalent valid profiles.

**Out of scope:** The `gh-release` command runtime does not yet read `bundle.releases.github` to select a profile automatically — that follow-up will be a separate PR.

## Verify

```bash
dotnet test tests/Elastic.Changelog.Tests/
# BundleProfileGitHubReleaseTests.SourceGithubRelease_EmitsDeprecationWarning
# BundleCdnSourcingTests.ProfileSourceGithubRelease_EmitsDeprecationWarning
```


lcawl:
I haven't tried running any tests with these code changes, but here are some thoughts just based on the description:

> This replaces source: github_release on profiles — it was redundant with the gh-release command
>...
> source: github_release on a profile now emits a deprecation warning at config load time

IIRC the specific reason I added that  profile and option on the bundle command was so that teams could use git-tag-based bundles even if they created changelogs in advance. That is to say, your choice of bundling method should not limit your options for changelog creation, IMO. If this new workflow can still provide that flexibility, that's fine.

> three named sub-keys (github, unified, serverless)

… could we maybe have a different label than “serverless”? My assumption is that workflow applies to all date-based releases (which affects more products than just `cloud-serverless`).
If there’s some other difference between the “unified” and “serverless” release, it’s not clear from the examples.

> A GitHub code search confirmed every output_products: usage in the elastic org targets a single product, making the placeholder syntax unnecessary overhead for the common case.

The only one I can think of at this moment that would be a valid multi-product case would be: `kibana, security, observability`, since releases in the kibana repo currently yield https://www.elastic.co/docs/release-notes/kibana, https://www.elastic.co/docs/release-notes/observability, and https://www.elastic.co/docs/release-notes/security from a single commit range (and I'd love to just handle that in a single release bundle if at all possible). Would that still be supported after this PR?

> Each profile now accepts a product: field that names a single validated product ID. This replaces the output_products: "{product} {version} {lifecycle}" pattern. 

Is the `{version}` still passed into the bundle nonetheless? For example, we currently pass in a date-based version for the Kibana Serverless bundles like this:

```sh
docs-builder changelog bundle serverless-release 2026-05-06 ./docs/temp/prs.txt
```
... and we'd need to continue to be able to do the same until such a time as the bundles are completely automated. And maybe even thereafter since we'll continue to need adhoc note bundles.

Mpdreamz:
Thanks for the review @lcawl!

> IIRC the specific reason I added that profile and option on the bundle command was so that teams could use git-tag-based bundles even if they created changelogs in advance.

Fully preserved. `bundle.releases.github` (tag glob → profile) handles the routing; the bundling mechanism itself is unchanged. Teams that pre-create changelog entries work exactly as before — `gh-release` (or the bundle-on-release automation) just picks the right profile from the config instead of inferring it.

> could we maybe have a different label than serverless? … it's not clear from the examples.

`serverless:` is gone — collapsed into `products:`, which now covers all product-scoped releases (versioned stack and date-based alike). Both `github:` and `products:` are now plain dictionaries (key → value), which reads more naturally:

```yaml
bundle:
  releases:
    github:                          # tag glob: profile
      "v*": my-profile
      "elasticsearch-*": es-release
    products:                        # product ID: profile
      elasticsearch: es-release
      cloud-serverless: serverless-release
```

> kibana, security, observability … Would that still be supported?

Yes — since `output_products` is deprecated (warns) but still works, the Kibana multi-product case keeps working as-is for now. The new `product:` field is intentionally singular; multi-product-from-one-profile is a separate use case we'll design when the need is concrete. The deprecation warning gives teams time to migrate at their own pace.

> Is the `{version}` still passed into the bundle nonetheless?

Yes, completely unchanged. `docs-builder changelog bundle serverless-release 2026-05-06 ./docs/temp/prs.txt` works exactly as before. `product:` is syntactic sugar for the output name — the version argument is still the positional second arg to the bundle command.

PR #4179, opened 2026-09-21
Pilot C# 15 union and closed on LinkResolution

Mpdreamz:
Introduces the `LinkResolution` union replacing the `bool TryResolve(Action<string>, Uri, out Uri?)` interface, and annotates four intentionally-finite hierarchies with `closed`. Call sites become exhaustive switches enforced by CS8509; the diagnostic side-channel is deleted entirely.

**Affects:** Cross-link resolution (`Elastic.Documentation.Links`), Markdown pipeline (`Elastic.Markdown`), code style guidance (`AGENTS.md`)

**Prompt summary:** Plan and apply C# 15 `union` and `closed` to docs-builder, starting with the quietest project in the repo as a pilot. Demonstrate both features together, fix real defects, and update the coding guidelines to prefer unions over ad-hoc bool-plus-out patterns going forward.

## Why

`ICrossLinkResolver.TryResolve` carries an `Action<string> errorEmitter` side channel through five method signatures solely because the return type can't carry *why* resolution failed. Nine distinct failure modes collapse into the same `false` + emitter call, so callers can't distinguish a synthesised dev URL from a genuine success, and two call sites (`AssemblerBuilder`, `CodexBuildService`) discard the bool entirely and rely wholly on the lambda firing. Three of four real consumers immediately write an adapter lambda just to bridge the emitter into a local diagnostic sink.

`closed` on the four annotated hierarchies (`DocumentationFile`, `RoleLeaf`, `AskAiEvent`, `TopNavRenderItem`) removes the implicit possibility of unknown subtypes appearing at a match site without compile-time error.

#### `LinkResolution` union (new type)
`src/Elastic.Documentation.Links/CrossLinks/LinkResolution.cs` declares a nine-case union — `LinkResolved`, `LinkSynthesized`, and seven failure cases — each carrying exactly the payload its diagnostic needs. A `ResolvedUri()` extension method holds the single exhaustive switch, so CS8509 fires in one place if the union gains a new case. A `ToDiagnosticMessage()` extension replaces the six inline format strings that were previously injected via the emitter.

#### `ICrossLinkResolver` interface redesigned
`Resolve(Uri) → LinkResolution` replaces `TryResolve(Action<string>, Uri, out Uri?)`. The `Action<string>` parameter is deleted from five signatures. `NoopCrossLinkResolver` and `StepCrossLinkResolver` — byte-identical implementations in two assemblies — both collapse to `return new LinkResolutionUnavailable()`. `LinkSynthesized` makes a synthesised development URL distinguishable from a real `LinkResolved` for the first time.

#### Two silent discard sites fixed
`AssemblerBuilder.cs` and `CodexBuildService.cs` were `_ = linkResolver.TryResolve(…)` — the bool discarded, correctness resting solely on the lambda. Both now match on `LinkResolution` and emit an error when `ResolvedUri()` returns null.

#### `closed` annotations (behaviour-neutral)
`DocumentationFile`, `RoleLeaf`, `AskAiEvent`, and `TopNavRenderItem` gain `closed`; the redundant `abstract` keyword is removed from each. Wire format is unchanged — `closed` does not affect `[JsonPolymorphic]` serialization.

#### Code-style rule updated
`AGENTS.md` replaces the `TryGet` bullet with a union-first rule: a method with more than one outcome returns a `union`. `TryGetX(key, out value)` remains the documented exception for genuine keyed lookups where the out parameter carries no error information.

## Verify

```bash
dotnet test tests/Navigation.Tests/
./build.sh unit-test
./build.sh lint
```

**Stack:** 4 of 4, on top of [#4166](https://github.com/elastic/docs-builder/pull/4166) (`ReadOnlyMemoryStream`)

**Out of scope:** `Elastic.Changelog` — highest union density in the repo but 301 file-touches in eight weeks; items are parked in the plan backlog until that area settles.


Mpdreamz (commented):
Fixed in 27a8e636c — added `is LinkResolutionUnavailable` early-return guards in `DiagnosticLinkInlineParser`, `DirectiveLinkValidator`, and `DocumentationSetNavigation` before the diagnostic path. All three sites now pass silently when the resolver is intentionally unavailable, and still emit errors for real resolution failures (LinkNotInIndex, LinkAnchorNotFound, etc.).

PR #4180, opened 2026-09-21
Add a stepper option that keeps step titles out of the page outline

florent-leborgne:
## Summary

`{stepper}` can set `:toc: false`.

- Step titles keep the size and weight of the heading level the stepper would have used.
- They render as a `div`, not an HTML heading, so they do not appear in **On this page**.
- Each step still has an anchor.
- The default stays unchanged: step titles are headings and they appear in the table of contents.

This supports elastic/docs-content#8123. The chart pages will use the option once this release is available to the docs build.

## Generative AI disclosure

- [x] Yes
- [ ] No

Tool and model used: Cursor Grok 4.7. It drafted the option, the tests, and the syntax note.


florent-leborgne:
Moved onto a branch in elastic/docs-builder so CI can run. Replacement: https://github.com/elastic/docs-builder/pull/4181

PR #4181, opened 2026-09-21
Add a stepper option that keeps step titles out of the page outline

florent-leborgne:
`:toc: false` on `{stepper}` keeps step titles on the page and out of the page outline. They keep the same size, and they are not HTML headings.

**Affects:** Authoring, Site UI

**Prompt summary:** Florent wants chart procedures to keep their step titles on the page and out of the page outline. He asked for a stepper option that does not render those titles as HTML headings, so they look the same and drop out of On this page.

## Why

A stepper step title is an HTML heading. The page outline collects those headings, so a short procedure adds several entries next to the real sections. The only current opt-out nests the stepper in another directive. That wrap adds a tab or a dropdown around the procedure.

#### Opt out of headings

`:toc: false` on `{stepper}` keeps each step title at the size of the heading level the stepper would have used. The title is a `div`, not a heading. Without the option, the title stays a heading.

#### Page outline

The page outline skips those titles. Each step still has an anchor, so a link to that step still resolves.

#### Syntax note

`docs/syntax/stepper.md` documents `:toc: false` beside the existing table-of-contents section.

## Verify

```bash
dotnet test tests/Elastic.Markdown.Tests/ --filter FullyQualifiedName~StepperToc
```

**Out of scope** The chart pages that will use this option are in [elastic/docs-content#8453](https://github.com/elastic/docs-content/pull/8453).


docs/syntax/stepper.md
```diff
@@ -205,6 +205,44 @@ To see how dynamic mapping works, add a new document to the `books` index with a
...
+
+### Omit steps from the table of contents
+
+Set `:toc: false` on `{stepper}` when step titles should not be headings. The titles keep the same size and weight as the heading level the stepper would have used. They are not HTML headings, and they do not appear in **On this page**. Each step still has an anchor, so you can link to it.
+
+The default is to include step titles in the table of contents.
+
+This stepper is rendered on the page. Its titles are not listed in **On this page**. [Prepare the project](#prepare-the-project) still opens the first step.
+
+:::::{stepper}
```
florent-leborgne, line 218: Intentionally not using the markdown/output tabs that we use for examples... because nesting that in a tab would automatically remove from the TOC anyways, so this is to show they're indeed not showing in the TOC.

docs/syntax/stepper.md
```diff
@@ -205,6 +205,44 @@ To see how dynamic mapping works, add a new document to the `books` index with a
...
+To keep those titles in the procedure and out of the table of contents, set `:toc: false`. See [Omit steps from the table of contents](#omit-steps-from-the-table-of-contents).
+
+### Omit steps from the table of contents
+
+Set `:toc: false` on `{stepper}` when step titles should not be headings. The titles keep the same size and weight as the heading level the stepper would have used. They are not HTML headings, and they do not appear in **On this page**. Each step still has an anchor, so you can link to it.
+
+The default is to include step titles in the table of contents.
+
+This stepper is rendered on the page. Its titles are not listed in **On this page**. [Prepare the project](#prepare-the-project) still opens the first step.
+
```
florent-leborgne, line 217:
```suggestion

**Example**:

```

docs/syntax/stepper.md
```diff
@@ -205,6 +205,46 @@ To see how dynamic mapping works, add a new document to the `books` index with a
...
+**Example**:
```
florent-leborgne:
```suggestion
This stepper is rendered on the page. Its titles are not listed in **On this page**. [Prepare the project](#prepare-the-project) still opens the first step:
```

Mpdreamz (changes requested):
`The titles keep the same size and weight as the heading level the stepper would have used`


This to me reads really ambiguous, does it retain the steppers auto subheading behavior e.g +1 # of the context it's used under? If so how if it renders as diffs? 

The not wanting it in the toc as an option makes sense to me.

florent-leborgne:
@Mpdreamz do the changes help?

Mpdreamz (approved):

